PR #12, opened 2021-01-22
Feature/roi clip for surface

toloudis:
Volume-viewer supports axis-aligned clipping regions, but until now it has never been implemented for isosurface meshes.

This code change uses threejs's built-in clip plane support to orient the clipping regions for isosurface meshes so they are matched up with those used in volume rendering.

### DIFF
--- a/src/MeshVolume.js
+++ b/src/MeshVolume.js
@@ -77,6 +77,7 @@ export default class MeshVolume {
 
   setRotation(eulerXYZ) {
     this.meshPivot.rotation.copy(eulerXYZ);
+    this.updateClipFromBounds();
   }
 
   setResolution(x, y) {}
@@ -197,6 +198,9 @@ export default class MeshVolume {
     const xmax = this.bounds.bmax.x;
     const ymax = this.bounds.bmax.y;
     const zmax = this.bounds.bmax.z;
+
+    const euler = this.meshPivot.rotation;
+
     // assuming these are 0..1
     for (let channel = 0; channel < this.meshrep.length; ++channel) {
       if (!this.meshrep[channel]) {
@@ -205,22 +209,22 @@ export default class MeshVolume {
       const planes = [];
       // up to 6 planes.
       if (xmin > -0.5) {
-        planes.push(new Plane(new Vector3(1, 0, 0), this.meshRoot.position.x + -xmin * this.scale.x));
+        planes.push(new Plane(new Vector3(1, 0, 0).applyEuler(euler), this.meshRoot.position.x + -xmin * this.scale.x));
       }
       if (ymin > -0.5) {
-        planes.push(new Plane(new Vector3(0, 1, 0), this.meshRoot.position.y + -ymin * this.scale.y));
+        planes.push(new Plane(new Vector3(0, 1, 0).applyEuler(euler), this.meshRoot.position.y + -ymin * this.scale.y));
       }
       if (zmin > -0.5) {
-        planes.push(new Plane(new Vector3(0, 0, 1), this.meshRoot.position.z + -zmin * this.scale.z));
+        planes.push(new Plane(new Vector3(0, 0, 1).applyEuler(euler), this.meshRoot.position.z + -zmin * this.scale.z));
       }
       if (xmax < 0.5) {
-        planes.push(new Plane(new Vector3(-1, 0, 0), this.meshRoot.position.x + xmax * this.scale.x));
+        planes.push(new Plane(new Vector3(-1, 0, 0).applyEuler(euler), this.meshRoot.position.x + xmax * this.scale.x));
       }
       if (ymax < 0.5) {
-        planes.push(new Plane(new Vector3(0, -1, 0), this.meshRoot.position.y + ymax * this.scale.y));
+        planes.push(new Plane(new Vector3(0, -1, 0).applyEuler(euler), this.meshRoot.position.y + ymax * this.scale.y));
       }
       if (zmax < 0.5) {
-        planes.push(new Plane(new Vector3(0, 0, -1), this.meshRoot.position.z + zmax * this.scale.z));
+        planes.push(new Plane(new Vector3(0, 0, -1).applyEuler(euler), this.meshRoot.position.z + zmax * this.scale.z));
       }
       this.meshrep[channel].traverse(function(child) {
         if (child instanceof Mesh) {
--- a/src/MeshVolume.js
+++ b/src/MeshVolume.js
@@ -183,6 +183,8 @@ export default class MeshVolume {
   }
 
   updateClipRegion(xmin, xmax, ymin, ymax, zmin, zmax) {
+    // incoming values expected to be between 0 and 1.
+    // I shift them here to be between -0.5 and 0.5
     this.bounds = {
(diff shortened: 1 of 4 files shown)


src/MeshVolume.js
```diff
@@ -108,6 +118,7 @@ export default class MeshVolume {
       specular: new Color(defaultMaterialSettings.specularColor),
       opacity: alpha,
       transparent: alpha < ALPHA_THRESHOLD,
+      side: DoubleSide,
```
schoinh, line 121: Just curious if this change is relevant to clipping?

src/MeshVolume.js
```diff
@@ -171,6 +182,60 @@ export default class MeshVolume {
     return this.meshrep[channel].userData.isovalue;
   }
 
+  updateClipRegion(xmin, xmax, ymin, ymax, zmin, zmax) {
+    this.bounds = {
+      bmin: new Vector3(xmin - 0.5, ymin - 0.5, zmin - 0.5),
+      bmax: new Vector3(xmax - 0.5, ymax - 0.5, zmax - 0.5),
```
schoinh, line 188: Why `- 0.5`?

src/ThreeJsPanel.js
```diff
@@ -67,6 +67,7 @@ export class ThreeJsPanel {
         // set pixel ratio to 0.25 or 0.5 to render at lower res.
         this.renderer.setPixelRatio(window.devicePixelRatio);
         this.renderer.state.setBlending(NormalBlending);
+        this.renderer.localClippingEnabled = true;
```
schoinh, line 70: Probably nit picking but is this something that can't be set under line 64 when you're constructing `this.renderer`?

schoinh (commented):
Could you please explain for me:
1. What are all the 0.5's for?
2. What is a `meshrep` (and `channel`)?
😆

toloudis:
> 1. What are all the 0.5's for?

The clipping region is specified in the UI by choosing ranges from 0..1 as a percentage of the x,y, and z size of the volume. (For example, as a user I would say clip Z between 0.25 and 0.75 of its range.). Then most of the internalized representations are more naturally in the -0.5..0.5 range, so I accomplish that by subtracting 0.5 from all the values.

> 2. What is a `meshrep` (and `channel`)?

In this code, a `meshrep` is short for "mesh representation".  It's the actual THREEjs Object3D that holds the isosurface mesh for use in the threejs renderer.  There can be one isosurface per volume channel.  

Channel in this case is a general term describing the volume data: the volume data is really composed of several superimposed volumes (or you can think of it as several intensities at each voxel).  For each volume channel, we can generate a separate isosurface.

src/MeshVolume.js
toloudis: this change is relevant because now that objects can be clipped, the insides of their surfaces can be seen.  Without that line, the inside surfaces would be invisible due to WebGL backface culling (an optimization to save from rendering faces that are pointing away from the viewer and wouldn't normally be seen).

src/ThreeJsPanel.js
toloudis: Glad you made me look at this -- I think this same line needs to be added in the block below here also.  I'll see if it's allowed in the properties passed in to the constructor too.

src/MeshVolume.js
```diff
@@ -70,15 +75,20 @@ export default class MeshVolume {
...
+  setRotation(eulerXYZ) {
+    this.meshPivot.rotation.copy(eulerXYZ);
+    this.updateClipFromBounds();
   }
 
   setResolution(x, y) {}
 
   setOrthoThickness(value) {}
 
-  setAxisClip(axis, minval, maxval, isOrthoAxis) {}
```
meganrm, line 81: was this a placeholder you had written as a TODO?

src/MeshVolume.js
```diff
@@ -171,6 +182,60 @@ export default class MeshVolume {
...
+    const xmin = this.bounds.bmin.x;
+    const ymin = this.bounds.bmin.y;
+    const zmin = this.bounds.bmin.z;
+    const xmax = this.bounds.bmax.x;
+    const ymax = this.bounds.bmax.y;
+    const zmax = this.bounds.bmax.z;
+
+    const euler = this.meshPivot.rotation;
+
+    // assuming these are 0..1
```
meganrm, line 203: nit I think, you mean 0.1 here right?

meganrm (approved):

src/MeshVolume.js
toloudis: yes, i had left it here as an empty function just in case some day it would need to actually do something, like the ones above it.  The raymarch and pathtrace representations have the same api, so it was here for consistency.   Some day there will be a typescript conversion and it will be declared as an interface.

src/MeshVolume.js
toloudis: no, i am using `0..1` to mean in the range from 0 to 1. i can change the comment to make that more clear

meganrm: ooooh, yeah `0 <= x <= 1` would make more sense to me, or just `between 0 and 1, inclusive`

src/MeshVolume.js
schoinh: Ah!! 💡

src/MeshVolume.js
toloudis: ugh, the comment makes no sense where it is, anyway.  It's meant to refer to the bounding box extent values.

src/ThreeJsPanel.js
toloudis: The threejs docs don't list this as an allowed parameter for the constructor of the WebGLRenderer.  But I did move it down so that it is applied to both WebGL 1 and WebGL2 cases.

src/MeshVolume.js
```diff
@@ -171,6 +182,61 @@ export default class MeshVolume {
...
+      }
+      if (xmax < 0.5) {
+        planes.push(new Plane(new Vector3(-1, 0, 0).applyEuler(euler), this.meshRoot.position.x + xmax * this.scale.x));
+      }
+      if (ymax < 0.5) {
+        planes.push(new Plane(new Vector3(0, -1, 0).applyEuler(euler), this.meshRoot.position.y + ymax * this.scale.y));
+      }
+      if (zmax < 0.5) {
+        planes.push(new Plane(new Vector3(0, 0, -1).applyEuler(euler), this.meshRoot.position.z + zmax * this.scale.z));
+      }
```
schoinh, line 228: Maybe I'm missing something..... but can lines 209-228 be pulled outside of the for loop? Doesn't look like they depend on `channel`?

src/MeshVolume.js
schoinh: The new comment you just added makes a lot of sense!

src/MeshVolume.js
toloudis: They don't depend on channel but they need to be added once for each channel that has a mesh representation (because each single mesh needs to have its own clip planes set in threejs)

toloudis: I guess I could precompute them outside the loop and then add them inside the loop, of course!

schoinh: Yes, that's what I meant, I wasn't sure how many channels we might have or how much this would matter. Your call!

schoinh (approved):

PR #14, opened 2021-03-05
Update threejs version and fix isosurface color bug

toloudis:
Update threejs version to latest. 

Some small code fixups based on deprecation warnings:  matrix inverse has changed, and also they got rid of the Geometry class which caused one mesh data-filling function to be somewhat rewritten.

In testing the update, I found and fixed the following bug:

When updating a surface's isovalue, we destroy and re-create the surface, and to set the new surface to the correct color, the code was looking at the previous threejs object to find out what color it was.  But when the isovalue is changed to a value that doesn't generate any triangles, then there's no mesh in which to look up the color for the next change.

The fix is to just store the current colors (and opacities) in their own array.

### DIFF
--- a/src/MeshVolume.js
+++ b/src/MeshVolume.js
@@ -27,6 +27,9 @@ export default class MeshVolume {
 
     this.meshrep = [];
 
+    this.channel_colors = [];
+    this.channel_opacities = [];
+
     this.bounds = {
       bmin: new Vector3(-0.5, -0.5, -0.5),
       bmax: new Vector3(0.5, 0.5, 0.5),
@@ -93,6 +96,10 @@ export default class MeshVolume {
   //////////////////////////////
 
   updateMeshColors(channel_colors) {
+    // stash values here for later changes
+    this.channel_colors = channel_colors;
+
+    // update existing meshes
     for (var i = 0; i < this.volume.num_channels; ++i) {
       if (this.meshrep[i]) {
         var rgb = channel_colors[i];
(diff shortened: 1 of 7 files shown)


toloudis:
Also hopefully this will be a better version of #13 and we can get rid of that one.

src/vr/vrObjectControls.js
```diff
@@ -204,7 +204,7 @@ export class vrObjectControls {
       v1 = v1.normalize();
 
       var mio = new Matrix4();
-      mio.getInverse(obj3d.matrixWorld);
+      mio.copy(obj3d.matrixWorld).invert();
```
meganrm, line 207: is this a file you wrote or is it copied from some library?

src/MeshVolume.js
```diff
@@ -147,30 +155,11 @@ export default class MeshVolume {
...
-    } else {
-      this.meshrep[channel].traverse(function(child) {
-        if (child instanceof Mesh) {
-          opacity = child.material.opacity;
-          color = child.material.color;
-        }
-      });
-    }
+    let opacity = this.channel_opacities[channel];
+    let color = this.channel_colors[channel];
```
meganrm, line 159: this is real nit but here and below you're using snake_case in javascript files. In other repos I think our eslint rules are set up to enforce camelCase but I guess not here?

meganrm (approved):

src/vr/vrObjectControls.js
toloudis: wrote

toloudis: vr support in this lib is pretty much dead at the moment anyway

src/MeshVolume.js
toloudis: yup, I noticed this too, I've been in python land for too long. The code is due for a big typescript conversion at some point soon.  Probably at the time of the next big feature addition, I will take a pause and start adding conformance with our other more modern js/ts code.

blairlyons (approved):

PR #27, opened 2021-07-28
Feature/update deps and ts

toloudis:
Beginning to pull this codebase toward modern typescript.
Many files changed but nothing too huge.

* Update almost all the dev dependencies
* Add typescript support in the linter and babel parsers
* Convert a couple of files to typescript as a test
* Make sure the test app still runs
* remove old gradle and Jenkins files that are no longer used
* eslint is now only linting the src/ directory for the moment, and there are so many warnings I am running it in quiet mode for now.  Future changes will fix all lint warnings.
* ViveController had to be converted to a class but most of the changes there are just indenting/whitespace


src/Histogram.ts
```diff
@@ -1,12 +1,8 @@
 import { getColorByChannelIndex } from "./constants/colors.js";
 
-(function() {
-  if (!Math.clamp) {
-    Math.clamp = function(val, cmin, cmax) {
-      return Math.min(Math.max(cmin, val), cmax);
-    };
-  }
-})();
+function clamp(val: number, cmin: number, cmax: number): number {
```
EricJIsaac, line 3: Do the Math.clamp -> clamp changes correspond to a js vs. ts difference? (were they needed to pass the linter, or changed for another reason)

EricJIsaac (approved):

src/Histogram.ts
toloudis: Yes I think typescript was complaining but also, I have always felt bad about monkeypatching the global Math object.  It's really not great practice.

.eslintrc.js
```diff
@@ -1,16 +1,57 @@
...
   parserOptions: {
-    ecmaVersion: "2018",
+    project: "./tsconfig.json",
+    ecmaVersion: "2020",
     sourceType: "module",
   },
+  plugins: ["@typescript-eslint", "react"],
   rules: {
+    "@typescript-eslint/ban-types": ["warn"],
+    "@typescript-eslint/naming-convention": [
```
schoinh, line 17: Ooh, I didn't know you could lint for this

README.md
```diff
@@ -58,16 +58,6 @@ view3D.updateLuts(aimg);
...
-    rules: [
-        ...,
-        {
-            test: /Worker\.js$/,
-            use: 'worker-loader?inline=true'
-        }
-    ]
-```
 - in `VolumeViewer.jsx`
 ```JavaScript
```
schoinh, line 62: I know this is not part of this PR but I just noticed that this code block is not closed properly. The backticks in line 131 need to be un-indented. 

image

src/test/lut.test.js
```diff
@@ -1,8 +1,12 @@
 import { expect } from "chai";
 
-import Histogram from "../Histogram.js";
+import Histogram from "../Histogram.ts";
 import VolumeMaker from "../VolumeMaker.js";
 
+function clamp(val, cmin, cmax) {
+  return Math.min(Math.max(cmin, val), cmax);
+}
+
```
schoinh, line 9: Maybe throw this in a util file if it's going to be used in multiple places (here and in Histogram.ts)?

schoinh (approved):
This must feel good to do...

README.md
toloudis: Thanks! will fix that

src/test/lut.test.js
toloudis: The only reason I didn't is that the repeated use is in a unit test.  One more repetition and I will refactor it :)

toloudis:
> This must feel good to do...

It'll feel even better to convert the whole thing to typescript (but at the same time I dread how long it will take)

PR #28, opened 2021-08-13
Feature/expose histogram percentile

toloudis:
* expose a public Histogram function to get a percentile bin (for use in client apps)
* Add types to Histogram.ts
* add tests and improve the minMax generator function for edge cases

### DIFF
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -173,7 +173,7 @@ export default class Histogram {
   lutGenerator_labelColors() {
     const lut = new Uint8Array(LUT_ARRAY_LENGTH).fill(0);
     // TODO specify type for control point
-    const controlPoints:any[] = [];
+    const controlPoints: any[] = [];
     controlPoints.push({ x: 0, opacity: 0, color: [0, 0, 0] });
     let lastr = 0;
     let lastg = 0;
@@ -226,6 +226,26 @@ export default class Histogram {
   }
 
   /**
+   * Find the bin that contains the percentage of pixels below it
+   * @return {number}
+   * @param {number} pct
+   */
+  findBinOfPercentile(pct) {
+    const pixcount = this.nonzeroPixelCount + this.bins[0];
+    const limit = pixcount * pct;
+
+    var i = 0;
+    var count = 0;
+    for (i = 0; i < this.bins.length; ++i) {
+      count += this.bins[i];
+      if (count > limit) {
+        break;
+      }
+    }
+    return i;
+  }
+
+  /**
    * Generate a lookup table based on histogram percentiles
    * @return {Lut}
    * @param {number} pmin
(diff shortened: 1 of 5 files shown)


src/Histogram.ts
```diff
@@ -414,15 +449,15 @@ export default class Histogram {
...
+  lutGenerator_fromControlPoints(controlPoints: ControlPoint[]): Lut {
     const lut = new Uint8Array(LUT_ARRAY_LENGTH).fill(0);
 
     if (controlPoints.length === 0) {
       return { lut: lut, controlPoints: controlPoints };
     }
 
     // ensure they are sorted in ascending order of x
-    controlPoints.sort((a, b) => (a.x > b.x ? 1 : -1));
+    controlPoints.sort((a, b) => (a.x > b.x ? 1 : a.x < b.x ? -1 : 0));
```
EricJIsaac, line 460: Would `=> (a.x - b.x)` work here for the compare function?

EricJIsaac (approved):

src/Histogram.ts
toloudis: I believe it would.  This is a place that is actually potentially broken when you try to make a step function and pass it in.
The step function has two values at the same x and I'm not clear whether the sort is swapping their order or not. (The underlying implementation of sort is unknown to me; maybe it's documented somewhere)

toloudis: ha! it is documented:
```
Note: The ECMAScript Standard, 10th edition (2019) algorithm mandates stable sorting, which means elements that compare equal must remain in their original order with respect to each other. This behaviour may not be respected by older browsers.
```

src/Histogram.ts
```diff
@@ -27,6 +27,15 @@ const LUT_ARRAY_LENGTH = LUT_ENTRIES * 4;
  * @property {Array.<number>} lut LUT_ARRAY_LENGTH element lookup table as array (maps scalar intensity to a rgb color plus alpha)
  * @property {Array.<ControlPoint>} controlPoints
  */
+type ControlPoint = {
+  x: number;
+  opacity: number;
+  color: [number, number, number];
```
schoinh, line 33: Just noticed that this type for `color` doesn't match what's in the JSDoc block in line 22

src/Histogram.ts
schoinh: So... should we use `a.x - b.x` here for readability?

toloudis: No prob, will make that change!

src/Histogram.ts
toloudis: will fix it.  I wonder if jsdoc could make documentation from types directly...

schoinh (approved):

src/Histogram.ts
```diff
@@ -108,38 +117,66 @@ export default class Histogram {
    * @param {number} b
    * @param {number} e
    */
-  lutGenerator_minMax(b, e) {
-    var lut = new Uint8Array(LUT_ARRAY_LENGTH);
-    let range = e - b;
-    if (range < 1) {
-      range = 255;
+  lutGenerator_minMax(b: number, e: number): Lut {
```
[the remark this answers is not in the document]
toloudis: b=begin, e=end

src/Histogram.ts
```diff
@@ -226,37 +263,35 @@ export default class Histogram {
...
-   * @param {number} pmax
+   * Find the bin that contains the percentage of pixels below it
+   * @return {number}
+   * @param {number} pct
    */
-  lutGenerator_percentiles(pmin, pmax) {
-    // e.g. 0.50, 0.983 starts from 50th percentile bucket and ends at 98.3 percentile bucket.
-
+  findBinOfPercentile(pct: number): number {
     const pixcount = this.nonzeroPixelCount + this.bins[0];
```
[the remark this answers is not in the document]
toloudis: bins[0] contains the number of pixels of intensity zero.  nonzero pixel count + count of intensity 0 = total number of pixels.
should be the same as the length of the data array used to initialize the histogram.

src/Histogram.ts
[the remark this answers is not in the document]
schoinh: 🤯

PR #29, opened 2021-08-21
build es modules and use that as the primary distribution

toloudis:
This PR is a bit of cleanup/robustness in the build toolchain.
The "main" entry in package.json was pointing to an obsolete lib directory which doesn't exist anymore.
I also update mocha/chai and make sure settings are correct to run tests on es modules.


MatteBailey (approved):

schoinh (approved):
!!

PR #30, opened 2021-08-31
Bugfix/noise pattern

toloudis:
Two bug fixes are included here.
1. In some volumes, (I believe this might be limited to texture atlases whose tiles fill the entire texture space with no empty tiles at the end), there was a banding artifact at the top or bottom edge of the volume.
image (1)
2. There was excessive noise when looking at side views of the volume.
image (2)

Bug 1 was fixed by correcting the way I select the z slices from which to sample.  This was cleaned up and there seemed to be a clear off-by-one error in there.

Bug 2 was "improved" but not 100% resolved, by changing the step sizes for ray marching through the volume.  The step size now accounts for the nonuniform scale of the volume cube, which was implicitly folded into the transformation of the ray direction.  Noise is inherent to this volume rendering algorithm, and is intentional to overcome aliasing.

### DIFF
--- a/src/RayMarchedAtlasVolume.js
+++ b/src/RayMarchedAtlasVolume.js
@@ -102,6 +102,9 @@ export default class RayMarchedAtlasVolume {
     this.scale = scale;
 
     this.cubeMesh.scale.copy(new Vector3(scale.x, scale.y, scale.z));
+    this.setUniform("volumeScale", scale);
+    
+    console.log("scale= ", scale);
   }
 
   setRayStepSizes(primary, secondary) {}
@@ -161,6 +164,9 @@ export default class RayMarchedAtlasVolume {
       this.setOrthoThickness(thicknessPct);
     }
     // else don't set? use previous value?
+    else {
+      this.setOrthoThickness(1.0);
+    }
 
     this.setUniform("AABB_CLIP_MIN", this.bounds.bmin);
     this.setUniform("AABB_CLIP_MAX", this.bounds.bmax);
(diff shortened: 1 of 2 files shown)


toloudis:
"after" pictures: 
1) Banding removed
AICS 3D Cell Viewer 2021-08-31 11-42-23
2) Noise is still present but is less offensive
AICS 3D Cell Viewer 2021-08-31 11-44-43

src/RayMarchedAtlasVolume.js
```diff
@@ -161,6 +163,9 @@ export default class RayMarchedAtlasVolume {
       this.setOrthoThickness(thicknessPct);
     }
     // else don't set? use previous value?
```
MatteBailey, line 165: Does this comment need to be updated?

MatteBailey (approved):

src/constants/volumeRayMarchShader.js
```diff
@@ -188,8 +192,11 @@ export const rayMarchingFragmentShaderSrc = [
 
   //'  //estimate step length',
   "  const int maxSteps = 512;",
-  "  float csteps = clamp(float(BREAK_STEPS), 1.0, float(maxSteps));",
-  "  float invstep = 1.0/csteps;",
+  // modify the 3 components of eye_d by volume scale
+  "  float scaledSteps = float(BREAK_STEPS) * length((eye_d.xyz/volumeScale));",
+  "  float csteps = clamp(float(scaledSteps), 1.0, float(maxSteps));",
+  "  float invstep = (tfar-tnear)/csteps;",
+//  "  float invstep = 1.0/csteps;",
```
kmitcham, line 199: did you mean to comment out this line instead of deleting it?

kmitcham (approved):

PR #33, opened 2021-10-02
Fix/handle same min-max values in lutGenerator_minMax

schoinh:
This should fix the bug we're currently seeing in CFE / website-3d-cell-viewer: [website-3d-cell-viewer crashes when trying to view volume settings for contour channels](https://aicsjira.corp.alleninstitute.org/browse/ANIMCELLGR-686)

### Cause of bug: 
NaN values getting passed in as control points. The NaN values were coming from divisions by zero that were happening in the case where the min and max values passed into the min-max lookup table generator (`lutGenerator_minMax` in Histogram.ts) were equal. The direct lookup table generation was handling the case where min = max (`b === e`) but the code that generates control points was not handling that case.

### Solution: 
Add more assertions to tests to check that no NaN values are generated, and modify control point generation logic to deal with min and max values being equal (avoid division by zero and create a step function with an almost vertical slope)

### Other changes:
- Wrote tests for a few more edge cases and wrote more code to make those tests pass
- Added comments (hopefully they're accurate)

@toloudis and I worked on this together.

- [x] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [x] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_
- [x] Provide description and context of changes.
- [x] Provide relevant tests for your feature or bug fix.
- [x] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

### DIFF
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -16,16 +16,16 @@ const LUT_ENTRIES = 256;
 const LUT_ARRAY_LENGTH = LUT_ENTRIES * 4;
 
 /**
- * @typedef {Object} ControlPoint
+ * @typedef {Object} ControlPoint Used for the TF (transfer function) editor GUI. Need to be converted to LUT for rendering.
  * @property {number} x The X Coordinate
  * @property {number} opacity The Opacity, from 0 to 1
  * @property {Array.<number>} color The Color, 3 numbers from 0-255 for r,g,b
  */
 
 /**
- * @typedef {Object} Lut
- * @property {Array.<number>} lut LUT_ARRAY_LENGTH element lookup table as array (maps scalar intensity to a rgb color plus alpha)
- * @property {Array.<ControlPoint>} controlPoints
+ * @typedef {Object} Lut Used for rendering.
+ * @property {Array.<number>} lut LUT_ARRAY_LENGTH element lookup table as array (maps scalar intensity to a rgb color plus alpha, with each value from 0-255))
+ * @property {Array.<ControlPoint>} controlPoints 
  */
 type ControlPoint = {
   x: number;
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -172,11 +172,7 @@ export default class Histogram {
     // Add starting point at x = 0
     let startVal = 0;
     if (b < 0) {
-      if (e === b) {
-        startVal = 1;
-      } else {
-        startVal = -b / (e - b);
-      }
+      startVal = -b / (e - b);
     }
     controlPoints.push({ x: 0, opacity: startVal, color: [255, 255, 255] });
     
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -156,7 +156,7 @@ export default class Histogram {
         ],
       };
     }
-    if (b > 255 && e > 255) {
+    if (b >= 255 && e >= 255) {
       return {
         lut: lut,
         controlPoints: [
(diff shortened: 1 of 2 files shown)


src/Histogram.ts
```diff
@@ -144,21 +146,64 @@ export default class Histogram {
...
+      };
+    }
+
+    // Generate 2 to 4 control points for a minMax LUT, from left to right
     const controlPoints: ControlPoint[] = [];
+
+    // Add starting point at x = 0
     let startVal = 0;
-    if (b <= 0) {
-      startVal = -b / (e - b);
```
schoinh, line 150: when `e === b` this (and ln 161) would divide by 0 and produce a NaN

src/Histogram.ts
```diff
@@ -144,21 +146,63 @@ export default class Histogram {
...
+    if (b < 0) {
+      if (e === b) {
```
toloudis, line 175: I think your handling of edge case e<0 && b<0 (line 150) makes line 175 unreachable.  I'll leave it up to you if you'd prefer to keep it here but it could be removed.

src/test/lut.test.js
```diff
@@ -144,6 +144,43 @@ describe("test histogram", () => {
...
+      // Spot check but all opacity values should be 255
+      expect(lut.lut[3]).to.eql(255);
+      expect(secondlut.lut[3]).to.eql(255);
+      expect(lut.lut[120 * 4 + 3]).to.eql(255);
+      expect(secondlut.lut[120 * 4 + 3]).to.eql(255);
+      expect(lut.lut[255 * 4 + 3]).to.eql(255);
+      expect(secondlut.lut[255 * 4 + 3]).to.eql(255);
+      // Make sure no NaN values
+      lut.controlPoints.forEach(controlPoint => {
+        expect(controlPoint.opacity).to.be.finite;
```
toloudis, line 182: in this test case, all the control point opacities should be exactly 1.0

src/test/lut.test.js
```diff
@@ -175,6 +220,26 @@ describe("test histogram", () => {
...
+      // Spot check but all opacity values should be 0
+      expect(lut.lut[3]).to.eql(0);
+      expect(secondlut.lut[3]).to.eql(0);
+      expect(lut.lut[120 * 4 + 3]).to.eql(0);
+      expect(secondlut.lut[120 * 4 + 3]).to.eql(0);
+      expect(lut.lut[255 * 4 + 3]).to.eql(0);
+      expect(secondlut.lut[255 * 4 + 3]).to.eql(0);
+      // Make sure no NaN values
+      lut.controlPoints.forEach(controlPoint => {
+        expect(controlPoint.opacity).to.be.finite;
```
toloudis, line 241: in this test case you can make this stricter and check all opacities === 0

toloudis (approved):

src/Histogram.ts
schoinh: Good point, removed

schoinh: This made me think this change (https://github.com/AllenInstitute/volume-viewer/pull/33/commits/a59293b66be7da9f669050dd8b260c37df27b4f1) would also be good, but let me know if that was overkill / more confusing!

toloudis: looks good to me!  if b>=255 then the whole range has to be zero opacity (assuming e>=b which it must be!).

src/test/lut.test.js
```diff
@@ -144,6 +144,42 @@ describe("test histogram", () => {
       expect(secondlut.lut[1 * 4 + 3]).to.eql(255);
       expect(lut.lut[255 * 4 + 3]).to.eql(255);
       expect(secondlut.lut[255 * 4 + 3]).to.eql(255);
+      // Make sure no NaN values
+      lut.controlPoints.forEach(controlPoint => {
+        expect(controlPoint.opacity).to.be.finite;
+      })
+    });
+    it("is consistent for minMax edge case 120,120", () => {
```
meganrm, line 152: instead of calling this "edge case 120,120" could you put what makes this an edge case in the description? Same for the tests that follow

schoinh: Good call. I made the descriptions more clear and grouped some tests together. (see new groups `lutGenerator_minMax` and `lutGenerator_windowLevel`)

image

meganrm (approved):

PR #34, opened 2021-10-06
convert MarchingCubes to a class

toloudis:
Fixing a bug in which we are suddenly incompatible with threejs's es modules.  We want MarchingCubes to inherit from THREE.ImmediateRenderObject properly.
This was causing a serious exception in production.

### DIFF
--- a/src/MarchingCubes.js
+++ b/src/MarchingCubes.js
@@ -8,693 +8,693 @@
 
 import { BufferAttribute, BufferGeometry, ImmediateRenderObject } from "three";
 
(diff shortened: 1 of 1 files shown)


meganrm (approved):

schoinh (approved):
Why did this "suddenly" happen? Threejs version update?

toloudis:
> Why did this "suddenly" happen? Threejs version update?

I guess so... maybe something in package-lock.  Honestly I'm not sure but this seems to fix it without changing any functionality.

PR #35, opened 2021-10-18
typescript conversion

toloudis:
Massive code overhaul to convert most of the project to typescript while maintaining all existing functionality and public interfaces. 

This change should not break compatibility with client apps like website-3d-cell-viewer and nbvv.   The purpose is to modernize the code and position it well for further development.

This branch was started as a test for timelapse data but in order to complete that work it became apparent that the typescript conversion should happen as well.  The timelapse data changes are nearly entirely in the example app in public/index.js.

VR functionality is removed. It was essentially running deprecated code anyway and the extra 3d assets for hand controllers were excessive unnecessary bloat.

Many variables were renamed to comply with the naming conventions in the eslint configuration.  In a few cases, variables that violate the naming convention were in public interfaces used by clients and so had to be allowed to pass linting by temporarily disabling the rule.

### DIFF
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -53,7 +53,8 @@ export default class RayMarchedAtlasVolume {
 
     this.cubeTransformNode = new Group();
     this.cubeTransformNode.name = "VolumeContainerNode";
-    // REINSTATE THIS TO SHOW BOUNDING BOX!
+    // TODO: when bounding box UX is determined,
+    // uncomment the following line to show the box
     //this.cubeTransformNode.add(this.boxHelper);
     this.cubeTransformNode.add(this.cubeMesh);
 


src/FuseWorker.js
```diff
@@ -1,17 +1,17 @@
-var channels = [];
-var npx = 0;
+const channels = [];
+let npx = 0;
```
schoinh, line 2: Does this file still have work left to do before changing the file extension to .ts?

src/MarchingCubes.ts
```diff
@@ -420,169 +501,6 @@ class MarchingCubes extends ImmediateRenderObject {
       renderCallback(this);
     };
 
-    /////////////////////////////////////
-    // Metaballs
-    /////////////////////////////////////
-
-    // Adds a reciprocal ball (nice and blobby) that, to be fast, fades to zero after
-    // a fixed distance, determined by strength and subtract.
```
schoinh, line 428: Why was this deleted?

src/TrackballControls.js
```diff
@@ -7,548 +7,554 @@
...
-      this.screen.width = box.width;
-      this.screen.height = box.height;
-    }
-  };
+// this file is heavily derived from threejs and so we disable a lot of our
+// linting rules just for this file, to allow it to keep a similar form to the
+// original.
+/* eslint-disable @typescript-eslint/naming-convention */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-this-alias */
```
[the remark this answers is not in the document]
toloudis: Yes, switching this file to ts causes a lot of errors that would have required it to be modified which would make it harder to keep up with any threejs modifications.  It's a question of either letting it diverge completely or not... I could be easily convinced to go full ts also.

kmitcham (approved):

src/RayMarchedAtlasVolume.ts
```diff
@@ -1,37 +1,69 @@
...
 
     this.cube = new BoxGeometry(1.0, 1.0, 1.0);
     this.cubeMesh = new Mesh(this.cube);
     this.cubeMesh.name = "Volume";
 
+    this.boxHelper = new BoxHelper(this.cubeMesh, 0xffff00);
+
     this.cubeTransformNode = new Group();
     this.cubeTransformNode.name = "VolumeContainerNode";
+    // REINSTATE THIS TO SHOW BOUNDING BOX!
```
kmitcham, line 56: Maybe a TODO tag, if that is an idiom here.

kmitcham (approved):
My eyes did glaze over, and it isn't my first language, but nothing looked crazy.

babel.config.js
```diff
@@ -2,9 +2,6 @@ module.exports = {
   presets: [
     [
       "@babel/preset-env",
-      {
```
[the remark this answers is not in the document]
toloudis: could have sworn I had a reason for this but I put it back and nothing broke.  Will reinstate, thanks for spotting that!

PR #36, opened 2021-11-01
Feature/convert public/index.js to TypeScript

schoinh:
Resolves: [volume-viewer: convert public/index.js to TypeScript](https://aicsjira.corp.alleninstitute.org/browse/ANIMCELLGR-682)

What I did:
* Renamed public/index.js to index.ts and fixed all the new type errors that came up
    * Part of this involved installing the dat.gui package (and its types) with npm instead of embedding it in index.html
* Also fixed some ESlint errors
* Also changed some variable names to proper camelCase

How I checked that I didn't break things:
* Ran the test viewer in both normal and time series modes and verified that the app still behaves like normal without errors

Sorry this PR is kind of all over the place, but a lot of it is just variable renaming. 🙈 

- [x] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [x] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_
- [x] Provide description and context of changes.
- [ ] Provide relevant tests for your feature or bug fix.
- [x] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!


public/index.ts
```diff
@@ -11,12 +12,14 @@ import {
...
   SKY_LIGHT,
 } from "../src";
+import { State } from "./types";
+import { defaultImageInfo } from "../src/Volume";
 
-let view3D = null;
+let view3D: View3d;
 
-const myState = {
+const myState: State = {
```
schoinh, line 20: I thought about renaming `myState` as `state` but it was going to be way too many more changes.

public/index.ts
```diff
@@ -28,7 +31,7 @@ const myState = {
   exposure: 0.75,
   aperture: 0.0,
   fov: 20,
-  focal_distance: 4.0,
+  focalDistance: 4.0,
```
schoinh, line 34: @toloudis I hope this wasn't one of those snake case names that depend on json data

src/types.ts
```diff
@@ -58,21 +58,40 @@ export interface VolumeChannelDisplayOptions {
...
+  maskChannelIndex?: number;
+  maskAlpha?: number;
+  clipBounds?: [number, number, number, number, number, number];
+  scale?: [number, number, number];
+  maxProjection?: boolean;
+  renderMode?: number;
+  shadingMethod?: number;
+  gamma?: [number, number, number];
+  primaryRayStepSize?: number;
+  secondaryRayStepSize?: number;
```
schoinh, line 74: Checked with @toloudis that these can all be optional

tsconfig.json
```diff
@@ -16,9 +16,11 @@
...
   },
   "files": [
-    "src/index.ts"
+    "src/index.ts",
+    "public/index.ts"
   ],
   "include": [
-    "src/**/*"
+    "src/**/*",
+    "public/**/*"
```
schoinh, line 24: Question: What's the reason for having both of these? Is it not sufficient to just have the `include`?

src/Volume.ts
```diff
@@ -27,7 +28,34 @@ export interface ImageInfo {
...
+  tile_width: 1,
+  tile_height: 1,
+  atlas_width: 1,
+  atlas_height: 1,
+  transform: {
+    translation: [0, 0, 0],
+    rotation: [0, 0, 0],
+  },
+  channelGui: [],
+});
```
schoinh, line 57: @toloudis came up with this default object

kmitcham (approved):

src/Volume.ts
toloudis: The only thing that may be a bit risky is the 0 channels.  If something is broken (which it sounds like it isn't) we could try making one single channel in there.

EricJIsaac (approved):

src/Volume.ts
```diff
@@ -112,32 +140,34 @@ export default class Volume {
...
+  constructor(imageInfo?: ImageInfo) {
     this.scale = new Vector3(1, 1, 1);
     this.currentScale = new Vector3(1, 1, 1);
     this.physicalSize = new Vector3(1, 1, 1);
     this.normalizedPhysicalSize = new Vector3(1, 1, 1);
 
     this.loaded = false;
-    this.imageInfo = imageInfo;
-    this.name = imageInfo.name;
+    this.imageInfo = imageInfo || defaultImageInfo;
```
toloudis, line 150: is there a typescript way to say that defaultImageInfo is the defaulted constructor parameter?
`constructor(imageInfo: ImageInfo = defaultImageInfo)`
and then you can just do `this.imageInfo = imageInfo`
Not a big difference I guess...

toloudis: I'm still slightly concerned about the Object.freeze for the default parameter.  It would be much safer if we just instantiated a new one every time.

tsconfig.json
toloudis: I honestly don't know the answer without reading the typescript compiler docs

src/types.ts
```diff
@@ -58,21 +58,40 @@ export interface VolumeChannelDisplayOptions {
...
+  window: number;
+  level: number;
+  glossiness: number;
+  isovalue: number;
+  isosurface: boolean;
+  enabled: boolean,
+  autoIJ: (channelNum: number) => void;
+  auto0: (channelNum: number) => void;
+  bestFit: (channelNum: number) => void;
+  pct50_98: (channelNum: number) => void;
```
toloudis, line 93: this `pct50_98` needs a linter disable comment.. also, if this `ChannelGuiOptions` is only used in the public/ demo viewer, then it doesn't need to be under `src/`

src/Volume.ts
```diff
@@ -27,7 +28,34 @@ export interface ImageInfo {
     translation: [number, number, number];
     rotation: [number, number, number];
   };
+  channelGui: ChannelGuiOptions[];
```
toloudis, line 31: I suspect this is a hack that is only here for the demo viewer and probably shouldn't pollute the ImageInfo.  I think the channelGui objects could be stored entirely in the "state" object in the demo viewer.  You could do a quick check in the website-3d-cell-viewer codebase for any use of `channelGui`.  I doubt you'll find it...

public/index.ts
toloudis: all the names of properties in the state object of the test app should definitely not be part of any exported interfaces.  should all be local to that app only.

schoinh:
@toloudis I think I've addressed all your comments, let me know what you think

public/types.ts
```diff
@@ -54,4 +54,25 @@ export interface State {
 
   channelFolderNames: string[];
   infoObj: ImageInfo;
+  channelGui: ChannelGuiOptions[];
+}
+
+interface ChannelGuiOptions {
```
toloudis, line 60: can this interface move into the example app now?

src/Volume.ts
```diff
@@ -140,22 +139,20 @@ export default class Volume {
   private pixel_size: [number, number, number];
   /* eslint-enable @typescript-eslint/naming-convention */
 
-  constructor(imageInfo?: ImageInfo) {
+  constructor(imageInfo: ImageInfo = getDefaultImageInfo()) {
```
toloudis, line 142: are you sure that `getDefaultImageInfo` gets called every time you do a `new Volume()` ?

schoinh: Okay, looks like we're safe: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#evaluated_at_call_time

toloudis: great! thank you for checking that.. I was definitely thinking of the Python thing I had read recently.

toloudis (approved):

PR #37, opened 2021-11-05
Feature/draw bounding box

toloudis:
Add api for showing bounding box lines and setting their color.

This remains unimplemented in Path Trace mode, and the bounds will not show up.

In order to position them correctly without breaking the "align" functionality, I had to modify the way transforms are applied to the volume data.  Some changes are to ensure that all three of path trace, ray march, and isosurface meshes are displayed with the same transforms. 

Lastly, there is a bonus bug fix for switching between path trace and ray march modes when clipping is applied.  (Currently if you apply axis clipping and then switch modes, your clipping gets reset.)

AICS 3D Volume Viewer 2021-11-05 14-39-21

### DIFF
--- a/src/PathTracedVolume.ts
+++ b/src/PathTracedVolume.ts
@@ -402,8 +402,8 @@ export default class PathTracedVolume {
     // apply volume translation and rotation:
     // rotate camera.up, camera.direction, and camera position by inverse of volume's modelview
     const m = new Matrix4().makeRotationFromQuaternion(new Quaternion().setFromEuler(this.rotation).invert());
-    mypos.applyMatrix4(m);
     mypos.sub(this.translation);
+    mypos.applyMatrix4(m);
     myup.applyMatrix4(m);
     mydir.applyMatrix4(m);
 
--- a/src/PathTracedVolume.ts
+++ b/src/PathTracedVolume.ts
@@ -803,6 +803,7 @@ export default class PathTracedVolume {
     }
   }
 
+  // 0..1 ranges as input
   updateClipRegion(xmin: number, xmax: number, ymin: number, ymax: number, zmin: number, zmax: number): void {
     this.bounds = {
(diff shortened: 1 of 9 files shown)


schoinh (approved):
I tried it out, and it's really neat (and probably useful) to see the bounding box when clipping the axes!

src/View3d.ts
```diff
@@ -232,6 +232,20 @@ export class View3d {
...
+    }
+    this.redraw();
+  }
+
+  setBoundingBoxColor(volume: Volume, color: [number, number, number]): void {
+    if (this.image) {
+      this.image.setBoundingBoxColor(color);
+    }
+    this.redraw();
+  }
```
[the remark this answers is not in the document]
toloudis: It was an attempt to distinguish between api calls that modify some aspect of the volume being displayed versus parameters global to the viewer/rendering system.  Plus: In case there are ever multiple volumes loaded at the same time, this would also be used in that scenario.

This whole View3d public interface needs to be looked at closely and brainstormed for something cleaner/simpler/more obvious.

kmitcham (approved):

src/PathTracedVolume.ts
```diff
@@ -354,6 +354,14 @@ export default class PathTracedVolume {
     // this.visible = isVisible;
   }
 
+  public setShowBoundingBox(_showBoundingBox: boolean): void {
+    // TODO: NOT IMPLEMENTED YET
+  }
+
+  public setBoundingBoxColor(_color: [number, number, number]): void {
+    // TODO: NOT IMPLEMENTED YET
```
EricJIsaac, line 362: Are these functions needed (part of the API) or just set as reminders?

EricJIsaac (approved):

src/PathTracedVolume.ts
toloudis: Yep, they are needed as interface methods.

PR #38, opened 2021-12-21
website-3d-cell-viewer wants to import these

toloudis:
some work on converting website-3d-cell-viewer to typescript would be nice to have these exports from volume-viewer.  stopgap until we merge the two repositories.

### DIFF
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -29,12 +29,12 @@ const LUT_ARRAY_LENGTH = LUT_ENTRIES * 4;
  * (maps scalar intensity to a rgb color plus alpha, with each value from 0-255)
  * @property {Array.<ControlPoint>} controlPoints
  */
-type ControlPoint = {
+export type ControlPoint = {
   x: number;
   opacity: number;
   color: [number, number, number];
 };
-type Lut = {
+export type Lut = {
   lut: Uint8Array;
   controlPoints: ControlPoint[];
 };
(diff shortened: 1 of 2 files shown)


toloudis:
What is the proper way to export types?  This works in the type checker for the client code but I get warnings like 
```
WARNING in ../volume-viewer/es/index.js 9:0-167
export 'ImageInfo' (reexported as 'ImageInfo') was not found in './Volume' (possible exports: default, getDefaultImageInfo)
```

schoinh:
> What is the proper way to export types? This works in the type checker for the client code but I get warnings like
> 
> ```
> WARNING in ../volume-viewer/es/index.js 9:0-167
> export 'ImageInfo' (reexported as 'ImageInfo') was not found in './Volume' (possible exports: default, getDefaultImageInfo)
> ```

Looking at [simularium-viewer](https://github.com/simularium/simularium-viewer/blob/main/src/index.ts#L4)... have you tried this:
```javascript
export interface { ImageInfo } from "./Volume"; // or maybe export type instead of interface?
```
instead of importing `ImageInfo` into src/index.ts then exporting it again?

toloudis:
> Looking at [simularium-viewer](https://github.com/simularium/simularium-viewer/blob/main/src/index.ts#L4)... have you tried this:
> 
> ```js
> export interface { ImageInfo } from "./Volume"; // or maybe export type instead of interface?
> ```

Thanks, that's exactly what I needed.

schoinh (approved):

PR #49, opened 2022-05-18
add fuse on gpu

toloudis:
To draw the multichannel volumes efficiently, I pre-combine the channels into a single object using an operation named "fuse".   The fuse takes the max value of r,g,b,a over all channels, for every pixel in the volume.

The Fuse operation was implemented on CPU with web workers, but can be immensely optimized by moving it to gpu.

I implement it using "Max" blending mode, so that all I have to do is draw each channel one by one and apply the RGBA lut to each intensity (in fuseShader.ts).  The max blending takes care of the rest.

Because this is no longer asynchronous in a web worker, the redraw logic had to be changed.  Basically any time something changes which could require a fuse, we make sure the fuse is requested and then ask for a redraw.  In the regular redraw logic, the fuse-on-gpu will be added.

One key resource usage change is that now each Channel has a Texture object that holds the gpu texture containing the channel's data, and a second Texture object to hold the intensity-to-rgba lookup table.


src/FusedChannelData.ts
```diff
@@ -58,128 +61,61 @@ export default class FusedChannelData {
...
+        },
+      ]),
+      vertexShader: fuseVertexShaderSrc,
+      fragmentShader: fuseShaderSrc,
+      depthTest: false,
+      depthWrite: false,
+      blending: CustomBlending,
+      blendSrc: OneFactor,
+      blendDst: OneFactor,
+      blendEquation: MaxEquation,
```
toloudis, line 104: here is where we set the "max" blend operation

src/FusedChannelData.ts
```diff
@@ -198,110 +134,37 @@ export default class FusedChannelData {
...
-          fused[fx + 3] = Math.max(aa, la);
-        }
+    // webgl draw one mesh per channel to fuse.  clear texture to 0,0,0,0
+    this.fuseScene.clear();
+    for (let i = 0; i < combination.length; ++i) {
+      if (combination[i].rgbColor) {
+        const chIndex = combination[i].chIndex;
+        // add a draw call per channel here.
+        // TODO create these at channel creation time!
+        const mat = this.fuseMaterial.clone();
```
toloudis, line 155: these clone operations are relatively cheap and don't affect runtime performance in any nontrivial way but still could be optimized away by pre-creating them

kmitcham (approved):

GabeMedrash (approved):
LGTM, FWIW

PR #54, opened 2022-08-02
feature/add background color setter

frasercl:
Adds the method `view3d.setBackgroundColor`.

This function both calls `renderer.setClearColor` (tied to the already existing `view3d.backgroundColor` property but in practice seems to only matter for the first frame) and sets the canvas element's `style.backgroundColor`. Should it do only one or the other, or something else?

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -66,6 +66,7 @@ const myState: State = {
 
   showBoundingBox: false,
   boundingBoxColor: [255, 255, 0],
+  backgroundColor: [0, 0, 0],
   flipX: 1,
   flipY: 1,
   flipZ: 1,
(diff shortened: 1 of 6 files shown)


toloudis:
Could you add a button to the public/index.html and code in public/index.ts so it's testable in the demo app in here?

toloudis:
My guess is that it's more important to set the renderer's clear color.  I believe the canvas style has more to do with compositing with of other html content and generally this component is probably preferable to not composite over anything else anyway.  The threejs webglrenderer will fill the whole canvas with the clear color.

On the other hand, it's probably safe to keep both in and won't hurt anything.

toloudis:
On Chrome the bg color change is not kicking in right away.  The setter might need to force a redraw.

frasercl:
Testing control added, and you're right: background color updates on the next redraw. Removed the style setting from the function, it's cleaner that way and I think will be less likely to cause problems with rapid calls, e.g. live previewing changes.

src/View3d.ts
```diff
@@ -243,6 +243,12 @@ export class View3d {
     this.redraw();
   }
 
+  setBackgroundColor(color: [number, number, number]): void {
+    const hexColor = new Color().fromArray(color).getHex();
+    this.backgroundColor = hexColor;
+    this.canvas3d.renderer.setClearColor(hexColor, 1);
```
toloudis, line 249: add `this.redraw();`

toloudis:
Pathtrace mode also needs some change to pick up the bg color.

frasercl:
Hmm... setting the element's style solves this, if inelegantly. I don't know how much knowledge and work it might take to make the pathtrace renderer not make the element transparent. A further complication is that switching back out of pathtrace appears to set the clear color to transparent (at `FusedChannelData.ts:159`), and this creates graphical problems if removed.

public/index.ts
```diff
@@ -1033,19 +1034,24 @@ function main() {
     myState.showBoundingBox = !myState.showBoundingBox;
     view3D.setShowBoundingBox(myState.volume, myState.showBoundingBox);
   });
+
+  // convert value to rgb array
+  function hexToRgb(hex, last: [number, number, number]): [number, number, number] {
```
meganrm, line 1039: I know you didn't actually add this (just moved it), but I'm surprised we don't already have this function somewhere in this repo. Doesn't `Color` have some built in function fo this?

meganrm (approved):

toloudis:
> Hmm... setting the element's style solves this, if inelegantly. I don't know how much knowledge and work it might take to make the pathtrace renderer not make the element transparent. A further complication is that switching back out of pathtrace appears to set the clear color to transparent (at `FusedChannelData.ts:159`), and this creates graphical problems if removed.

I'll take a look at pathtrace mode too.  If interested, we can pair on this and dig into the pathtrace version.  If it proves to be too horrible, we can punt on it and use the canvas css background.

public/index.ts
toloudis: Yes, this app layer code could do the conversion in other ways, such as farming it out to threejs. 

We are putting the responsibility for this conversion outside of the main volume viewer, because the View3d interface wants a [number, number, number] for rgb color settings.  And we are avoiding having any explicit threejs type dependencies in the public api for View3d.  So the apps that embed this viewer have to do whatever color format conversions are needed.

toloudis (approved):

PR #57, opened 2022-08-05
Feature/nightly build

toloudis:
Does a nightly build using github-actions, and publishes it to github-pages at allen-cell-animated.github.io/volume-viewer/
The webpack config was modified, so there is no longer a `volumeviewer` subdirectory when running in dev-server.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -3,7 +3,9 @@ name: github pages
 env:
   NODE_VERSION: "14.x"
 
-on: [push]
+on:
+  schedule:
+    - cron: "0 0 * * *"
 
 jobs:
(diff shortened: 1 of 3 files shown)


frasercl (approved):
GitHub workflows are new to me, but from what I'm able to learn this looks good. I am curious though: why deploy a development build rather than production?

toloudis:
> GitHub workflows are new to me, but from what I'm able to learn this looks good. I am curious though: why deploy a development build rather than production?

Good question and I think it's just a matter of misleading file names.  The webpack.dev.js is a configuration for the test app, which is what would typically be used during dev to make sure things are working.   And that test app is the thing we want to deploy to github-pages.

PR #64, opened 2022-09-06
Fix bug with multiple viewers in same page

toloudis:
This change set was driven by a bug when multiple viewers are embedded on the same page.  Only one 3d view could be manipulated at a time and the other would go blank.  

This change set turned into a small mixture of bug fixes:

* While debugging, I discovered that the canvas and its parent div were getting html id attributes set. All html elements should either have unique ids or no id.  In the case where we have multiple viewers on the page, some of these elements would have had the same ids.

* I also was (needlessly) suspicious of threejs corrupting global state.  I went ahead and updated its version to latest.  While doing that, I updated our modified TrackballControls to stay as close as possible to the threejs source code.

* Along the way I also found a minor bug with "reset camera" while the Axis Helper is being displayed.  That bug is fixed here.

* It turned out that the key bug was probably introduced when I implemented "fuse on gpu".  The result texture needed to be updated more proactively and that resolved the issue of secondary viewers going blank.

### DIFF
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -123,6 +123,7 @@ export default class RayMarchedAtlasVolume {
     }
 
     this.channelData.gpuFuse(canvas.renderer);
+    this.setUniform("textureAtlas", this.channelData.getFusedTexture());
 
     this.cubeTransformNode.updateMatrixWorld(true);
 
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -427,6 +427,13 @@ export class ThreeJsPanel {
   }
 
   render(): void {
+    // update the axis helper in case the view was rotated
+    if (!isOrthographicCamera(this.camera)) {
+      this.axisHelperObject.rotation.setFromRotationMatrix(this.camera.matrixWorldInverse);
+    } else {
+      this.orthoScale = this.controls.scale;
+    }
+
     // do whatever we have to do before the main render of this.scene
     for (let i = 0; i < this.animateFuncs.length; i++) {
       if (this.animateFuncs[i]) {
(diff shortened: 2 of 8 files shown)


kmitcham (approved):

src/RayMarchedAtlasVolume.ts
```diff
@@ -123,6 +123,7 @@ export default class RayMarchedAtlasVolume {
     }
 
     this.channelData.gpuFuse(canvas.renderer);
+    this.setUniform("textureAtlas", this.channelData.getFusedTexture());
```
toloudis, line 126: this is the fix for the key bug that started these changes

src/ThreeJsPanel.ts
```diff
@@ -62,11 +62,9 @@ export class ThreeJsPanel {
 
   constructor(parentElement: HTMLElement, _useWebGL2: boolean) {
     this.containerdiv = document.createElement("div");
-    this.containerdiv.setAttribute("id", "volumeViewerContainerDiv");
     this.containerdiv.style.position = "relative";
 
     this.canvas = document.createElement("canvas");
-    this.canvas.setAttribute("id", "volumeViewerCanvas");
```
toloudis, line 69: removed ids that are not unique

src/ThreeJsPanel.ts
```diff
@@ -429,6 +427,13 @@ export class ThreeJsPanel {
   }
 
   render(): void {
+    // update the axis helper in case the view was rotated
+    if (!isOrthographicCamera(this.camera)) {
+      this.axisHelperObject.rotation.setFromRotationMatrix(this.camera.matrixWorldInverse);
+    } else {
+      this.orthoScale = this.controls.scale;
+    }
+
```
toloudis, line 436: changes in this file fix issue with axis helper not moving when reset camera happens.

frasercl (approved):

Issue #68, opened 2022-10-05
export Channel type


toloudis:
## Use Case

website-3d-cell-viewer has code that holds on to Channels but doesn't have the type definition. 

## Solution

Export type Channel so that downstream apps can see it.

## Alternatives

Consider WHY the downstream app needs to have a Channel - could it not?

frasercl:
Resolved by #69.

PR #69, opened 2022-10-05
Fix/include type declarations in build

frasercl:
This package is not currently configured to include type declaration files on build - users can import its classes, but they all implicitly have the `any` type. This PR adds a build step to generate `.d.ts` files and configures `package.json` to point to those declarations.

Also:
- Builds no longer include test files
- Changes source path in `transpileES` from `'src/'` to `src`. The former wasn't working for me, but it's possible this could break things on other OS'es?


frasercl:
Might as well also resolve #68 here

babel.config.js
```diff
@@ -23,4 +23,7 @@ module.exports = {
       },
     ],
   ],
+  ignore: [
+    "**/*.test.js",
+  ],
```
toloudis, line 28: is this ignore redundant with the one in package.json's transpileES command line?

frasercl: That line was seemingly not working for me. Removing single quotes around the path got it working again, as with `src`.

toloudis (approved):
LGTM, thanks!   I don't know about quotes around the dir names, we'll see if the build fails.  Another possibility is to use escaped double quotes `\"src\"` ... whatever works.  I guess the github build runner is linux but I also develop for this lib in macos too.

frasercl:
No quotes works fine on WSL Ubuntu. Gonna call that a conclusive enough test and merge.

PR #70, opened 2022-10-10
Fix/type tweaks

frasercl:
Fixing some type errors that appear in website-3d-cell-viewer when types are exported properly from here (#69) which I thought indicated problems with volume-viewer types rather than cell-viewer ones. Including:
- Change axis types from numeric indices to letters (`xyz`) to match three's `Vector3` type
- Make all properties of `VolumeChannelDisplayOptions` optional (matches `VolumeDisplayOptions`)
- Publicize any private fields that 3d-cell-viewer was accessing anyways. NOTE: I did this uncritically to just flag that the cell viewer was doing this. Changing `private` to `public` may not be the best final answer for a couple of these.

### DIFF
--- a/src/MeshVolume.ts
+++ b/src/MeshVolume.ts
@@ -119,7 +119,7 @@ export default class MeshVolume {
     // no op
   }
 
-  setAxisClip(axis: number, minval: number, maxval: number, _isOrthoAxis: boolean): void {
+  setAxisClip(axis: "x" | "y" | "z", minval: number, maxval: number, _isOrthoAxis: boolean): void {
     this.bounds.bmax[axis] = maxval;
     this.bounds.bmin[axis] = minval;
     this.updateClipFromBounds();
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -507,7 +507,7 @@ export class View3d {
    * @param {number=} oh Ignored.
    * @param {Object=} eOpts Ignored.
    */
-  resize(comp: HTMLElement | null, w: number, h: number, ow?: number, oh?: number, eOpts?: unknown): void {
+  resize(comp: HTMLElement | null, w?: number, h?: number, ow?: number, oh?: number, eOpts?: unknown): void {
     w = w || this.parentEl.offsetWidth;
     h = h || this.parentEl.offsetHeight;
     this.canvas3d.resize(comp, w, h, ow, oh, eOpts);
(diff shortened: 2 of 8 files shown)


src/MeshVolume.ts
```diff
@@ -119,7 +119,7 @@ export default class MeshVolume {
     // no op
   }
 
-  setAxisClip(axis: number, minval: number, maxval: number, _isOrthoAxis: boolean): void {
+  setAxisClip(axis: "x" | "y" | "z", minval: number, maxval: number, _isOrthoAxis: boolean): void {
```
toloudis, line 122: Did you make sure that the bounds.bmin and bounds.bmax are types that are keyed with those letters?

src/View3d.ts
```diff
@@ -507,7 +507,7 @@ export class View3d {
    * @param {number=} oh Ignored.
    * @param {Object=} eOpts Ignored.
    */
-  resize(comp: HTMLElement | null, w: number, h: number, ow?: number, oh?: number, eOpts?: unknown): void {
+  resize(comp: HTMLElement | null, w?: number, h?: number, ow?: number, oh?: number, eOpts?: unknown): void {
```
toloudis, line 510: I wonder if the signature for this could just be something like HTMLResizeEvent... Is this function directly invoked from window.addEventListener('resize'...) ?

src/VolumeDrawable.ts
```diff
@@ -199,43 +199,43 @@ export default class VolumeDrawable {
     // merge to current channel options
     this.channelOptions[channelIndex] = Object.assign(this.channelOptions[channelIndex], options);
 
-    if (Object.hasOwnProperty.call(options, "enabled")) {
+    if (options.enabled !== undefined) {
```
toloudis, line 202: I can't remember but I thought we had a good reason for using hasOwnProperty instead of just checking undefined. oh well, this seems ok really.

toloudis (approved):
You are right to be nervous about changing private to public.   It's generally bad practice ... but this api is probably due for a global pass to make sure everything that is public actually should be.
My rules of thumb: 
1. start private and allow public if must
2. keep implementation details private and hide them with public functions that have less implementation-specific names, when possible.

toloudis:
I named a couple things to double check but the changes look ok to me!

src/View3d.ts
frasercl: Looks like it goes through an arrow function on that event listener (View3d.ts:62). cell-viewer was calling it without arguments and creating a type error; I added `null` to fill in that first argument, but am thinking now that maybe I should just have made all arguments optional.

src/VolumeDrawable.ts
frasercl: Problem here was that, with all fields of `VolumeChannelDisplayOptions` optional, `hasOwnProperty` didn't guard types properly so that TypeScript knew the fields weren't `undefined`. Looks like `setOptions` just above had already been using that syntax anyways.

src/MeshVolume.ts
frasercl: Yep, `Bounds` is two `Vector3`s

PR #71, opened 2022-11-01
clean up zarr dimensions to support more data from brain science

toloudis:
Modify the way zarr dimensions are handled to be more local to the multiresolution layer being loaded.
The current strategy only loads the lowest resolution but was trying to do what we do with other downsampled data where we try to preserve the full pixel dimensions for display.  The catch was that AICS data doesn't typically downsample in Z and so things were getting inconsistent with data that WAS downsampled in Z. 

The fix is to just disregard the original dimensions and fully use the dimensions of the downsampled level being loaded.

### DIFF
--- a/src/VolumeLoader.ts
+++ b/src/VolumeLoader.ts
@@ -193,10 +193,6 @@ export default class VolumeLoader {
     const imageIndex = 0;
     // there is one dataset for each multiscale level.
     const dataset0 = allmetadata.multiscales[imageIndex].datasets[0];
-    // technically there can be any number of coordinateTransformations
-    // but there must be only one of type "scale".
-    // Here I assume that is the only one.
-    const scale5d = dataset0.coordinateTransformations[0].scale;
 
     // TODO get metadata sizes for each level?  how inefficient is that?
     // update levelToLoad after we get size info about multiscales?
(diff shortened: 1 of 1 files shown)


src/VolumeLoader.ts
```diff
@@ -205,25 +201,29 @@ export default class VolumeLoader {
 
     const level0 = await openArray({ store: store, path: imagegroup + "/" + dataset0.path, mode: "r" });
     // full res info
-    const w = level0.meta.shape[4];
-    const h = level0.meta.shape[3];
-    const z = level0.meta.shape[2];
+    // const w = level0.meta.shape[4];
```
kmitcham, line 204: comment out, not delete?

kmitcham (approved):

src/VolumeLoader.ts
toloudis: I knew you were going to say that. Agree in principle. Currently this serves as documentation on how to get these dimensions and will almost surely be reinstated later (soon) as we get the viewer handling multiresolution data better.

kmitcham: If you knew I was gonna say it, you could have added that comment in the code, and saved us both a trip.  :)

PR #72, opened 2022-11-03
Fix/max project brightness

### LINKED ISSUE #53: max project mode is too dark compared to raymarch mode
## Description

When switching from regular raymarch volume rendering to max-project mode, the max projection mode often is too dark to see anything.

## Expected Behavior

Max project mode with same density and brightness settings should appear similar in color and brightness to the raymarch mode.

frasercl:
Resolve #53 (?)

Modify behavior of ray march shader when in max project mode to:
- Not multiply color components (rgb) by alpha while accumulating colors
- Not apply `luma2Alpha` to colors before accumulating

Additionally, converts shader code in `volumeRayMarchShader.ts` from an array of strings to a single template literal. NOTE this makes the diff challenging to read - review the second commit only for actual substantive code changes.

### DIFF
--- a/src/constants/volumeRayMarchShader.ts
+++ b/src/constants/volumeRayMarchShader.ts
@@ -165,14 +165,6 @@ vec4 accumulate(vec4 col, float s, vec4 C) {
   return C;
 }
 
-vec4 accumulateMax(vec4 col, float s, vec4 C) {
-  if (col.x > C.x) { C.x = col.x; }
-  if (col.y > C.y) { C.y = col.y; }
-  if (col.z > C.z) { C.z = col.z; }
-  if (col.w > C.w) { C.w = col.w; }
-  return C;
-}
-
 vec4 integrateVolume(vec4 eye_o,vec4 eye_d,
                      float tnear,   float tfar,
                      float clipNear, float clipFar,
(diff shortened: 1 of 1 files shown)


src/constants/volumeRayMarchShader.ts
```diff
@@ -1,297 +1,293 @@
...
+
+  C = (1.0-C.w)*col + C;
+  return C;
+}
+
+vec4 accumulateMax(vec4 col, float s, vec4 C) {
+  if (col.x > C.x) { C.x = col.x; }
+  if (col.y > C.y) { C.y = col.y; }
+  if (col.z > C.z) { C.z = col.z; }
+  if (col.w > C.w) { C.w = col.w; }
```
toloudis, line 172: looking at this now, I think you may be able to do 
```
return max(col, C);
```
I believe glsl does a component-wise maximum that way and it has the possibility to be vectorized in the hardware.

src/constants/volumeRayMarchShader.ts
```diff
@@ -1,297 +1,293 @@
...
+
+    col.xyz *= BRIGHTNESS;
+
+    if (maxProject != 0) {
+      C = accumulateMax(col, s, C);
+    } else {
+      // for practical use the density only matters for regular volume integration
+      col.w *= DENSITY;
+      C = accumulate(col, s, C);
+    }
```
toloudis, line 229: small micro-optimization:  combine the if(maxProject==0) on line 217 with the if(maxProject!=0) statement on lines 223-229 so there is only a single if conditional. something like this?
```
if (maxProject == 0) {
      col = luma2Alpha(col, GAMMA_MIN, GAMMA_MAX, GAMMA_SCALE);
      col.xyz *= BRIGHTNESS;
      // for practical use the density only matters for regular volume integration
      col.w *= DENSITY;
      C = accumulate(col, s, C);
    }
    else {
      col.xyz *= BRIGHTNESS;
      C = accumulateMax(col, s, C);
    } 
```

toloudis: honestly i'm not even sure if that would have much perf impact so i leave it up to you

frasercl: yeah I realized I ought to do that as soon as I PR'ed - for code cleanliness as well as performance

toloudis (approved):
github's side-by-side diff mode made this much more readable 👍

src/constants/volumeRayMarchShader.ts
frasercl: in that case, does it make sense to keep `accumulateMax` around at all or just write `C = max(col, C);` straight in the loop?

toloudis: yeah, you could put it directly in the loop without the wrapper function

PR #73, opened 2022-11-04
Feature/add max project button to test viewer

frasercl:
## Problem

When testing approaches for maximum intensity projections, I had no way of switching the test viewer from ray march mode to max project mode.

## Solution

Add a button that does this.

#### Screenshot:
image

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -18,6 +18,7 @@
         <button id="xfBtn">Align</button>
         <button id="resetCamBtn">ResetCam</button>
         <button id="ptBtn" disabled>Pathtrace</button>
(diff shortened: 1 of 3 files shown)


toloudis (commented):
to see if i understand:  
click pathtrace --> go into pathtrace mode if not already, else raymarch
click max project --> go into max project mode if not already, else raymarch 
?

my first thought is to just make a 3 item dropdown menu but this is ok too

frasercl:
That would make some sense - I'll try that out

toloudis (approved):

PR #76, opened 2022-11-17
Feature/disable interpolation

### LINKED ISSUE #62: allow nearest-neighbor voxel interpolation
## Use Case

"the option to toggle voxel interpolation like in napari" - voxels either blend/interpolate into their neighbors or they have hard edges.

## Solution

A render setting toggle switch.  
implementation: Volume Texture sampling switches from linear to nearest.  There is part of one shader that will probably have to do some different math too (new uniform var to configure it).

frasercl:
Resolve #62: add "the option to toggle voxel interpolation like in napari." (In ray march mode - would disabling interpolation in pathtrace mode mean anything or make sense?) Via:
- new shader math to do nearest-neighbor interpolation, with a new uniform to switch to that code
  - also, a tweak to ray dithering offsets
- new method `setInterpolationEnabled` to change interpolation behavior
- new button in the test viewer to enable/disable smooth interpolation

And a couple stray modifications:
- shrink a couple arrow functions
- give `rayMarchedAtlasVolume.setUniform` a solid type signature

## Screenshots
image

(from website-3d-cell-viewer)
image
image

I also attempted to replicate the same image of a cell membrane in both website-3d-cell-viewer and Fiji, to verify that the shader math is correct and samples pure pixel values. Fiji is first, 3d viewer is second:

695296_fiji
695296_3dv

### DIFF
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -194,8 +194,8 @@ export default class RayMarchedAtlasVolume {
     }
   }
 
-  public setInterpolationActive(active: boolean): void {
-    this.setUniform("doInterpolation", active);
+  public setInterpolationEnabled(active: boolean): void {
+    this.setUniform("interpolationEnabled", active);
   }
 
   public viewpointMoved(): void {
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -194,6 +194,10 @@ export default class RayMarchedAtlasVolume {
     }
   }
 
+  public setInterpolationActive(active: boolean): void {
+    this.setUniform("doInterpolation", active);
+  }
+
   public viewpointMoved(): void {
     // no op
   }
@@ -258,7 +262,10 @@ export default class RayMarchedAtlasVolume {
   //////////////////////////////////////////
   //////////////////////////////////////////
 
-  private setUniform(name, value) {
+  private setUniform<U extends keyof typeof rayMarchingShaderUniforms>(
+    name: U,
+    value: typeof rayMarchingShaderUniforms[U]["value"]
+  ) {
     if (!this.uniforms[name]) {
       return;
     }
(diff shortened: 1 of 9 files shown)


src/constants/volumeRayMarchShader.ts
```diff
@@ -184,7 +206,7 @@ vec4 integrateVolume(vec4 eye_o,vec4 eye_d,
   float invstep = (tfar-tnear)/csteps;
   // special-casing the single slice to remove the random ray dither.
   // this removes a Moire pattern visible in single slice images, which we want to view as 2D images as best we can.
-  float r = (SLICES==1.0) ?  0.0 : 0.5 - 1.0*rand(eye_d.xy);
+  float r = (SLICES==1.0) ? 0.0 : rand(eye_d.xy);
```
toloudis, line 209: so the difference here is that the ray offsets are always pushed in to the volume edge instead of distributed half in and half out.  theoretically they are scaled down enough that they shouldn't substantially affect the overall brightness at the edges but just want to be aware that that is what's changing here.

src/RayMarchedAtlasVolume.ts
```diff
@@ -258,7 +262,10 @@ export default class RayMarchedAtlasVolume {
   //////////////////////////////////////////
   //////////////////////////////////////////
 
-  private setUniform(name, value) {
+  private setUniform<U extends keyof typeof rayMarchingShaderUniforms>(
+    name: U,
+    value: typeof rayMarchingShaderUniforms[U]["value"]
+  ) {
```
toloudis, line 268: this is real nice

src/constants/volumeRayMarchShader.ts
frasercl: There is an effect in no-interpolation mode where some voxels appear grainy when the volume is viewed edge-on, as ray steps per voxel becomes very small and some rays end up missing some voxels entirely (I tried to capture this a bit in the second screenshot above). I didn't investigate to my satisfaction if that was improved when this offset was shifted back by a half, but I did try shrinking the offset by some constant factor, e.g. `rand(eye_d.xy)*0.5`, and saw the moire pattern reappear pretty quickly.

toloudis:
I think there might be still one more thing to do here. When texture2d is called in the shader, the gpu filters the lookup according to filtering settings on the texture that are set in the JavaScript. Do a search for LINEAR or NEAREST in the code.  It might be set at texture creation time but it can also be set later.  This could also be the way to make it work for pathtrace mode too.

toloudis (approved):

PR #80, opened 2022-12-14
Feature/zarr multiscales

toloudis:
Two improvements to zarr loading:
1. auto select the largest multiscale that will fit tiled into a 4096x4096 texture atlas
2. allow loading data that doesn't have a T or C dimension.

For zarr loading, the viewer still assumes TCZYX as standard dimension names, and expects data to at least have ZYX as the last 3 dims.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -1241,7 +1241,7 @@ function main() {
     myState.isMP = mp;
     view3D.setVolumeRenderMode(pt ? RENDERMODE_PATHTRACE : RENDERMODE_RAYMARCH);
     view3D.setMaxProjectMode(myState.volume, mp);
-  }
+  };
   renderModeSelect?.addEventListener("change", ({ currentTarget }) => {
     if ((currentTarget as HTMLOptionElement)!.value === "PT") {
       if (view3D.hasWebGL2()) {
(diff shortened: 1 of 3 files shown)


public/index.ts
```diff
@@ -1283,7 +1283,9 @@ function main() {
   } else if (loadTestData) {
     //fetchOpenCell();
     //fetchTiff("https://animatedcell-test-data.s3.us-west-2.amazonaws.com/AICS-12_881.ome.tif", 0);
-    fetchZarr("https://animatedcell-test-data.s3.us-west-2.amazonaws.com/Lamin_multi-06-Deskew-28.zarr", "Image_0", 0);
+    //fetchZarr("https://animatedcell-test-data.s3.us-west-2.amazonaws.com/Lamin_multi-06-Deskew-28.zarr", "Image_0", 0);
```
kmitcham, line 1286: Things we know we shouldn't do, but are much to useful to NOT do.

kmitcham (approved):

public/index.ts
toloudis: At this point it would be nicer to just stash these as an array of urls and let the code select one rather than using commenting like this.  Maybe that would look friendlier.

toloudis: I added #81 to improve this soon.

Issue #66, opened 2022-09-23
show scale bar


toloudis:
## Use Case

Users may be interested to know the physical spatial dimensions of the volume.

## Solution

In 2d modes, can show scale bar as a ruler in the corner.   Ruler value should be updated with zoom in / out.
In 3d perspective view, scale bar could be represented as tickmarks on bounding box lines, along with a ruler in the corner.  Needs a bit of spec to work out details, probably.

Implementation detail:  can the ruler be html that floats above canvas?  that would probably be ideal rather than rendering webgl text for the numbers.

toloudis:
This task could/should be split into two, one for 2D and one for 3D modes since they will be very different.

toloudis:
need to add unit support to Volume object

toloudis:
TBD: where to put the ruler on screen?  @lynwilhelm

toloudis:
Is this properly put in volume-viewer as opposed to website-3d-cell-viewer?   There may be some interplay between the two.  Should there be a show/hide rulers toggle?

frasercl:
The ruler is currently in volume-viewer as an additional div element injected into `ThreeJsPanel.containerdiv`. I think it definitely makes sense to have the option to disable the ruler in the volume-viewer API, but I'm not sure if it's obtrusive enough to be worth adding UI to get rid of it, especially given that the 3d mode version might be somewhat entangled with the bounding box and its button.

Now may also be the time to address website-3d-cell-viewer#73 with some way to set a box for elements that live in corners, i.e. minimally `View3d.setInfoElementBox(box: [number, number, number, number]): void`. The scale bar, if placed right in the bottom-right corner, would land directly under the "clipping" button.

PR #82, opened 2023-01-06
Feature/orthographic scale bar

frasercl:
Partially address #66: in orthographic (XY/YZ/XZ) modes, where scale is the same regardless of distance from the camera, show a 2d scale bar in the corner of the viewport. The scale bar shows the length of some round number of physical units on screen (depending on zoom level) along with a unit symbol, which is either read in from volume data or set manually via methods.

#### Also:
- Make type of cameras more specific, removing the need for several explicit casts
- Ensure device pixel ratio is updated on resize when necessary (e.g. when moving between monitors of different resolutions)
- Remove unused `Volume.currentScale` property (always a copy of `Volume.scale`)
- vscode wanted to turn `VolumeLoader.loadJson` into an async function, so I let it

#### Work remaining on #66:
- Specify and implement what a scale indicator looks like in 3D mode: likely tickmarks on the bounding box, with a label either in the corner (as in Simularium) or next to the bounding box.
- Provide a means for clients to specify a new position for the scale bar, e.g. viewport corner + horizontal and vertical offset, since the scale bar in its current position would overlap the clipping drawer button in website-3d-cell-viewer. If this work added a means of moving the axis indicator similarly, this could open an easy path to resolving allen-cell-animated/website-3d-cell-viewer#73.

### Screenshots:
image
image

### DIFF
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -311,10 +311,13 @@ export class ThreeJsPanel {
     this.axisCamera.position.set(-this.axisOffset[0], -this.axisOffset[1], this.axisScale * 2.0);
   }
 
-  orthoScreenPixelsToPhysicalUnits(pixels: number, scale: number): number {
-    // at orthoScale = 0.5, the viewport is 1 world unit tall
+  orthoScreenPixelsToPhysicalUnits(pixels: number, physicalUnitsPerWorldUnit: number): number {
+    // At orthoScale = 0.5, the viewport is 1 world unit tall
     const worldUnitsPerPixel = this.orthoScale * 2 / this.getHeight();
-    return pixels * window.devicePixelRatio * worldUnitsPerPixel * scale;
+    // Multiply by devicePixelRatio to convert from scaled CSS pixels to physical pixels
+    // (to account for high dpi monitors, e.g.). We didn't do this to height above because
+    // that value comes from three, which works in physical pixels.
+    return pixels * window.devicePixelRatio * worldUnitsPerPixel * physicalUnitsPerWorldUnit;
   }
 
   setupOrthoScaleBar(): void {
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -36,7 +36,6 @@ export class ThreeJsPanel {
   public renderer: WebGLRenderer;
   private timer: Timing;
   public orthoScale: number;
-  public orthoHorizontalAxis: number;
   private fov: number;
   private perspectiveCamera: PerspectiveCamera;
   private perspectiveControls: TrackballControls;
@@ -59,6 +58,8 @@ export class ThreeJsPanel {
   private axisHelperObject: Object3D;
   private axisCamera: PerspectiveCamera | OrthographicCamera;
 
+  // Index of the horizontal axis in orthographic view modes, used for picking pixel scales for the scale bar
+  public orthoHorizontalAxis: number;
   private orthoScaleBarElement: HTMLDivElement;
   private scaleBarUnit?: string;
 
@@ -472,6 +473,11 @@ export class ThreeJsPanel {
       this.axisCamera.updateProjectionMatrix();
     } else {
       this.axisCamera.updateProjectionMatrix();
+    }
+
+    // The window may have moved to a monitor of a different resolution
+    if (this.renderer.getPixelRatio() !== window.devicePixelRatio) {
+      this.renderer.setPixelRatio(window.devicePixelRatio);
     }
 
     this.renderer.setSize(w, h);
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -29,7 +29,7 @@ export class ThreeJsPanel {
   private canvas: HTMLCanvasElement;
   public scene: Scene;
   private zooming: boolean;
-  public animateFuncs: ((ThreeJsPanel) => void)[];
+  public animateFuncs: ((panel: ThreeJsPanel) => void)[];
   private inRenderLoop: boolean;
   private requestedRender: number;
   public hasWebGL2: boolean;
@@ -58,7 +58,9 @@ export class ThreeJsPanel {
   private axisHelperObject: Object3D;
   private axisCamera: Camera;
 
-  private dataurlcallback?: (string) => void;
+  private orthoScaleBarElement: HTMLDivElement;
+
+  private dataurlcallback?: (url: string) => void;
 
   constructor(parentElement: HTMLElement, _useWebGL2: boolean) {
     this.containerdiv = document.createElement("div");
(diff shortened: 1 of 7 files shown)


src/ThreeJsPanel.ts
```diff
@@ -308,6 +311,55 @@ export class ThreeJsPanel {
     this.axisCamera.position.set(-this.axisOffset[0], -this.axisOffset[1], this.axisScale * 2.0);
   }
 
+  orthoScreenPixelsToPhysicalUnits(pixels: number, scale: number): number {
```
toloudis, line 314: for readability, rename `scale` to something like `physicalSizePerWorldUnit`?

src/ThreeJsPanel.ts
```diff
@@ -308,6 +311,55 @@ export class ThreeJsPanel {
...
+    // at orthoScale = 0.5, the viewport is 1 world unit tall
+    const worldUnitsPerPixel = this.orthoScale * 2 / this.getHeight();
+    return pixels * window.devicePixelRatio * worldUnitsPerPixel * scale;
```
toloudis, line 317: I would just add a comment explaining that this.getHeight already has the devicePixelRatio multiplied in because it's coming from the threejs renderer which sets things up that way by default.

src/ThreeJsPanel.ts
```diff
@@ -355,23 +407,27 @@ export class ThreeJsPanel {
         this.replaceCamera(this.orthographicCameraX);
         this.replaceControls(this.orthoControlsX);
         this.axisHelperObject.rotation.set(0, Math.PI * 0.5, 0);
+        this.setOrthoScaleBarVisible(true);
```
toloudis, line 410: always visible?  should we have some api to show/hide it? and eventually the inevitable button..

src/VolumeLoader.ts
```diff
@@ -28,6 +28,34 @@ interface PackedChannelsImage {
...
+  }
+
+  // SI prefixes not in unitSymbols are abbreviated by first letter, capitalized if prefix ends with "a"
+  if (unitName.endsWith("meter")) {
+    const capitalize = unitName[unitName.length - 6] === "a";
+    const prefix = capitalize ? unitName[0].toUpperCase() : unitName[0];
+    return prefix + "m";
+  }
+
+  return null;
```
toloudis, line 56: just curious, might it simplify things to return "" instead?

toloudis:
Awesome work!  I can't find where in the code our json format (used by CFE data) gets turned into microns.  I didn't see VolumeLoader doing any fixups on the incoming json.

src/VolumeLoader.ts
frasercl: you know what, it would

frasercl: Nevermind - that's there in case an unrecognized unit is specified, so that the loader can fall back to the full name at the call site. It could still return `unitName` instead though.

src/ThreeJsPanel.ts
```diff
@@ -308,6 +311,58 @@ export class ThreeJsPanel {
     this.axisCamera.position.set(-this.axisOffset[0], -this.axisOffset[1], this.axisScale * 2.0);
   }
 
+  orthoScreenPixelsToPhysicalUnits(pixels: number, physicalUnitsPerWorldUnit: number): number {
+    // At orthoScale = 0.5, the viewport is 1 world unit tall
+    const worldUnitsPerPixel = this.orthoScale * 2 / this.getHeight();
```
toloudis, line 316: Just for the sake of discussion, and fully understanding the coordinate transforms, another way to say this is that the ortho cameras are setting up a coordinate system that spans `-orthoScale` to `+orthoScale` from bottom to top (and `-orthoScale*aspect`, `+orthoScale*aspect`) from left to right.  so the total range from bottom to top = top-bottom = orthoScale - (-orthoScale) = 2.0 * orthoScale.

frasercl:
`loadJson` expects the data it's pointed to to have type `ImageInfo` right? In that case we'd have no way of knowing the unit symbol for these images unless either the field `unit_symbol` was added to the underlying data, or the client called `setScaleUnit`. (or there's another source of metadata I'm not aware of.)

toloudis:
My thought was that since the only known creator of that json is us (AICS), then all images in that format must be microns so we could just enforce it in the code - if unit not provided then use microns for volumeloader's load from json function.

src/ThreeJsPanel.ts
```diff
@@ -308,6 +314,64 @@ export class ThreeJsPanel {
...
+  updateOrthoScaleBar(scale: number, unit?: string): void {
+    // We want to find the largest round number of physical units that keeps the scale bar within this width on screen
+    const SCALE_BAR_MAX_WIDTH = 150;
+    // Convert max width to volume physical units
+    const physicalMaxWidth = this.orthoScreenPixelsToPhysicalUnits(SCALE_BAR_MAX_WIDTH, scale);
+    // Round off all but the most significant digit of worldSpaceMaxWidth
+    const digits = Math.floor(Math.log10(physicalMaxWidth));
+    const div10 = 10 ** digits;
+    const scaleValue = Math.floor(physicalMaxWidth / div10) * div10;
+    let scaleStr = scaleValue.toString();
```
toloudis, line 356: did you consider using `toFixed` instead of this log10 stuff?

src/ThreeJsPanel.ts
```diff
@@ -308,6 +314,64 @@ export class ThreeJsPanel {
...
+    Object.assign(this.orthoScaleBarElement.style, orthoScaleBarStyle);
+    this.containerdiv.appendChild(this.orthoScaleBarElement);
+  }
+
+  updateOrthoScaleBar(scale: number, unit?: string): void {
+    // We want to find the largest round number of physical units that keeps the scale bar within this width on screen
+    const SCALE_BAR_MAX_WIDTH = 150;
+    // Convert max width to volume physical units
+    const physicalMaxWidth = this.orthoScreenPixelsToPhysicalUnits(SCALE_BAR_MAX_WIDTH, scale);
+    // Round off all but the most significant digit of worldSpaceMaxWidth
```
toloudis, line 352: this comment has wrong var name in it

src/ThreeJsPanel.ts
toloudis: On second thought, `toFixed` might eliminate the need for the digits<1 check below but not the log10 math.

toloudis: Nevermind!

src/Volume.ts
```diff
@@ -15,6 +15,7 @@ export interface ImageInfo {
   pixel_size_x: number;
   pixel_size_y: number;
   pixel_size_z: number;
+  unit_symbol: string;
```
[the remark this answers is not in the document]
frasercl: Seems like it makes some sense given that this is the only way that physical dimensions are specified in `ImageInfo`. I might wonder if there are cases where the symbol is used in context with derived physical dimensions of a whole volume, where specifying pixel size in the name might be a bit unclear. But given that the unit is immediately copied into `Volume.physicalUnitSymbol` and used from there that shouldn't be an issue.

Issue #79, opened 2022-12-13
UX for scale bars


toloudis:
We want to add scale bars to the 3d and 2d modes of the volume-viewer.

See #66

We need help completing the spec details, including placement, styling, responsiveness and functionality.
Answer questions like:
Where to put the ruler on screen?  What happens if bounding box is on vs off?  How should it move if clipping drawer is open/closed?  Can we copy simularium for the 3d ruler? etc.

toloudis:
also see https://github.com/allen-cell-animated/volume-viewer/pull/82

toloudis:
also need recommendation for a Hide/Show Rulers button in the main UI (toolbar or advanced settings?)

frasercl:
Closed by feedback from @TravisKroeker and #101

PR #85, opened 2023-01-10
Feature/indicator position setters

frasercl:
Adds methods `setAxisPosition` and `setScaleBarPosition` to `View3d`, allowing the axis and scale indicators to be moved within the viewport. This gives a potential path forward for #66, #79, and allen-cell-animated/website-3d-cell-viewer#73, which all have to do with how these indicators interact with web-3d-viewer's clipping panel.

Both methods have the signature `(marginX: number, marginY: number, corner?: "top_left" | "top_right" | "bottom_left" | "bottom_right") => void`, where `marginX` and `marginY` are pixel offsets into the viewport from `corner`, and `corner` defaults to the current positions of these elements (`"bottom_left"` for the axes, `"bottom_right"` for the scale bar).

Also, removes unused parameters `oldScale` and `currentScale` from `View3d`, and cleans up some initialization code for the axis indicator.

Screenshot from web-3d-viewer, with both elements moved out of the way of the clipping drawer:
image

### DIFF
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -311,10 +311,10 @@ export class ThreeJsPanel {
     // at offsets lower than BASE_MARGIN, axes may extend off screen
     const BASE_MARGIN = 50;
     this.axisOffset = [marginX + BASE_MARGIN, marginY + BASE_MARGIN];
-    if (corner.startsWith("top")) {
+    if (corner % 4 < 2) { // top
       this.axisOffset[1] = this.getHeight() - this.axisOffset[1];
     }
-    if (corner.endsWith("right")) {
+    if (corner % 2) { // right
       this.axisOffset[0] = this.getWidth() - this.axisOffset[0];
     }
     this.axisCamera.position.set(-this.axisOffset[0], -this.axisOffset[1], this.axisScale * 2.0);
@@ -387,14 +387,8 @@ export class ThreeJsPanel {
     style.removeProperty("left");
     style.removeProperty("right");
 
-    let [yProp, xProp] = corner.split("_");
-    // Sanity check: must be valid css property names
-    if (xProp !== "left" && xProp !== "right") {
-      xProp = "right";
-    }
-    if (yProp !== "top" && yProp !== "bottom") {
-      yProp = "bottom";
-    }
+    const xProp = corner % 2 ? "right" : "left";
+    const yProp = corner % 4 < 2 ? "top" : "bottom";
 
     Object.assign(style, {
       [xProp]: marginX + "px",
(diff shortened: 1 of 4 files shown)


toloudis (approved):

frasercl:
Check-in before merging: would `ViewportCorner` make more sense as an enum? Pros: potentially more elegant as far as data is concerned and less open to errors; cons: would require users of this API to import the type, or else read the documentation more closely for which numeric variants corresponded to which corners.

toloudis:
> Check-in before merging: would `ViewportCorner` make more sense as an enum? Pros: potentially more elegant as far as data is concerned and less open to errors; cons: would require users of this API to import the type, or else read the documentation more closely for which numeric variants corresponded to which corners.

Good question... to me, not obvious whether you should or not.  Having said that:
In general, I favor stronger typing so I tend to lean toward using an enum... having type-checked string literals is nice too, rather than having to use the enum symbolic names.  I also have no problem with exporting another type out of this library if it's part of the public API.
Does css provide a built in type to validate things like top, left, bottom, right?

frasercl:
> Does css provide a built in type to validate things like top, left, bottom, right?

Yes, there are a couple types (React has its own I believe) that specify valid CSS properties, though with the way the element style interface works enforcement can be patchy in practice. E.g. `CSSStyleDeclaration.removeProperty` does not enforce that its argument be a valid property name, and will quietly do nothing if it doesn't get one.

I also would vaguely prefer to use an enum in the code rather than doing string processing, I just wasn't sure if it was worth the bit of extra interface complexity.

src/types.ts
```diff
@@ -80,3 +80,10 @@ export interface VolumeDisplayOptions {
 
 export const isOrthographicCamera = (def: Camera): def is OrthographicCamera =>
   def && (def as OrthographicCamera).isOrthographicCamera;
+
+export const enum ViewportCorner {
+  TOP_LEFT = 0,
```
toloudis, line 85: I'm pretty sure you can use string values! This is what I thought you meant to do all along:
TOP_LEFT = "top_left"

src/ThreeJsPanel.ts
```diff
@@ -372,6 +379,23 @@ export class ThreeJsPanel {
...
+  setOrthoScaleBarPosition(marginX: number, marginY: number, corner: ViewportCorner) {
+    const { style } = this.orthoScaleBarElement;
+
+    style.removeProperty("top");
+    style.removeProperty("bottom");
+    style.removeProperty("left");
+    style.removeProperty("right");
+
+    const xProp = corner % 2 ? "right" : "left";
+    const yProp = corner % 4 < 2 ? "top" : "bottom";
```
toloudis, line 391: This `%2` and `%4` stuff is a bit clunky and would probably be better as a function that lives closer to the enum itself (can TS enums have member functions?).. or else make it much less dependent on the count and order of the enum values

Issue #87, opened 2023-01-11
refactor VolumeLoader


toloudis:
## Use Case

In part driven by zarr, we want to prepare for more flexible loading of different volume data sources, including presenting user-selectable options when loading 

## Solution

Current thought:  Split apart VolumeLoader into a subdirectory... each different volume loader function gets put into its own file.

Refactor to an interface that supports a sequence like this:
* retrieve metadata such as dimensions and names
* allow for some optional user interaction such as a dialog box or other ui
* create Volume object from a given ImageInfo object
* start retrieving pixel data asynchronously and interruptably

For future planned work to support multiple data sources, we may need to reconsider some of this to allow "appending" loaded channels to an already existing Volume object.

PR #88, opened 2023-01-23
Feature/volume loaders refactor

toloudis:
Addresses #87 

Until now, all the different volume format loaders were in one big single module.  This PR splits them out into their own files.

I also add a dropdown selector to the test app to load a variety of test data via the UI.

The loader implementations are very much the same but this also prepares for a bit more flexibility:  the IVolumeLoader now has a loadDimensions function that can be used to predetermine memory usage etc before constructing a Volume or starting to download the pixel data.  This is in anticipation of multiresolution zarr data where there might be some user intervention before actually starting to load the files.

This is a BREAKING change with respect to website-3d-cell-viewer - all volume loading has to go through the IVolumeLoaders.


public/index.ts
```diff
@@ -1082,13 +1111,74 @@ function createTestVolume() {
   };
 }
 
+function createLoader(type: string): IVolumeLoader {
```
toloudis, line 1114: this function could go into IVolumeLoader and become the primary (or only?) way to create a IVolumeLoader.

public/index.ts
```diff
@@ -1082,13 +1111,74 @@ function createTestVolume() {
...
+  switch (type) {
+    case "opencell":
+      return new OpenCellLoader();
+    case "omezarr":
+      return new OMEZarrLoader();
+    case "ometiff":
+      return new TiffLoader();
+    // case "procedural":
+    //   return new RawVolumeLoader();
```
toloudis, line 1123: the above is a hint to imply that we could make a loader that just accepts a TypedArray with the volume data but currently the creation is so simple it seemed overkill.

kmitcham (approved):

public/index.ts
```diff
@@ -7,18 +7,84 @@ import {
...
   RENDERMODE_PATHTRACE,
   RENDERMODE_RAYMARCH,
   SKY_LIGHT,
 } from "../src";
-import { State } from "./types";
+import { IVolumeLoader, LoadSpec } from "../src/loaders/IVolumeLoader";
+import { JsonImageInfoLoader } from "../src/loaders/JsonImageInfoLoader";
+import { OMEZarrLoader } from "../src/loaders/OmeZarrLoader";
+import { OpenCellLoader } from "../src/loaders/OpenCellLoader";
+import { TiffLoader } from "../src/loaders/TiffLoader";
```
frasercl, line 20: For external users src/index.ts will cut down on all this importing, but I wonder if this might be worth a src/loaders/index.ts

public/index.ts
frasercl: I like this as an option for apps that may switch between many file formats, but I also like the idea of clients that don't expect to use more than one format only needing to import the loader they need.

toloudis: Yeah, probably most useful as a way to pass in a LoadSpec, and have some logic to automatically figure out what kind of loader to use so that the client doesn't need to care.  Factory method pattern.  But not sure how often that will come up.  Something for down the road...

src/index.ts
```diff
@@ -2,20 +2,26 @@ import { RENDERMODE_PATHTRACE, RENDERMODE_RAYMARCH, View3d } from "./View3d";
...
+import { OMEZarrLoader } from "./loaders/OmeZarrLoader";
+import { TiffLoader } from "./loaders/TiffLoader";
 
 import { Light, AREA_LIGHT, SKY_LIGHT } from "./Light";
 
 export type { ImageInfo } from "./Volume";
 export type { ControlPoint, Lut } from "./Histogram";
 export {
   Histogram,
+  IVolumeLoader,
```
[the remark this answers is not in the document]
toloudis: Oops that's what I get for throwing that in at the last minute thinking it was harmless.  I'll fix that.

src/loaders/OpenCellLoader.ts
```diff
@@ -0,0 +1,165 @@
...
+   * }, {
+   *     "name": "AICS-10_5_5.ome.tif_atlas_2.png",
+   *     "channels": [6, 7, 8]
+   * }], mycallback);
+   */
+  static loadVolumeAtlasData(
+      volume: Volume,
+      imageArray: PackedChannelsImage[],
+      onChannelLoaded: PerChannelCallback
+    ): PackedChannelsImageRequests {
```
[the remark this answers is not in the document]
toloudis: So the OpenCell data just happens to be formatted as tiled rgb images, totally coincidental with our json format. Maybe better if OpenCellLoader calls into Json loader.  Otherwise I'm also kind of not worried if they stay separate.  OpenCellLoader is really not likely to be used by any production code -- in fact, maybe this loader could be moved into the public dir to live with the test app.
I'll have a look and refactor that.

public/index.ts
toloudis: Agreed - I'll try to change this to use src/index.ts for all the imports, as if the test app were really using the packaged volume-viewer.

frasercl (approved):

PR #89, opened 2023-01-25
Feature/3d scale indicator

### LINKED ISSUE #66: show scale bar

frasercl:
Add tick marks to the bounding box as in Simularium, along with Simularium's 2d indicator of spacing between tick marks.

The spacing between tick marks is always some power of 10 physical units that produces at least 2 tick marks along the volume's longest axis. The scale indicator matches the color of the bounding box, and is only visible when all these conditions are met: the viewer is in 3d mode and not path tracing, the bounding box is enabled, and scale bars have not been disabled with `View3d.setShowScaleBar`.

Combined with #82, this resolves #66.

### Screenshots
image
(From Cell Feature Explorer)
image

### DIFF
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -73,7 +73,7 @@ export default class RayMarchedAtlasVolume {
     this.cubeTransformNode = new Group();
     this.cubeTransformNode.name = "VolumeContainerNode";
 
-    this.cubeTransformNode.add(this.boxHelper, this.cubeMesh, this.tickMarksMesh);
+    this.cubeTransformNode.add(this.boxHelper, this.tickMarksMesh, this.cubeMesh);
 
     this.uniforms = rayMarchingShaderUniforms;
 
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -29,9 +29,6 @@ import { ThreeJsPanel } from "./ThreeJsPanel";
 
 import { Bounds, FuseChannel } from "./types";
 
-// TODO calculate tick mark sizes better
-const NUM_TICK_MARKS = 4;
-
 const BOUNDING_BOX_DEFAULT_COLOR = new Color(0xffff00);
 
 export default class RayMarchedAtlasVolume {
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -42,6 +42,7 @@ export default class RayMarchedAtlasVolume {
   private boxHelper: Box3Helper;
   private tickMarksMesh: LineSegments;
   private scaleTextMesh: Mesh;
+  private rulerMesh: Mesh;
   private tickMarksPhysicalLength: number;
   private cubeTransformNode: Group;
   private uniforms: typeof rayMarchingShaderUniforms;
@@ -73,6 +74,7 @@ export default class RayMarchedAtlasVolume {
 
     this.tickMarksMesh = new LineSegments();
     this.scaleTextMesh = new Mesh();
+    this.rulerMesh = new Mesh();
     this.tickMarksMesh.updateMatrixWorld();
     this.tickMarksMesh.visible = false;
     this.tickMarksPhysicalLength = 1;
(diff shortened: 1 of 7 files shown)


toloudis:
This is looking awesome and will solve a real "quantitative" use case for our users.

First impressions from the screen shots:
- tick marks are too long, half size might be fine?
- 10 tick marks seems like too many. Smaller tickmarks may simply be enough to solve that and make it look less intrusive. 
Otherwise we might want to have a dynamic scale value or get one more UX intervention.   By dynamic scale value, find an integer number of units where the number of tick marks on the longest axis will be 8 (for example).

src/RayMarchedAtlasVolume.ts
```diff
@@ -56,16 +61,19 @@ export default class RayMarchedAtlasVolume {
...
+    this.tickMarksMesh = this.createTickMarks();
+    this.tickMarksMesh.updateMatrixWorld();
+    this.tickMarksMesh.visible = false;
+
     this.cubeTransformNode = new Group();
     this.cubeTransformNode.name = "VolumeContainerNode";
 
-    this.cubeTransformNode.add(this.boxHelper);
-    this.cubeTransformNode.add(this.cubeMesh);
+    this.cubeTransformNode.add(this.boxHelper, this.cubeMesh, this.tickMarksMesh);
```
toloudis, line 76: Nitpick, but logically I feel like this order should be boxhelper, then tickmarks, then cubemesh.. just having boxhelper next to tickmarks...

src/ThreeJsPanel.ts
```diff
@@ -369,18 +399,36 @@ export class ThreeJsPanel {
...
-  updateOrthoScaleBarVisibility(): void {
-    const visible = isOrthographicCamera(this.camera) && this.showOrthoScaleBar;
-    this.orthoScaleBarElement.style.display = visible ? "" : "none";
+  updatePerspectiveScaleBar(length: number, unit?: string): void {
+    const labeldiv = this.perspectiveScaleBarElement.firstChild as HTMLDivElement;
+    labeldiv.innerHTML = `${length}${unit || ""}`;
+  }
+
+  setPerspectiveScaleBarColor(color: [number, number, number]): void {
+    const svgdiv = this.perspectiveScaleBarElement.lastChild as HTMLDivElement;
```
toloudis, line 408: maybe a comment about how this function is very specific to the contents of scaleBarSVG.ts ?  Or is there a better way to name the element to be colored?

src/View3d.ts
```diff
@@ -223,6 +227,9 @@ export class View3d {
 
       if (unit) {
         this.image.volume.setUnitSymbol(unit);
+      } else if (!isOrthographicCamera(this.canvas3d.camera)) {
+        // `else if` because `setUnitSymbol` would have done this already
```
toloudis, line 231: I'm not sure I understand this comment

src/ThreeJsPanel.ts
frasercl: as in a note that this will only change the color of the line in the SVG, not of the bounding box or tick marks?

src/View3d.ts
frasercl: Hmm... actually, the `else if` statement it's referencing shouldn't be there in the first place. I thought the function called on 229 would call `updatePerspectiveScaleBar`, and therefore this function should only check to call it itself if no unit was provided (this is what the comment was supposed to mean). Not only does the function on 229 not call `updatePerspectiveScaleBar`, but `updatePerspectiveScaleBar` should be called regardless of whether the current camera is orthographic or not.

src/ThreeJsPanel.ts
toloudis: Yes, something like
// we are coloring only part of the svg.  if the structure of the svg ever changes this may break due to using `lastChild`

toloudis: or alternatively, extract the element in a more direct way by getting it by name.  I'm not sure if that is possible.  But a comment is maybe sufficient to warn future eyes looking at this code.

frasercl: This is definitely a somewhat brittle way of doing this, but it's not actually reaching into the SVG. It's setting the font color of its container, and only some paths in the SVG are set up to respond to that. I could leave a comment to explain this, something like:

``// set the font color of the SVG container. only paths with `stroke="currentColor"` will react to this.``

toloudis: Thanks!  Yes I think that comment would be perfect.

toloudis (approved):
The shorter tick marks look much less intrusive.  Thanks!

frasercl:
Will be worth circling back with UX team later about number of tick marks, among many other small questions, but I'm comfortable with the state this is in now.

PR #96, opened 2023-02-24
select optimal zarr multiresolution based on a 2kX2k tiled texture size

toloudis:
The zarr dimension code tries to auto-select a multiresolution level to display.  The autoselection of multiresolution level was picking the worst resolution even if a better one would fit.  
This fix computes an estimate of the best multiresolution layer based on how we would ideally tile the z slices in our internal texture for rendering.
The code in the zarr loader is a bit ugly because I want to remap the dimension order back to TCZYX even if it is ordered differently inside the file. So there are some indirections through the index remappings.  Some of this remapping stuff was being repeated in two places so I did a DRY refactor into functions.

The result of this code is that the zarr loader won't always just select the worst resolution.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -89,7 +89,6 @@ class OMEZarrLoader implements IVolumeLoader {
     // get all shapes
     for (const i in multiscales) {
       const shape = await fetchShapeOfLevel(store, imagegroup, multiscales[i]);
-      // just stick it in multiscales for now.
       if (shape.length != axes.length) {
         console.log("ERROR: shape length " + shape.length + " does not match axes length " + axes.length);
       }
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -1,9 +1,68 @@
 import { IVolumeLoader, LoadSpec, PerChannelCallback, VolumeDims } from "./IVolumeLoader";
-import { computePackedAtlasDims, spatialUnitNameToSymbol } from "./VolumeLoaderUtils";
+import { computePackedAtlasDims, estimateLevelForAtlas, spatialUnitNameToSymbol } from "./VolumeLoaderUtils";
 import { ImageInfo } from "../Volume";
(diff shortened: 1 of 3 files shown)


src/loaders/OmeZarrLoader.ts
```diff
@@ -156,7 +160,13 @@ class OMEZarrLoader implements IVolumeLoader {
...
     }
 
-    const downsampleZ = 2; // z/downsampleZ is number of z slices in reduced volume
+    const optimalLevel = estimateLevelForAtlas(spatialDims, 2048);
+    // assume all levels are decreasing in size.  If a larger level is optimal then use it:
+    if (optimalLevel < levelToLoad) {
+      levelToLoad = optimalLevel;
+    }
+
+    const downsampleZ = 1; // z/downsampleZ is number of z slices in reduced volume
```
toloudis, line 169: previous zarr loading code was only loading half the z slices... not what we want to do just yet.

kmitcham (approved):

src/loaders/OmeZarrLoader.ts
```diff
@@ -16,42 +73,13 @@ class OMEZarrLoader implements IVolumeLoader {
...
-    if (axisTCZYX[3] > -1) {
-      spatialAxes.push(axisTCZYX[3]);
-    }
-    if (axisTCZYX[4] > -1) {
-      spatialAxes.push(axisTCZYX[4]);
-    }
-    if (spatialAxes.length != 3) {
-      console.log("ERROR: zarr loader expects a z, y, and x axis.");
-    }
+    const spatialAxes: number[] = findSpatialAxesZYX(axisTCZYX);
```
frasercl, line 82: Any reason to type this and the function it calls `[number, number, number]`, or would that just create problems?

frasercl (approved):

src/loaders/OmeZarrLoader.ts
toloudis: wouldn't hurt.  I think for proper ome-zarr, Z spatial axis is not required, but for the time being this viewer expects it.

PR #97, opened 2023-03-02
Feature/option to append viewer to document manually

frasercl:
## Problem

`View3d`'s constructor must currently be supplied with a parent element to which to append the viewer element on creation. This leads to a pattern in website-3d-cell-viewer where the viewer is only created after its containing component has mounted and rendered a container, leading in turn to TypeScript requiring null checks any time `view3d` is used because it is not definitely assigned in the constructor. Three.js exposes a `domElement` property on renderers and allows users to add it where they want in the document; if `View3d` used a similar pattern, it would allow website-3d-cell-viewer to initialize the viewer on component creation and avoid null checks.

## Solution

- Remove the required `parentElement` property on `View3d`'s constructor and move it into an optional property on the options object.
- When resizing to fit the containing element, get the container by querying `canvas3d.containerdiv`'s parent, rather than storing the provided parent element.
- Add method `View3d.getDOMElement` to get the viewer element.

This is a BREAKING CHANGE to `View3d`'s constructor signature. Some trivial, but ugly, changes could be made to make the constructor backwards-compatible if necessary.

Also, replace a bunch of `image` null checks with TypeScript's more concise `?.` syntax.

### DIFF
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -135,10 +135,7 @@ export class View3d {
    * @param {VolumeChannelDisplayOptions} options
    */
   setVolumeChannelOptions(volume: Volume, channelIndex: number, options: VolumeChannelDisplayOptions): void {
-    if (!this.image) {
-      return;
-    }
-    this.image.setChannelOptions(channelIndex, options);
+    this.image?.setChannelOptions(channelIndex, options);
     this.redraw();
   }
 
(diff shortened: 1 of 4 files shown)


src/View3d.ts
```diff
@@ -135,10 +136,7 @@ export class View3d {
    * @param {VolumeChannelDisplayOptions} options
    */
   setVolumeChannelOptions(volume: Volume, channelIndex: number, options: VolumeChannelDisplayOptions): void {
-    if (!this.image) {
-      return;
-    }
-    this.image.setChannelOptions(channelIndex, options);
+    this.image?.setChannelOptions(channelIndex, options);
```
toloudis, line 139: these are incurring a functionality change: when this.image is null, redraw will be called now where it wasn't before.

toloudis (commented):
LGTM

PR #98, opened 2023-03-06
imageMetadata

toloudis:
Each Volume now gets a imageMetadata object.
The object is empty by default, and is intended to be populated by the IVolumeLoaders.
There is a simple utility function for now, which does the same thing for every volume.  But in the future now there is room for the VolumeLoaders to inject metadata that is available only in the file format they are loading (e.g. OME Metadata).

The intent here is that the website-3d-cell-viewer can now just pick up this `volume.imageMetadata` for its metadata display, and not need to consider what its contents are.

2 open questions: 
1. should I type this any differently than a generic Record<string, unknown> ?
2. do we need a code path to create an imageMetadata for volumes that do not get created by IVolumeLoaders?

### DIFF
--- a/src/loaders/VolumeLoaderUtils.ts
+++ b/src/loaders/VolumeLoaderUtils.ts
@@ -1,5 +1,7 @@
 import "regenerator-runtime/runtime";
 
+import { ImageInfo } from "../Volume";
+
 export type TypedArray =
   | Uint8Array
   | Int8Array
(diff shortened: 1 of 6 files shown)


src/loaders/VolumeLoaderUtils.ts
```diff
@@ -76,3 +78,25 @@ export function estimateLevelForAtlas(spatialDimsZYX: number[][], maxAtlasEdge =
   }
   return levelToLoad;
 }
+
+// currently everything needed can come from the imageInfo
+// but in the future each IVolumeLoader could have a completely separate implementation.
+export function buildDefaultMetadata(imageInfo: ImageInfo): Record<string, unknown> {
+  const metadata = {};
+  metadata["Dimensions"] = { x: imageInfo.tile_width, y: imageInfo.tile_height, z: imageInfo.tiles };
```
frasercl, line 95: This could also be expressed as a large object literal, e.g.
```
metadata = {
  "Original dimensions": { x: imageInfo.width, y: imageInfo.height, z: imageInfo.tiles },
  ...
}
```
...though I don't have strong feelings about which is preferable.

src/loaders/VolumeLoaderUtils.ts
```diff
@@ -76,3 +78,37 @@ export function estimateLevelForAtlas(spatialDimsZYX: number[][], maxAtlasEdge =
...
   return levelToLoad;
 }
+
+function isEmpty(obj) {
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
+      return false;
+    }
+  }
+  return true;
```
frasercl, line 88: I'm unfamiliar with this approach to testing for object emptiness. My instinct would be that `Object.keys(obj).length === 0` would do it, but it looks like this way might be more robust for objects with prototypes? Though `userData` is unlikely to have an inheritance chain.

frasercl (approved):
I'm liking seeing metadata extraction at this level a lot better, especially the idea of having a default metadata builder that can be extended/overridden by specific loaders.

src/loaders/VolumeLoaderUtils.ts
toloudis: Same here, don't really care.  Imperative code is more ingrained in me than declarative, especially when there might be some complexity to arrive at the values being set.

src/loaders/VolumeLoaderUtils.ts
toloudis: Yeah when I went to make userData optional, I just quickly googled it and found that `{}` is truthy, so ultimately I picked this one from lodash's implementation.

src/loaders/VolumeLoaderUtils.ts
frasercl: My instinct would have been to go declarative, but given that "User data" may or may not be added to the object at all, I think it makes sense the way it is

src/loaders/VolumeLoaderUtils.ts
frasercl: JavaScript's strange approach to truth strikes again

PR #100, opened 2023-03-15
Fix/horizontal scale in orthographic modes breaks on resize

frasercl:
Provides a fix for allen-cell-animated/website-3d-cell-viewer#122.

The problem had to do with resize handling code erroneously using the zoom level of the orthographic camera when updating aspect ratio. Switching to a 2d mode, zooming in or out, then triggering a resize (which in website-3d-cell-viewer also occurs on actions like toggling bounding box or interpolation) would thus set the camera to an incorrect aspect.

The zoom level appeared to have made it into that calculation due to confusion over whether the value (`ThreeJsPanel.orthoScale`) referred to the initial/"default" scale or the current scale. To clarify, I removed this property and replaced it with a constant (`ORTHO_SCALE`) where the former was expected and a getter (`getOrthoScale`) where the latter was expected.

### DIFF
--- a/src/PathTracedVolume.ts
+++ b/src/PathTracedVolume.ts
@@ -31,7 +31,7 @@ import {
 import { LUT_ARRAY_LENGTH } from "./Histogram";
 import Volume from "./Volume";
 import { Bounds, isOrthographicCamera } from "./types";
-import { ThreeJsPanel } from "./ThreeJsPanel";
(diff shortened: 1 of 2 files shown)


toloudis (approved):
LGTM.  It is common/typical in realtime/game programming to have a pattern like this:
while (true) {
  updateThingsBasedOnUserInput();
  redrawScene();
}
where the update call would check our controls object and make sure that the new orthoscale is registered wherever it needs to be.

PR #101, opened 2023-03-16
Fix/tickmarks only appear in 3d mode

frasercl:
Feedback from @TravisKroeker re: #79: tickmarks only appear on the bounding box in 3D view mode.


toloudis (approved):
this makes me wonder if we can set the visibility of tickMarksMesh at render time instead of having to track it in both places.  it feels like it could get more complicated with more options.  Or even something like "updateTickMarksVisibility()" to have all the logic in one single place.  Anyway, this works!

Issue #93, opened 2023-02-01
allow the ability to load a different time from time series


toloudis:
## Use Case

Microscopists want to view time series zarr data using web viewer in the near-term.

## Solution

Internals implementation separate from ux:  
* make sure we can select time when loading zarr files by url  (website-3d-cell-viewer only?)
* we need a prop so that embedding apps can fully specify the volume data to load (Can't quite remember if we made a LoadSpec prop that can have fields like `time`).  
* could experiment with a temporary time slider in dat.gui or tweakpane (see #92)
* default loading time 0 if unspecified

If the volume loader detects time series, we can present a time slider.  On changing time slider, request volume data with all settings the same as currently loaded except for a different time index.  Time slider should be presented via experimental ui according to #92

toloudis:
It would be good to have consistent usage patterns across Agave and the web viewer for dealing with time series (and overall).

toloudis:
This can consider zarr-only for now.  We need not make other attempts to do time series loading yet.  
There is another possible use case in which the time series may be in a sequence of separate files as opposed to one single zarr url.  We would have to consider how the sequence of separate files would be specified.

frasercl:
After some discussion, first steps for this API should be:
- The `createVolume` method on `IVolumeLoader` should be split into two methods:
  - `createVolume`, which creates a new `Volume` object with the correct properties as specified by a `LoadSpec` but *does not fetch any data*;
  - and something like `loadVolumeData`, which actually fetches data and places it in a preexisting `Volume`
- `Volume`s should keep hold of their own `LoadSpec`
- The `View3d` method for loading different time points in a time series should look like `setTime(volume: Volume, time: number, loader: IVolumeLoader)`, which just modifies `volume`'s `LoadSpec` and calls `loader.loadVolumeData(volume);`.
  - Though come to think of it, could this function also pick and create the appropriate loader? Perhaps `loader` can start as an optional parameter for in case loaders ever start maintaining their own state.

frasercl:
Above API was implemented in #104, but I didn't mark this issue as resolved because that API seemed to be tentative/transitional. But remaining work towards a final polished interface may be better tracked by a task under allen-cell-animated/website-3d-cell-viewer#155?

toloudis:
If the demo viewer time slider in volume-viewer can use the new api (directly or indirectly) then we can close out this ticket.  The initial description is purposely excluding the website-3d-cell-viewer layer.   Any refinements can be in new tickets.

PR #104, opened 2023-05-15
Feature/time series API first steps

frasercl:
Implements the minimal first API changes described in the thread for #93:

- splits `IVolumeLoader.createVolume` into two methods: `createVolume` to initialize a `Volume` object, and `loadVolumeData` to load data into it
- `Volume` objects now own a `LoadSpec`, which `createVolume` sets and `loadVolumeData` uses
- Added `View3d.setTime`, which takes a `Volume`, an `IVolumeLoader`, and a `PerChannelCallback`. Note that, like `loadVolumeData`, `setTime` relies on the caller to trigger the propagation of channel data into the viewer in the per-channel callback.

`OMEZarrLoader` was the thorniest challenge when disentangling `Volume` creation with volume data loading, so the most intensive changes are in there. I also found it helpful to encode the OME-Zarr metadata spec in types to understand what was going on, so that file is now rigorously typed. Finally, while I was there I found a couple instances of an `await` statement within a for-loop which could be lifted out of the loop to allow promises to resolve concurrently, so I did that (and confirmed with `console.time` that it was worth doing).

Note also that I allowed `OMEZarrLoader` and `TiffLoader` to cache certain trivial and expensive but necessary values in `createVolume` to be used and deleted in the next call to `loadVolumeData`. This may introduce too much of a risk of unexpected behavior to keep around, or perhaps `createVolume` ought to just call `loadVolumeData` itself.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -311,7 +311,7 @@ class OMEZarrLoader implements IVolumeLoader {
     /* eslint-enable @typescript-eslint/naming-convention */
 
     // got some data, now let's construct the volume.
-    const vol = new Volume(imgdata);
+    const vol = new Volume(imgdata, loadSpec);
     vol.imageMetadata = buildDefaultMetadata(imgdata);
     return vol;
   }
@@ -334,7 +334,7 @@ class OMEZarrLoader implements IVolumeLoader {
       const spatialAxes = findSpatialAxesZYX(axisTCZYX);
 
       const levelToLoad = await pickLevelToLoad(multiscale, store, loadSpec, spatialAxes);
-      this.multiscalePath = multiscale.datasets[levelToLoad];
+      this.multiscalePath = multiscale.datasets[levelToLoad].path;
     }
 
     const storepath = loadSpec.subpath + "/" + this.multiscalePath;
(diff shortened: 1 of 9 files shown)


src/loaders/OmeZarrLoader.ts
```diff
@@ -132,71 +236,41 @@ class OMEZarrLoader implements IVolumeLoader {
...
-    const multiscales = allmetadata.multiscales[imageIndex].datasets;
-    const axes = allmetadata.multiscales[imageIndex].axes;
+    const multiscale: OMEMultiscale = allmetadata.multiscales[imageIndex];
+    const { datasets, axes } = multiscale;
 
     const axisTCZYX = remapAxesToTCZYX(axes);
-    const hasT = axisTCZYX[0] > -1;
-    const hasC = axisTCZYX[1] > -1;
+    this.hasT = axisTCZYX[0] > -1;
+    this.hasC = axisTCZYX[1] > -1;
```
toloudis, line 244: There probably should be a comment somewhere at the top of the IVolumeLoader class to say that loaders are stateful and should be kept alive.  If you call createVolume with one loader instance and then have to use a different one later, will hasT and hasC be initialized?  Maybe consider a function to lazily initialize hasT and has C just in case... not a strict requirement for this PR...

toloudis (approved):
Looks good!

PR #105, opened 2023-06-08
Feature/read time unit

frasercl:
### Problem

website-3d-cell-viewer time series designs show a time step indicator next to the scale bar. To display this we need to know the scale and units of the volume's time dimension.

### Solution

- Add fields to `ImageInfo`: `time_scale`, `time_unit`.
- Get `OMEZarrLoader` to populate them where available.
- Generalize `spatialUnitNameToSymbol` to handle converting time units to symbols too (as `unitNameToSymbol`).

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -1098,6 +1098,8 @@ function createTestVolume() {
     pixel_size_unit: "",
     transform: { translation: [0, 0, 0], rotation: [0, 0, 0] },
     times: 1,
+    time_scale: 1,
+    time_unit: "",
   };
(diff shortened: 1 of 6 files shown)


toloudis (approved):
Looks reasonable.  Consider moving all the unit handling utilities into a units.ts.

frasercl:
I think including unit-to-symbol conversion with other methods more obviously associated with data loading is still just on the edge of reasonable given the size of the file. But if either category grows by much more, that would probably tip the scale towards separating them out.

PR #108, opened 2023-06-13
fix ome-zarr dims as they are passed in to Volume class

toloudis:
The volume object sets up dimensions with an expectation that we may display downsampled data and certain values correspond to the full resolution data.  Correct the zarr loader to supply the values that Volume expects.

Key code is in Volume.ts in `setVoxelSize`.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -267,7 +267,9 @@ class OMEZarrLoader implements IVolumeLoader {
     const channels = this.hasC ? multiscaleShape[axisTCZYX[1]] : 1;
     const sizeT = this.hasT ? multiscaleShape[axisTCZYX[0]] : 1;
 
-    const scale5d = getScale(dataset);
+    // we want scale of level 0
+    const scale5d = getScale(datasets[0]);
+
     const timeScale = this.hasT ? scale5d[axisTCZYX[0]] : 1;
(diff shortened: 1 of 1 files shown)


src/loaders/OmeZarrLoader.ts
```diff
@@ -285,10 +287,14 @@ class OMEZarrLoader implements IVolumeLoader {
     for (let i = 0; i < displayMetadata.channels.length; ++i) {
       chnames.push(displayMetadata.channels[i].label);
     }
+
+    // get shape of level 0
+    const shape0 = await fetchShapeOfLevel(store, loadSpec.subpath, datasets[0]);
```
frasercl, line 292: Curious what the overhead of this function is (`async` because it calls zarr `openArray`). How much work does `openArray` do? Does it have a cache to work with?

frasercl (approved):

src/loaders/OmeZarrLoader.ts
toloudis: Yep, this is another small network round trip to get the shape info.  With ome-zarr you basically have to do this once per level if you want to get all the array dims for everything.  Pretty minor though, and negligible compared to the pixel data requests that come after.

toloudis: I doubt that the zarr.js lib does any internal caching of its own.  Would be amazing if it did.

ShrimpCryptid (approved):

PR #109, opened 2023-06-21
Feature/secret tweakpane

### LINKED ISSUE #92: add a secret tweakpane ui
## Use Case

We want to be able to deliver experimental features to power users in a way that does not interfere with releases.

## Solution

Simularium already implemented this - a ui that pops up over the viewport and uses tweakpane (a more maintained alternative to dat.gui).  The tweakpane ui is tied to a special keypress to hide/show it.  See simularium/simularium-viewer for reference.

This will allow us to test out things like a time slider for internal users before it has a full UX treatment.

frasercl:
Resolves #92: adds a secret tweakpane UI, accessible with the same key combo from Simularium (Ctrl+Alt+1), with controls for a subset of viewer and volume settings.

Adding settings, particularly channel settings, gave me trouble for a couple reasons:
- Tweakpane wants to bind to an object and be pointed to a key to modify. A number of settings are stored in ways that break this pattern: they are accessible only behind getters, or must be derived from other state values, etc.
  - For channel settings specifically, `VolumeDrawable` keeps an array of objects of type `VolumeChannelDisplayOptions`, but these are only filled in when new values are supplied by the user through a couple specific methods (`setOptions`/`setChannelOptions`). Other methods leave them unchanged and modify state in other places.
- From what I can tell, it appears to currently be impossible to bind Tweakpane to a color stored as an array (`[number, number, number]`). Though it's not impossible to design an involved workaround to this constraint, I eventually decided it was out of scope.

Owing to the above, this code contains some ugly workarounds - in particular, the use of a `Proxy` to pretend that the state of a single channel is stored in a single coherent place. I'd love to know if there's a better way to achieve this without moderate rearchitecting of the app, but I suspect there might not be. If nothing else, the lessons of this struggle will hopefully be helpful going into allen-cell-animated/website-3d-cell-viewer#155 as we think about how to redesign some of our data structures to better utilize the cloud.


src/View3d.ts
```diff
@@ -853,4 +874,52 @@ export class View3d {
...
+      display: "none",
+    };
+    Object.assign(pane.element.style, paneStyle);
+
+    // LIGHTS
+    const lights = pane.addFolder({ title: "Lights (isosurface)" });
+
+    const addFolderForLight = (light: ThreeLight, title: string): void => {
+      const folder = lights.addFolder({ title, expanded: false });
+      folder.addInput(light, "color", { color: { type: "float" } });
```
toloudis, line 909: My first thought for things that don't map 1-1 with an internal value, is to create an intermediate data structure like 
`tweakable:{ambientLight:{color:whateverMapsWellToTweakPane,...}, ...}`
And then in the tweakpane change handler do the transformation to set the value into the deep internal object.

src/VolumeDrawable.ts
```diff
@@ -685,4 +696,75 @@ export default class VolumeDrawable {
...
+      .on("change", ({ value }) => this.setRayStepSizes(undefined, value));
+
+    const channels = pane.addFolder({ title: "Channels", expanded: false });
+    this.channelOptions.forEach((options, channelIndex) => {
+      // eslint-disable-next-line @typescript-eslint/no-this-alias
+      const self = this;
+      type ChannelState = WithObjectColors<VolumeChannelDisplayOptions>;
+      // Volume channel options are not stored in a consistent place.
+      // Perhaps they should be. For now, this proxy lets us pretend they are.
+      const optionsProxy = new Proxy(options as ChannelState, {
```
toloudis, line 717: along the lines of my other comment, I would have just made a simple data structure to hold the "tweakable" values per channel and then map the change handlers of the tweakpane to drill in and set the stuff.  But this is very instructive to me about how to use Proxy!

src/VolumeDrawable.ts
```diff
@@ -10,14 +10,25 @@ import { Bounds, FuseChannel } from "./types";
...
 
 // A renderable multichannel volume image with 8-bits per channel intensity values.
 export default class VolumeDrawable {
   public PT: boolean;
   public volume: Volume;
   private onChannelDataReadyCallback?: () => void;
-  private translation: Vector3;
-  private rotation: Euler;
+  public translation: Vector3;
+  public rotation: Euler;
```
toloudis, line 31: do these need to be public because of `pane.addInput(this, "translation")` ?

toloudis:
Sorry this was more of an odyssey than intended.  All we needed was the tweakpane itself (and maybe one tweakable just to prove it can be hooked up) -  you went above and beyond by figuring out how to map all that stuff into it. 

I like how clean and minimally intrusive you kept it.   I could see evolving this to be a separate .ts file with the setupGui, and the tweakables (and/or the Proxy object you made).  Long term, I don't know if we really need the tweakpane full of all the per-channel settings that are already in the UI in other places.

src/View3d.ts
frasercl: This function is mostly an anti-redundant-code measure, as lights actually work pretty reasonably in this respect. It could be unwound into four separate blocks of bindings for these lights, or removed from the body of `setupGui`.

src/VolumeDrawable.ts
frasercl: Yeah, I don't necessarily love what I came up with here, but I wanted to avoid creating new values which could potentially become out of sync with the settings used for rendering. (Though these would still have to be deliberately propagated to tweakpane with `refresh`.)

Despite the effort it took to work out, I wouldn't mind if it didn't make it into the merged version of this PR. The main value of it for me was learning the providence of all the settings in it, and it will remain in the commit graph should it ever be useful to reference later.

src/VolumeDrawable.ts
frasercl: Right, `addInput` has a type constraint that the second argument must be a `keyof` the first, and `"translation"` doesn't appear in the interface when it's private (even within its own method!).

frasercl:
> Sorry this was more of an odyssey than intended.

That's mostly my bad for not checking in about scope creep and being stubborn about finding solutions. Again, I'm okay scaling back what's here to the least messy parts before merging, since the most "above and beyond" elements here are also the least legible.

src/View3d.ts
```diff
@@ -853,4 +874,52 @@ export class View3d {
...
+      this.tweakpaneOpen = !this.tweakpaneOpen;
+      this.tweakpane.element.style.display = this.tweakpaneOpen ? "block" : "none";
+    }
+  };
+
+  removeEventListeners(): void {
+    window.removeEventListener("keydown", this.handleKeydown);
+  }
+
+  setupGui(container: HTMLElement): Pane {
```
toloudis, line 890: maybe make this private just to be clear that no one else should be calling it

toloudis:
So in terms of trimming this back a bit, my take is to remove the per-channel settings stuff, since it should be already present in other UI whether in the testbed or the react front end, and then merge this.  I still like keeping a `image.setupGui` call.  The next time there is a "tweakable" to add (I'm thinking cache mem limit, number of slices to prefetch, etc), we can further remove stuff that is extra.

src/VolumeDrawable.ts
```diff
@@ -685,4 +696,15 @@ export default class VolumeDrawable {
...
+
+  setupGui(pane: Pane): void {
+    pane.addInput(this, "translation").on("change", ({ value }) => this.setTranslation(value));
+    pane.addInput(this, "rotation").on("change", ({ value }) => this.setRotation(value));
+
+    const pathtrace = pane.addFolder({ title: "Pathtrace", expanded: false });
+    pathtrace.addInput(this, "primaryRayStepSize").on("change", ({ value }) => this.setRayStepSizes(value));
+    pathtrace
+      .addInput(this, "secondaryRayStepSize")
+      .on("change", ({ value }) => this.setRayStepSizes(undefined, value));
```
frasercl, line 708: Do these options have reasonable min/max values?

toloudis: reasonable min is 1.0.  zero is a crashy value, and anything less than 1 would be odd.   Max is something like 20.  I'll double check from agave

toloudis: Agave uses setRange(1.0, 100.0);

toloudis: Basically the values are supposed to be roughly equivalent to voxels and you don't really want rays stepping over that many voxels, so a reasonable max is something like half the expected number of voxels. give or take.

toloudis (approved):
I really don't like that tweakpane forces you to make private things public, but it certainly shouldn't hold up this PR.

PR #116, opened 2023-06-27
Feature/RequestQueue for bulk async operations

### LINKED ISSUE #111: Write a RequestQueue
## Use Case

- Handles prefetching requests for 2D slice data

## Solution
- API should include:
  - Canceling all current requests
  - Request new frame data (or array/range of frames) - pass in LoadSpec and TRange?
    - Or should requests be formed outside of RequestQueue? (and just expose a unique, comparable index value?)
    - Check if requests overlap with existing requests, drop other requests in flight
- Unit testing (mocking, test server, let requests complete at different time intervals)
- Observer pattern -> let others observe completion/rejection of request

ShrimpCryptid:
## Problem
Resolves #111, adding a new RequestQueue class.

## Details
RequestQueue is a generic utility class that matches asynchronous operations (requests) to unique keys. Requests can be any asynchronous callback. While the request is in the queue and pending, additional calls to add the request will not cause the operation to be called again.

Multiple requests can be added and configured with an optional delay. A maximum number of concurrent operations can also be set to save resources and bandwidth!

### Additional Notes:
- Keys are stored and compared using the provided `keyStringifyFn`, which by default uses `JSON.stringify`.
- RequestQueue makes no guarantees about Promise return types (`Promise<unknown>`), as it is possible for multiple requests with the same key but different return types to be issued. (This should not happen in our normal use case.)
  - Users must do type checking themselves, examples of which can be found in the tests.
- I tried to be very detailed with docstrings on this class. Please let me know if any of it could be improved!


src/utils/RequestQueue.ts
```diff
@@ -0,0 +1,271 @@
...
+   * actions must be responsible for checking the RequestQueue, determining if the request is still valid (e.g.
+   * using `.hasRequest()`), and stopping or returning early.
+   *
+   * @returns A promise that will resolve on completion of the request, or reject if the request is cancelled.
+   *  If multiple requests are issued with the same key, a promise for the first request will be returned
+   *  until the request is resolved or cancelled.
+   *  Note that the return type of the promise will match that of the first request's instance.
+   */
+  public addRequest<T>(key: K, requestAction: () => Promise<T>): Promise<unknown> {
+    const keyString = this.keyStringifyFn(key);
```
toloudis, line 106: what if you let each key provide its own stringification without having to declare it in the RequestQueue class constructor?  For example, this code could do (pseudocode):
```
if key is string then keystring = key
else if key has a toString() function, then keystring=key.toString()
... any other way of trying to stringify it...
```
This type of system would let you use all sorts of different object types as keys as long as they provide their own hash function that our class recognizes.  
 I just worry slightly that JSON.stringify is too heavy-handed.
 This is not a must-have but something to consider!

src/utils/RequestQueue.ts
```diff
@@ -0,0 +1,271 @@
...
+   * NOTE: Cancelling a request while the action is running WILL NOT stop the action. If this behavior is desired,
+   * actions must be responsible for checking the RequestQueue, determining if the request is still valid (e.g.
+   * using `.hasRequest()`), and stopping or returning early.
+   *
+   * @returns A promise that will resolve on completion of the request, or reject if the request is cancelled.
+   *  If multiple requests are issued with the same key, a promise for the first request will be returned
+   *  until the request is resolved or cancelled.
+   *  Note that the return type of the promise will match that of the first request's instance.
+   */
+  public addRequest<T>(key: K, requestAction: () => Promise<T>): Promise<unknown> {
```
toloudis, line 105: could this take an optional delay parameter as its third arg?

src/test/RequestQueue.test.ts
```diff
@@ -0,0 +1,357 @@
...
+      const promises: Promise<unknown>[] = [];
+      for (let i = 0; i < iterations; i++) {
+        promises.push(
+          rq.addRequest(i, async () => {
+            await sleep(5);
+            throw new Error("Test error (should be caught)");
+          })
+        );
+      }
+      await Promise.allSettled(promises).catch((_) => {
```
toloudis, line 129: maybe slightly more interesting/nontrivial to have half fail and half succeed too.

toloudis: It would also be good to make sure that an error message passed out via reject can be received (do string comparison, as if the failures were for different reasons).

toloudis (approved):
Pretty awesome set of tests.  Thank you!!  This might be a good replacement for the task queue in simularium now!  I like that you reused the concurrency limit, which allows for requests to invoke workers.

src/test/RequestQueue.test.ts
ShrimpCryptid: Added! This was a useful activity for seeing how promises handle errors too :)

src/utils/RequestQueue.ts
ShrimpCryptid: Added, refactored some other code to be cleaner too 👍

src/utils/RequestQueue.ts
ShrimpCryptid: We talked about how RequestQueue never stores object references internally, only strings, so we made the decision to switch it to accepting string keys only. Thank you for the suggestion!

src/utils/RequestQueue.ts
```diff
@@ -0,0 +1,295 @@
...
+   * Adds multiple requests to the queue, with an optional delay between each.
+   * @param requests An array of RequestItems, which include a key and a request action.
+   * @param delayMs An optional minimum delay in milliseconds to be added between each request.
+   *  For example, a delay of 10 ms will cause the second request to be added to the processing queue
+   *  after 10 ms, the third to added after 20 ms, and so on. Set to 10 ms by default.
+   * @returns An array of promises corresponding to the provided requests. (i.e., the `i`th value
+   * of the returned array will be a Promise for the resolution of `requests[i]`). If a request
+   *  with a matching key is already pending, returns the promise for the initial request.
+   */
+  public addRequests<T>(requests: Request<T>[], delayMs = 10): Promise<unknown>[] {
```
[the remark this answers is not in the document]
ShrimpCryptid: Oh, yes! Good point now that addRequest also accepts delays.

src/utils/RequestQueue.ts
```diff
@@ -0,0 +1,295 @@
...
+  /** Action to be run. */
+  action: () => Promise<V>;
+  /** Reference to the promise object that will be resolved when the action is complete. */
+  promise: Promise<V>;
+  /** Callback used to resolve the promise. */
+  resolve: (value?: V | PromiseLike<V> | undefined) => void;
+  /** Callback used to reject the promise. */
+  reject: (reason?: unknown) => void;
+  /** Optional, used to track timeouts if the item will be added to the queue later. */
+  timeoutId?: NodeJS.Timeout;
```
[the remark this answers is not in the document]
toloudis: good catch.

src/utils/RequestQueue.ts
```diff
@@ -0,0 +1,295 @@
...
+          this.allRequests.delete(key);
+          requestItem.reject(err);
+          this.dequeue();
+        });
+    } catch (err) {
+      this.activeRequests.delete(key);
+      this.allRequests.delete(key);
+      requestItem.reject(err);
+      this.dequeue();
+    }
```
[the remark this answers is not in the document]
ShrimpCryptid: Done! Added some extra tests too to make sure this works the way I expected it to.

src/utils/RequestQueue.ts
```diff
@@ -0,0 +1,256 @@
...
+    // Mark that this request is active
+    this.activeRequests.add(key);
+    
+    await requestItem.action().then(
+      (fulfilledValue) => {
+        requestItem.resolve(fulfilledValue);
+      }, (rejectedValue) => {
+        requestItem.reject(rejectedValue);
+      }
+    );
```
frasercl, line 200: No need to wrap `resolve`/`reject` in arrow functions here, they can be passed directly: `.then(requestItem.resolve, requestItem.reject)`

frasercl (approved):
Looks good! One minor style comment remaining but this can merge with or without it.

src/utils/RequestQueue.ts
ShrimpCryptid: Thanks for the tip!

PR #117, opened 2023-06-27
Introduce a VolumeRenderImpl interface layer

toloudis:
Gather together the common functions for different Volume rendering implementations to be used inside of VolumeDrawable, preparing to add a new one for Agave remote rendering.


src/RayMarchedAtlasVolume.ts
```diff
@@ -366,4 +367,8 @@ export default class RayMarchedAtlasVolume {
     this.setUniform("textureAtlas", this.channelData.getFusedTexture());
     this.setUniform("textureAtlasMask", this.channelData.maskTexture);
   }
+
+  public setRenderUpdateListener(_listener?: ((iteration: number) => void) | undefined) {
+    0;
```
ShrimpCryptid, line 372: Any reason for this statement vs. a return?

ShrimpCryptid (approved):
Looks good to me!

src/RayMarchedAtlasVolume.ts
toloudis: No special reason. As I recall, typescript didn't want it to be just empty, and putting a 0 seemed like the least amount of typing to have the function do nothing. :).  I know things like lodash provide a `noop` but I figured this did the job without bringing a dependency.  Now that you mention this, `return;` seems a bit more obvious.

public/index.ts
```diff
@@ -1339,11 +1339,12 @@ function main() {
     view3D.setMaxProjectMode(myState.volume, mp);
   };
   renderModeSelect?.addEventListener("change", ({ currentTarget }) => {
-    if ((currentTarget as HTMLOptionElement)!.value === "PT") {
+    const target = (currentTarget as HTMLOptionElement)!;
```
frasercl, line 1342: Curious how smart typescript and/or babel and/or V8 are about deleting a simple alias like this

src/VolumeRenderImpl.ts
frasercl, line 1: This file could use a run through prettier to fix a bit of inconsistent spacing

frasercl (approved):
LGTM!

src/VolumeRenderImpl.ts
toloudis: thanks - not sure why it wasn't autoformatted before.

PR #119, opened 2023-06-30
Refactor/Add GLSL file support

ShrimpCryptid:
This refactor adds support for GLSL shader files! `.glsl`, `.frag`, `.vert`, `.fs`, and `.vs` filetypes can be imported with the following syntax and will be treated as strings:

```
import rayMarchVertexShader from "./shaders/raymarch.vert";
import rayMarchFragmentShader from "./shaders/raymarch.frag";
```

Babel will also automatically import them as inline strings when building. This refactor should make it easier to write and debug shaders with existing linting tools.

This update also moves the shader definitions for path tracing and ray marching into their own separate files.


toloudis (approved):
Cool. 
I still kinda wonder if there was a typescript-only way of doing it, as in what if we didn't use babel but still wanted the transformation into exported string modules...

frasercl (approved):
Into this, would love to get easy GLSL syntax highlighting in this repo. I think there may also be extensions/loaders/plugins that will do a bit of minification for you, if we want to look for them in the future (though that moves in the opposite direction to what Dan is describing, towards greater build complexity)

PR #122, opened 2023-07-05
Feature/xy rendering

ShrimpCryptid:
Hello! This PR tackles issue https://github.com/allen-cell-animated/volume-viewer/issues/113, rendering Z-slices with a custom shader.

- Dan and I discussed how RayMarchedAtlasVolume and Atlas2DSlice should be related, and we agreed that in the future there should probably be a refactor (some common base abstract class) on top of a larger change to how VolumeDrawable should pass settings and parameters to the render implementations.

### Changelog:
- Adds a new VolumeRenderImpl (`Atlas2DSlice`) and a supporting fragment and vertex shader.
- Modifies RayMarchedAtlasVolume to allow the new Atlas2DSlice to extend it and override its geometry and shader material.
- Adds a new method to VolumeRenderImpl for setting the Z-slice (`setZSlice`) on volume renderers that implement it.
- Adds UI for setting the Z-slice from the test visualizer.

### Screenshots
image


src/View3d.ts
```diff
@@ -464,16 +464,27 @@ export class View3d {
...
    * @param {string} mode Mode can be "3D", or "XY" or "Z", or "YZ" or "X", or "XZ" or "Y".  3D is a perspective view, and all the others are orthographic projections
    */
   setCameraMode(mode: string): void {
     this.canvas3d.switchViewMode(mode);
+    this.image?.setViewMode(mode);
     this.image?.setIsOrtho(mode !== "3D");
     this.canvas3d.redraw();
   }
 
+  setZSlice(slice: number): boolean {
```
toloudis, line 479: just for consistency, please add `volume: Volume` as the first arg.

src/Atlas2DSlice.ts
```diff
@@ -0,0 +1,53 @@
...
+      transparent: true,
+      depthTest: true,
+      depthWrite: false,
+    });
+
+    mesh.material = threeMaterial;
+    return [geom, mesh];
+  }
+
+  public setZSlice(_slice: number): boolean {
```
toloudis, line 43: code style:  this is not strictly enforced, but usually if a function arg is prefixed with underscore it is indicating that it's unused.   Typically to appease typescript/eslint.

src/VolumeDrawable.ts
```diff
@@ -385,8 +414,8 @@ export default class VolumeDrawable {
         this.pathTracedVolume.updateActiveChannels(this);
       }
     } else {
-      if (this.rayMarchedAtlasVolume) {
-        this.rayMarchedAtlasVolume.fuse(this.fusion, this.volume.channels);
+      if (this.volumeRendering instanceof RayMarchedAtlasVolume) {
+        this.volumeRendering.fuse(this.fusion, this.volume.channels);
```
toloudis, line 418: I believe in the upcoming refactor, `fuse` can just do a single call on this.volumeRendering and let each Impl take care of the details :)

src/VolumeRenderImpl.ts
```diff
@@ -24,6 +24,12 @@ export interface VolumeRenderImpl {
   setShowBoundingBox: (_show: boolean) => void;
   setBoundingBoxColor: (_color: [number, number, number]) => void;
   viewpointMoved: () => void;
+  /**
+   * Sets the visible z slice, for volume renderers that support it.
+   * @returns true if the z slice was set correctly, false if the slice is out of range.
+   * Defaults to true if z-slice rendering is not implemented on the volume renderer.
+   */
+  setZSlice: (_slice: number) => boolean;
```
toloudis, line 32: There is a trap you can fall into here which is that whenever there is a special case suddenly every other subclass has to deal with it because it had to be added to the polymorphic interface, and the interface keeps growing.    I don't have a better solution for now; just noting this.  The refactor has maybe some potential to reduce the number of interface methods.

toloudis (approved):
Looks good. My only concern is the interaction with pathtrace mode, whether functionality is same as before.  Looking forward to the refactor to split the two things apart, even though it will seem like redundant code :)

src/Atlas2DSlice.ts
ShrimpCryptid: Yup, `_slice` is marked with an underscore here because it's unused in some implementations of `VolumeRenderImpl`.

src/VolumeDrawable.ts
```diff
@@ -297,6 +306,26 @@ export default class VolumeDrawable {
...
+  // TODO: Change mode to an enum
+  /**
+   * Sets the camera mode of the VolumeDrawable.
+   * @param mode Mode can be "3D", or "XY" or "Z", or "YZ" or "X", or "XZ" or "Y".
+   */
+  setViewMode(mode: string): void {
+    this.viewMode = mode;
+    if (mode === "XY" || mode === "Z") {
+      // If currently in 3D raymarch mode, hotswap the 2D slice
+      if (this.rayMarchedAtlasVolume && !this.atlas2DSlice) {
```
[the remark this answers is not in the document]
ShrimpCryptid: Currently it will just show an orthographic z-view in pathtrace mode. Switching to raymarched rendering from pathtraced rendering (or vice-versa) will still show the slice correctly.

Should we show the 2D slice in pathtrace mode as well?
image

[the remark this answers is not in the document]
toloudis: Thanks for checking. That behavior seems fine to me.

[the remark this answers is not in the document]
toloudis: It might make sense to create a ticket for always using the sliceImpl for XY mode no matter what.  But no need to change it in this PR just yet.

src/Atlas2DSlice.ts
```diff
@@ -0,0 +1,53 @@
...
+    // Clamp the slice value
+    _slice = Math.floor(_slice);
+    if (_slice < 0 || _slice > this.volume.z - 1) {
+      return false;
+    }
+    this.setUniform("Z_SLICE", _slice);
+    this.geometryMesh.material.needsUpdate = true;
```
frasercl, line 50: This is already set in `setUniform`. (In fact, I've learned since putting that line there that `needsUpdate` doesn't actually need to be set when uniforms change at all.)

src/View3d.ts
```diff
@@ -464,16 +464,27 @@ export class View3d {
     this.scene.add(this.lightContainer);
   }
 
+  // TODO: Change mode to an enum
```
frasercl, line 467: Agreed!

src/View3d.ts
```diff
@@ -464,16 +464,27 @@ export class View3d {
...
     this.canvas3d.redraw();
   }
 
+  setZSlice(volume: Volume, slice: number): boolean {
+    const ret = this.image?.setZSlice(slice);
+    if (ret) {
+      this.canvas3d.redraw();
+      return true;
+    }
+    return false;
```
frasercl, line 485: nit: could return `ret` here

src/RayMarchedAtlasVolume.ts
```diff
@@ -58,9 +59,8 @@ export default class RayMarchedAtlasVolume implements VolumeRenderImpl {
     this.scale = new Vector3(1.0, 1.0, 1.0);
     this.isOrtho = false;
 
-    this.cube = new BoxGeometry(1.0, 1.0, 1.0);
-    this.cubeMesh = new Mesh(this.cube);
-    this.cubeMesh.name = "Volume";
+    this.uniforms = rayMarchingShaderUniforms;
```
frasercl, line 62: Do we copy `rayMarchingShaderUniforms` anywhere or will all instances of this class that may exist share the same reference? (Potentially niche issue and also one that predates this PR if it exists at all)

src/Atlas2DSlice.ts
frasercl: In this case I would name the argument with an underscore only in implementations where the field is unused. Argument names don't have to match between interfaces and implementations as long as their types do.

frasercl (approved):
Handful of minor comments but this looks solid!

src/VolumeRenderImpl.ts
ShrimpCryptid: This gets addressed in the VolumeRenderImpl refactor, yeah!

src/View3d.ts
ShrimpCryptid: Fixed, thank you

src/RayMarchedAtlasVolume.ts
toloudis: Yeah I think this has been a potential issue all along - might want to check if ShaderMaterial makes its own copy of the uniforms objects passed in, or else we should clone it when passing it in.

toloudis: Because we only ever have one of these active at any time, it hasn't ever been a concern.

ShrimpCryptid: Fixed based on our discussion, thank you!

src/constants/volumePTshader.ts
```diff
@@ -12,67 +12,69 @@ const SHADERTYPE_BRDF = 0;
...
+        mN: new Vector3(),
+        mScreen: new Vector4(), // left, right, bottom, top
+        mInvScreen: new Vector2(), // 1/w, 1/h
+        mFocalDistance: 0.0,
+        mApertureSize: 0.0,
+        mIsOrtho: 0.0,
+      },
+    },
+    gLights: {
+      value: [new Light(SKY_LIGHT), new Light(AREA_LIGHT)],
```
ShrimpCryptid, line 55: Formatting/indent change

src/constants/volumeRayMarchShader.ts
```diff
@@ -5,113 +5,116 @@ import rayMarchFragmentShader from "./shaders/raymarch.frag";
...
+    },
+    volumeScale: {
+      type: "v3",
+      value: new Vector3(1.0, 1.0, 1.0),
+    },
+    textureRes: {
+      type: "v2",
+      value: new Vector2(1.0, 1.0),
+    },
+  };
```
ShrimpCryptid, line 119: Formatting/indentation

PR #124, opened 2023-07-10
refactor: VolumeRenderSettings

### LINKED ISSUE #123: refactor VolumeRenderImpl classes
## Use Case

All VolumeRenderImpls should be considered independent and separate ways of drawing a volume.  Interface inheritance = good, implementation inheritance = less good.  Favor looser coupling between unrelated components.

## Solution

A followup change to https://github.com/allen-cell-animated/volume-viewer/pull/122

Atlas2DSlice should not inherit anything from RayMarchedAtlasVolume.  We can consider them completely separate ways for a VolumeDrawable to do its rendering.  This way if we need to change one or the other in the future, we won't have to worry that we are affecting both. 

It was also discussed that there may be some encapsulation of the settings that VolumeDrawable holds on to, that it can pass down to its VolumeRenderImpls for them to access.  Having such a third object (e.g. VolumeRenderSettings?) would also break a potential circular dependency where the code just passes the who VolumeDrawable itself down into its owned VolumeRenderImpls.
Ultimately, the VolumeDrawable may boil down to a VolumeRenderImpl, VolumeRenderSettings, and some logic about how to switch between different VolumeRenderImpls.   This is slightly speculative and not a strict specification to adhere to.

ShrimpCryptid:
This refactor simplifies `VolumeRenderImpl` and `VolumeDrawable` by moving the majority of the shared renderer settings into a new class, **`VolumeRenderSettings`.**

Fixes #123.

`VolumeRenderSettings` can be passed in the constructor to existing implementations of `VolumeRenderImpl`, meaning that swapping between renderers is much simpler and requires less boilerplate code to set and reset values. `VolumeDrawable` has had a lot of fields removed as a result!

- Settings are updated via an `updateRenderSettings` method on `VolumeRenderImpl`.
  - Most field setters in `RayMarchedAtlasVolume` and `PathTracedVolume` have been moved into this method.
  - Implementing volume renderers can ignore settings that are irrelevant.
- Added a utility class for working with the settings objects (cloning, deep equality, etc.)
- Added tests for the settings object and utility class.


src/VolumeRenderSettings.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** parameters: density, specular, emissive, glossiness */
+  MATERIAL = 32,
+  /** parameters: resolution, useInterpolation, pixelSamplingRate, primaryRayStepSize, secondaryRayStepSize*/
+  RESOLUTION_AND_SAMPLING = 64,
+  /** parameters: isOrtho, orthoScale, viewAxis, visible, maxProjectMode */
+  VIEW = 128,
+  ALL = Number.MAX_SAFE_INTEGER,
+}
+
+export enum Axis {
```
ShrimpCryptid, line 29: Finally added an Axis enum! View3d.ts 468, VolumeDrawable.ts 306 still take strings, however, because I wasn't sure which changes would affect the external API.

src/Atlas2DSlice.ts
ShrimpCryptid, line 1: These changes copy logic from RayMarchedAtlasVolume. I've removed the tick markers but everything else is more or less the same.

src/VolumeRenderSettings.ts
ShrimpCryptid: Also let me know if this belongs somewhere else.

src/Atlas2DSlice.ts
```diff
@@ -1,23 +1,148 @@
-import { BufferGeometry, Material, Mesh, PlaneGeometry, ShaderMaterial, ShapeGeometry } from "three";
-import { Volume } from ".";
-import { sliceFragmentShaderSrc, sliceVertexShaderSrc } from "./constants/volumeSliceShader";
-import RayMarchedAtlasVolume from "./RayMarchedAtlasVolume";
-import { rayMarchingShaderUniforms } from "./constants/volumeRayMarchShader";
+import { Box3, Box3Helper, BufferGeometry, Color, Group, LineBasicMaterial, Material, Matrix4, Mesh, PlaneGeometry, ShaderMaterial, ShapeGeometry, Vector2, Vector3 } from "three";
```
toloudis, line 1: I'm thinking autoformatting should have broken up this long line?

src/Atlas2DSlice.ts
```diff
@@ -1,23 +1,148 @@
...
+      VolumeRenderSettingUtils.resizeWithVolume(settings, volume);
+    }
+    this.updateSettings(settings, SettingsFlags.ALL);
+  }
+
+  public updateSettings(newSettings: VolumeRenderSettings, _dirtyFlags?: number | SettingsFlags) {
+    if (_dirtyFlags === undefined) {
+      _dirtyFlags = SettingsFlags.ALL;
+    } else if (_dirtyFlags === SettingsFlags.NONE) {
+      return;
```
toloudis, line 61: when would this happen?

src/Atlas2DSlice.ts
```diff
@@ -1,23 +1,148 @@
...
+    this.channelData = new FusedChannelData(volume.imageInfo.atlas_width, volume.imageInfo.atlas_height);
+
+    if (!settings) {
+      settings = defaultVolumeRenderSettings();
+      VolumeRenderSettingUtils.resizeWithVolume(settings, volume);
+    }
+    this.updateSettings(settings, SettingsFlags.ALL);
+  }
+
+  public updateSettings(newSettings: VolumeRenderSettings, _dirtyFlags?: number | SettingsFlags) {
```
toloudis, line 57: naming convention is to not have underscore for dirtyFlags

src/PathTracedVolume.ts
```diff
@@ -356,16 +337,80 @@ export default class PathTracedVolume implements VolumeRenderImpl {
...
+    if (_dirtyFlags === undefined) {
+      _dirtyFlags = SettingsFlags.ALL;
+    }
+    if (_dirtyFlags === SettingsFlags.NONE) {
+      return;
+    }
+    this.settings = newSettings;
+
+    // Update resolution
+    if (_dirtyFlags & SettingsFlags.RESOLUTION_AND_SAMPLING) {
```
toloudis, line 355: very minor/optional: resolution is also sampling so I suppose this could just be named SAMPLING!

src/PathTracedVolume.ts
```diff
@@ -636,13 +547,14 @@ export default class PathTracedVolume implements VolumeRenderImpl {
...
-  updateActiveChannels(image: VolumeDrawable): void {
+  updateActiveChannels(channelColors: FuseChannel[], channelData: Channel[]): void {
     const ch = [-1, -1, -1, -1];
     let activeChannel = 0;
     const NC = this.volume.num_channels;
     const maxch = 4;
     for (let i = 0; i < NC && activeChannel < maxch; ++i) {
-      if (image.isVolumeChannelEnabled(i) && image.getChannel(i).loaded) {
+      // check that channel is not hidden (rgbColor !== 0) and is loaded
+      if (channelColors[i].rgbColor !== 0 && channelData[i].loaded) {
```
toloudis, line 557: Also optional, minor comment:  I feel like `FuseChannel` should just have a member function `enabled():bool { return this.rgbColor !== 0; }` to hide this detail.  Perhaps using a named symbolic constant instead of literal 0 would help.  `const FUSE_DISABLED = 0;` or something.

src/Atlas2DSlice.ts
ShrimpCryptid: I DID NOT KNOW THIS WAS A THING
I AM VERY EXCITED ABOUT THIS FEATURE

src/Atlas2DSlice.ts
ShrimpCryptid: The only use case I could see if maybe setting values without restarting the rendering process in PathTracedVolume... but anytime the settings are being changed it should be changing the rendering process and restarting anyways.

Removed this and `SettingsFlags.NONE`!

src/View3d.ts
```diff
@@ -734,13 +735,13 @@ export class View3d {
    * Set clipping range (between 0 and 1) for a given axis.
    * Calling this allows the rendering to compensate for changes in thickness in orthographic views that affect how bright the volume is.
    * @param {Object} volume
-   * @param {number} axis 0, 1, or 2 for x, y, or z axis
+   * @param {number} axis x, y, or z axis
```
toloudis, line 738: looks like the type should be string here?

src/VolumeDrawable.ts
```diff
@@ -62,44 +52,19 @@ export default class VolumeDrawable {
...
-    this.gammaLevel = 1.0;
-    this.gammaMax = 1.0;
-
-    // TODO verify that these are reasonable
-    this.density = 0;
-    this.brightness = 0;
-
-    this.showBoundingBox = false;
-    this.boundingBoxColor = [1.0, 1.0, 0.0];
+    this.viewMode = Axis.NONE;
```
toloudis, line 67: none is the same as 3D?

src/View3d.ts
toloudis: This comment is only because the type in the actual function args is different, so maybe that is the one to change? something is out of sync here.

src/VolumeDrawable.ts
```diff
@@ -324,44 +318,64 @@ export default class VolumeDrawable {
...
+    this.volumeRendering.updateSettings(this.settings, SettingsFlags.CAMERA);
   }
 
   setFlipAxes(flipX: number, flipY: number, flipZ: number): void {
-    this.flipX = flipX;
-    this.flipY = flipY;
-    this.flipZ = flipZ;
-    this.volumeRendering.setFlipAxes(flipX, flipY, flipZ);
-    this.meshVolume.setFlipAxes(flipX, flipY, flipZ);
+    const flipAxes = new Vector3(flipX, flipY, flipZ);
```
toloudis, line 365: I wonder if the type hints for the args here could be `-1 | 1` instead of `number`

src/VolumeDrawable.ts
```diff
@@ -662,62 +667,28 @@ export default class VolumeDrawable {
 
     // create new
     if (isPathtrace) {
-      this.pathTracedVolume = new PathTracedVolume(this.volume);
+      this.pathTracedVolume = new PathTracedVolume(this.volume, this.settings);
```
toloudis, line 670: not sure if we still need to keep this.pathTracedVolume and this.rayMarchedAtlasVolume as member variables anymore?

src/VolumeDrawable.ts
```diff
@@ -662,62 +667,28 @@ export default class VolumeDrawable {
...
-      this.bounds.bmin.y + 0.5,
-      this.bounds.bmax.y + 0.5,
-      this.bounds.bmin.z + 0.5,
-      this.bounds.bmax.z + 0.5
-    );
-    this.setRayStepSizes(this.primaryRayStepSize, this.secondaryRayStepSize);
-
-    this.setShowBoundingBox(this.showBoundingBox);
-    this.setBoundingBoxColor(this.boundingBoxColor);
-
```
toloudis, line 720: should we be calling updateSettings here or is the one passed in to the constructors the way it happens?

src/VolumeDrawable.ts
```diff
@@ -726,35 +697,40 @@ export default class VolumeDrawable {
...
     pathtrace
-      .addInput(this, "secondaryRayStepSize", { min: 1, max: 100 })
+      .addInput(this.settings, "secondaryRayStepSize", { min: 1, max: 100 })
       .on("change", ({ value }) => this.setRayStepSizes(undefined, value));
   }
 
   setZSlice(slice: number): boolean {
-    if (this.volumeRendering.setZSlice(slice)) {
-      this.zSlice = slice;
+    if (slice < this.volume.z && slice > 0) {
```
toloudis, line 725: do you want to check to see if `this.settings.zSlice !== slice` here too?

src/VolumeRenderImpl.ts
```diff
@@ -1,38 +1,24 @@
...
-  setResolution: (_x: number, _y: number) => void;
-  setAxisClip: (_axis: "x" | "y" | "z", _minval: number, _maxval: number, _isOrthoAxis: boolean) => void;
-  setIsOrtho: (_isOrtho: boolean) => void;
-  setOrthoThickness: (_thickness: number) => void;
-  setInterpolationEnabled: (_enabled: boolean) => void;
-  setGamma: (_gmin: number, _glevel: number, _gmax: number) => void;
-  setFlipAxes: (_flipX: number, _flipY: number, _flipZ: number) => void;
   doRender: (_canvas: ThreeJsPanel) => void;
   cleanup: () => void;
-  onChannelData: (_batch: number[]) => void;
```
toloudis, line 18: just noticing this.  where did `onChannelData` end up?

src/VolumeRenderSettings.ts
toloudis: Not sure where it should live but it's easy enough to move later.  If we ever promote it to be part of the public API then it will certainly not want to live here.

src/VolumeRenderSettings.ts
```diff
@@ -0,0 +1,141 @@
...
+  primaryRayStepSize: number;
+  secondaryRayStepSize: number;
+};
+
+/**
+ * Returns a VolumeRenderSettings object with default fields. Default objects
+ * created with this method will not have shared references.
+ *
+ * Note that default objects have volume-dependent properties that should be updated
+ * with `VolumeRenderSettingUtils.updateWithVolume()`.
```
toloudis, line 90: why not just pass a Volume in to this function?  do you ever call it by itself?

src/VolumeRenderSettings.ts
frasercl: It may be safe to move to the public API now with string values that match what's used in those fields currently ("xy"/"xz"/"yz"/"3d"). (I believe it's technically possible to pass "x"/"y"/"z"? But I don't think it's done.)

src/VolumeRenderImpl.ts
ShrimpCryptid: None of the VolumeRenderImpl's implemented this method, so I'd removed it as part of the interface cleanup. Do you think it's worth keeping?

toloudis: if no one was using it then it can certainly go away :)

src/VolumeDrawable.ts
ShrimpCryptid: Mm, yeah, I can make that change. I can also remove the PT boolean flag and we can just check for what the current VolumeRenderImpl is an instance of.

src/VolumeRenderSettings.ts
ShrimpCryptid: This was a good suggestion, thank you! Added :)

ShrimpCryptid:
Added back the utility methods because of a discussion Dan and I had about future improvements to settings

src/VolumeRenderSettings.ts
ShrimpCryptid: `setCameraMode` in `View3d.ts` currently takes in a string and doesn't enforce typing ("XY"/"Z"/etc.) on it, so I think for now I'll skip making that change.

public/types.ts
```diff
@@ -67,9 +67,9 @@ export interface State {
 
   backgroundColor: [number, number, number];
 
-  flipX: number;
-  flipY: number;
-  flipZ: number;
+  flipX: -1 | 1;
+  flipY: -1 | 1;
+  flipZ: -1 | 1;
```
toloudis, line 72: this is increasingly out of scope but now these suddenly look like booleans to me 🤣

src/Atlas2DSlice.ts
```diff
@@ -1,23 +1,167 @@
...
+  ShapeGeometry,
+  Vector2,
+  Vector3,
+} from "three";
+import { Channel, Volume } from ".";
+import { sliceFragmentShaderSrc, sliceShaderUniforms, sliceVertexShaderSrc } from "./constants/volumeSliceShader";
+import { VolumeRenderImpl } from "./VolumeRenderImpl";
+import {
+  SettingsFlags,
+  VolumeRenderSettingUtils,
```
toloudis, line 22: lint says can remove...

public/types.ts
ShrimpCryptid: You're killing me here Dan 😭

ShrimpCryptid: I can make this change though, I just need to check that it doesn't touch external interfaces

toloudis: sorry, it's not a good candidate to happen in this PR, if at all.   no real reason to change it

src/VolumeRenderSettings.ts
```diff
@@ -137,4 +137,105 @@ export class VolumeRenderSettingUtils {
...
+   */
+  public static clone(src: VolumeRenderSettings): VolumeRenderSettings {
+    const dst = {};
+    for (const key of Object.keys(src)) {
+      const val = src[key];
+      if (val instanceof Array) {
+        dst[key] = this.deepCopyArray(val);
+      } else if (key === "bounds") {
+        // Bounds
+        dst["bounds"] = { bmax: new Vector3(), bmin: new Vector3() } as Bounds;
```
toloudis, line 227: i'd use `dst.bounds` here and make sure it's typed to VolumeRenderSettings

toloudis: also, can you use instanceof Bounds here instead of a string comparison ?

src/VolumeDrawable.ts
toloudis: I don't mind using "NONE" but maybe a comment here to say that NONE means full 3d volumetric view.  Maybe even an additional enum value that aliases to NONE.  `_3D` is ok, or maybe `VOLUMETRIC` or `XYZ`.

toloudis: At least a comment would be good, the rest is optional IMO.

src/VolumeDrawable.ts
```diff
@@ -343,7 +339,10 @@ export default class VolumeDrawable {
   }
 
   setOrthoThickness(value: number): void {
-    !this.PT && this.meshVolume.setOrthoThickness(value);
+    if (this.volumeRendering instanceof PathTracedVolume) {
```
toloudis, line 342: can we prefer checking the render mode as opposed to using instanceof whenever possible.  the use of instanceof is breaking the polymorphism so should be avoided in general. when you have to use it with a polymorphic type it means something's wrong with the interface...

src/VolumeDrawable.ts
```diff
@@ -613,23 +614,31 @@ export default class VolumeDrawable {
   }
 
   onStartControls(): void {
-    this.PT && this.pathTracedVolume && this.pathTracedVolume.onStartControls();
+    if (this.volumeRendering instanceof PathTracedVolume) {
```
toloudis, line 617: this should be `if(rendermode is pathtrace) then this.volumeRendering.onStartControls()` OR simply `this.volumeRendering.onStartControls()` and same for onChangeControls and onEndControls

src/VolumeRenderSettings.ts
ShrimpCryptid: Ah, I can't do an `instanceof Bounds` comparison because it's a type alias and not an actual class :((
> `'Bounds' only refers to a type, but is being used as a value here.ts(2693)`

ShrimpCryptid: I could check to see if the value is an object and if .bmax and .bmin exist on it? But that seems like it could run into the same problem if the names of `bmin/bmax/bounds` every changes

src/VolumeDrawable.ts
ShrimpCryptid: That's a good way to think about it, thank you. Fixing this!

src/VolumeDrawable.ts
```diff
@@ -458,7 +462,7 @@ export default class VolumeDrawable {
   }
 
   onChannelLoaded(batch: number[]): void {
-    this.volumeRendering.onChannelData(batch);
+    // this.volumeRendering.onChannelData(batch);
```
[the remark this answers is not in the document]
ShrimpCryptid: Deleted, thank you for the catch!

src/VolumeRenderSettings.ts
toloudis: yeah, if Bounds were a class it would work.  it's either change Bounds to be a class or leave it as is for now.  not a show stopper.

ShrimpCryptid: Going to leave it for now-- I tried changing Bounds to a class but there seem to be some consequences that cause other things to break (I suspect there are instances of Bounds not being declared correctly). Fortunately one of the tests for the VolumeRenderSettings will catch if `bmin/bmax/bounds` breaks or if the properties change, so I think this is okay.

src/VolumeDrawable.ts
```diff
@@ -298,70 +255,126 @@ export default class VolumeDrawable {
...
+        break;
+      case "XZ":
+      case "Y":
+        axis = Axis.Y;
+        break;
+      case "XY":
+      case "Z":
+        axis = Axis.Z;
+        break;
+    }
```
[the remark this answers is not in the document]
ShrimpCryptid: Fixed!

src/VolumeRenderSettings.ts
```diff
@@ -0,0 +1,243 @@
...
+  ROI = 8,
+  /** parameters: maskChannelIndex, maskAlpha */
+  MASK = 16,
+  /** parameters: density, specular, emissive, glossiness */
+  MATERIAL = 32,
+  /** parameters: resolution, useInterpolation, pixelSamplingRate, primaryRayStepSize, secondaryRayStepSize*/
+  SAMPLING = 64,
+  /** parameters: isOrtho, orthoScale, viewAxis, visible, maxProjectMode */
+  VIEW = 128,
+  ALL = Number.MAX_SAFE_INTEGER,
```
toloudis, line 25: the value of all should be 0xFFFFFFFF (the union of all single bits turned on).  I'm not sure what MAX_SAFE_INTEGER is.  Sometimes it can be clearer to write the flag constants as 0x1, 0x2, 0x4, 0x8, 0xF, 0x10, 0x20, 0x40, etc.

src/VolumeRenderSettings.ts
```diff
@@ -0,0 +1,243 @@
...
+      } else if (val instanceof String) {
+        dst[key] = "" + val;
+      } else {
+        // boolean, number, other primitives
+        dst[key] = val;
+      }
+    }
+    return dst as VolumeRenderSettings;
+  }
+}
```
toloudis, line 243: with all the code in this file, VolumeRenderSettings really looks more like a class so why not just make it one. Then all these functions can be member funcs in the class.  Is there any benefit to just keeping it as a type?

toloudis:
I see a few TODOs coming out of this PR which probably should be in separate tickets/pull requests.
* should we decide to change the flip axis parameter to be boolean or enum type at the public api level, and convert it to -1/1 when a shader needs to use it?  -- implies website-3d-cell-viewer may have to change
* same question for the 3d axis enum,  we should decide on one enum for the public API and then carry it through so it's consistent everywhere.
* we have to go over all this code and now use the FUSE_RGB_COLOR_DISABLE constant, and/or write a little fuseChannelIsEnabled(FuseChannel):boolean function that checks that value.

src/VolumeRenderSettings.ts
ShrimpCryptid: Hmm, the only thing I can think of is making declaration of the `VolumeRenderSettings` objects a little more readable (object declaration with properties labeled), but considering almost all initialization is done through the current `getDefaultVolumeRenderSettings` method anyways it's the same as having the constructor set up default properties. Any thoughts?

src/VolumeRenderSettings.ts
toloudis: even more explicit is `0b` notation where you can put the actual bit in position:
`var num8 = 0b001000; // binary`

src/VolumeRenderSettings.ts
toloudis: agree that constructor would be the default function, and clone would become a member function.   it's not a big problem -- we can leave things alone.  it's just one of those things where what you made is all the pieces of a class but it's not a class

src/VolumeRenderSettings.ts
toloudis: To be clear, the only important thing here is to know that ALL will actually have all bits turned on :)

src/VolumeRenderSettings.ts
ShrimpCryptid: Okay, will do!

src/VolumeRenderSettings.ts
```diff
@@ -38,107 +38,96 @@ export enum Axis {
...
       bmax: new Vector3(0.5, 0.5, 0.5),
-    },
-    boundingBoxColor: [1.0, 1.0, 0.0],
-    primaryRayStepSize: 1.0,
-    secondaryRayStepSize: 1.0,
-    useInterpolation: true,
-    visible: true,
-    maxProjectMode: false,
+    }),
+      (this.boundingBoxColor = [1.0, 1.0, 0.0]);
```
toloudis, line 111: weird extra parens here

src/VolumeRenderSettings.ts
```diff
@@ -214,21 +203,21 @@ export class VolumeRenderSettingUtils {
...
    * @param src The object to create a clone of.
    * @returns a new VolumeRenderSettings object with identical fields that do not
    * share references with the original settings object.
    */
-  public static clone(src: VolumeRenderSettings): VolumeRenderSettings {
-    const dst = defaultVolumeRenderSettings(new Volume()); // initialize with empty volume
-    for (const key of Object.keys(src)) {
-      const val = src[key];
+  public clone(): VolumeRenderSettings {
+    const dst = new VolumeRenderSettings(new Volume()); // initialize with empty volume
```
toloudis, line 212: maybe the arg to the ctor could be optional so you don't have to construct a new Volume here?

toloudis (approved):

frasercl (approved):

Issue #121, opened 2023-07-03
Do not convert XY slices to 8bit at fetch time


toloudis:
## Use Case

The viewer currently is programmed to convert zarr data to 8-bit intensity when it is fetched.  This 8-bit format is used throughout the viewer for performance reasons.  In XY mode for single slices, we can keep the slices in their native format for higher dynamic range and more faithful reproduction when viewing.

## Solution

Not sure how easy or hard this is yet:
Let slices stay in their native data type.  Load into textures and let the shader do the necessary LUTs and only go to 8-bit when drawing final pixels in fragment shader...

toloudis:
Also see https://github.com/allen-cell-animated/volume-viewer/pull/197

Issue #110, opened 2023-06-22
Zarr Loader can load single slices


ShrimpCryptid:
## Use Case

- Zarr loader needs to handle requesting 2D slices for time series data

## Solution

- Have a new API call in IVolumeLoader (`FetchSlice`) that only Zarr Loader implements
-- Other loaders will need to throw an error/noop/etc.
- Make use of Web Worker in the same way we currently do

## Optional
- Ideally can be unit tested, without relying on cloud resources? (Zarr store library -> create a fake Zarr array and try to load from it)
- Optimize for chunks containing multiple slices(?) (may need future issue/PR)

PR #127, opened 2023-07-12
Feature/load single slice

frasercl:
Resolve (?) #110: change relevant loading items (`OMEZarrLoader`, `IVolumeLoader`, `LoadSpec`) to allow loading of subregions from zarr arrays, including single slices.

### Changes:

- `LoadSpec` subregion fields (`minx`, `maxx`, `miny`, etc.) may be undefined to indicate loading the max extent of that axis 
- `OMEZarrLoader` now pays attention to these fields
- `FetchZarrWorker` now uses the `zarr.getRaw` method, which removes the need to de-nest typed arrays in the worker
- `IVolumeLoader` allows passing in an optional explicit `LoadSpec` on data load which overrides the one in the target `Volume`

### Does NOT (yet) change:

- Other discussed changes to `LoadSpec`: replace `subpath` with `scaleLevel`/`multiscale`, add array of channel indexes
- Replace the per-channel callback API with one based on an array of promises
- Make `Volume` aware that it contains a subset of its full extent. Currently, `OMEZarrLoader` will dutifully load a smaller-than-expected array of channel data into the full dimensions of the volume, and the volume will show up but with a lot of blank space. It's unclear how `Volume` should handle this.
- Allow loading original data types, rather than converting to uint8 (#121), though the simplifications to `FetchZarrWorker` will make this easier

### DIFF
--- a/src/test/RequestQueue.test.ts
+++ b/src/test/RequestQueue.test.ts
@@ -48,8 +48,8 @@ describe("test RequestQueue", () => {
       const rq = new RequestQueue();
       const startTime = Date.now();
       const delayMs = 15;
-      const immediatePromise = rq.addRequest("a", async () => {});
-      const delayedPromise = rq.addRequest("b", async () => {}, delayMs);
+      const immediatePromise = rq.addRequest("a", () => Promise.resolve());
+      const delayedPromise = rq.addRequest("b", () => Promise.resolve(), delayMs);
 
       const promises: Promise<unknown>[] = [];
       promises.push(
--- a/src/test/RequestQueue.test.ts
+++ b/src/test/RequestQueue.test.ts
@@ -48,18 +48,22 @@ describe("test RequestQueue", () => {
       const rq = new RequestQueue();
       const startTime = Date.now();
       const delayMs = 15;
-      const immediatePromise = rq.addRequest("a", async () => {return;});
-      const delayedPromise = rq.addRequest("b", async () => {return;}, delayMs);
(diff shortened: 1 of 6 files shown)


src/test/RequestQueue.test.ts
```diff
@@ -329,7 +333,7 @@ describe("test RequestQueue", () => {
       expect(count).to.equal(0);
     });
 
-    async function mockLoader(loadSpec: LoadSpec, maxDelayMs = 10.0): Promise<TypedArray> {
+    async function mockLoader(loadSpec: Required<LoadSpec>, maxDelayMs = 10.0): Promise<TypedArray> {
```
toloudis, line 336: TL;DG: what does Required mean ?

src/workers/FetchZarrWorker.ts
```diff
@@ -1,83 +1,61 @@
...
+  channelData.forEach((val: number, idx: number) => {
+    u8[idx] = ((val - min) / range) * 255;
+  });
+
   return u8;
 }
 
-self.onmessage = function (e) {
-  const time = e.data.time;
+self.onmessage = async (e: MessageEvent<FetchZarrMessage>) => {
```
toloudis, line 36: good use of typing for this

toloudis (approved):
looks great apart from one thing to check with the `minx || null` thing

ShrimpCryptid (approved):
Looks good, no questions beyond what Dan's brought up already

PR #130, opened 2023-07-27
Feature/load volume subsets

### LINKED ISSUE #110: Zarr Loader can load single slices

frasercl:
Adds support for volumes which contain a subset of their full source data, and enables `OMEZarrLoader` to take advantage of these features. This enables several future optimizations, most immediately the ability to load single slices (resolves #110).

### Changes to `OmeZarrLoader`

- `OmeZarrLoader` is now *stateful*: it caches information on its data source (multiscale sizes, metadata, etc.) when `createVolume` is called and assumes that `loadVolumeData` will only be called on this source.
- `createVolume` now also treats the extent fields of `LoadSpec` (`minx`, `maxy`, etc.) differently to `loadVolumeData`.
  - Fields passed into `createVolume` define the size of the `Volume` object relative to the size of the data source, and are cached for use in `loadVolumeData`. This could be used, for example, to load single cell views directly out of full field zarr images.
  - Fields passed into `loadVolumeData` define the size of the subset relative to the volume, as in the screenshot below.
- As before, `createVolume` and `loadVolumeData` pick a scale level based on a fixed memory limitation. This means requesting a smaller subset may mean automatically loading a larger scale level. Extent fields are therefore floats in the range 0-1 rather than pixel values to remain agnostic of scale level.

### Changes to `Volume`

Volumes are still assumed to have a fixed total size (i.e. the size of their bounding box), but may contain a smaller chunk of actual volume data. This chunk will be packed densely into the texture (that is, without inserting gaps to fill the entire volume). This works like this:
- The `ImageInfo` interface has new fields: `vol_size_`(`x`/`y`/`z`), which represents the total size of the volume; and `offset_`(`x`/`y`/`z`), which represents the offset of the subset into the total volume area. For backwards compatibility with `JsonImageInfoLoader`, which loads `ImageInfo` objects and atlased textures directly from a remote source, these new fields are optional. When not specified, `vol_size` fields default to the values of `tile_width`, `tile_height`, and `tiles`, which represent the size of the data; and `offset` fields default to 0.
- While I was there, I added docstrings to all `ImageInfo` fields for clarity, and removed the derivable `atlas_width` and `atlas_height` properties.
- When `ImageInfo` is updated, `Volume` derives new properties, all `Vector3`s in the range 0-1: `contentSize` and `contentOffset`, and the function `getContentCenter` which calculates the center point of the subset.
- `VolumeRenderImpl`s use the above fields to show volumes with subsets properly.

### Steps to verify

I tested this by manually modifying the `LoadSpec` in the test app. The place to do this is in public/index.ts, in `loadVolume`, between `loader.createVolume` and `loader.loadVolumeData` (line 1160). Drop some lines like `loadSpec.maxy = 0.9;` in there and see what happens. It's possible I could modify the dat.gui or tweakpane UI to add some reload capability and make testing easier, but I didn't consider it worth working out how that would work at the moment.

Note that, because channels load one at a time, doing a load of a different extent into a volume briefly puts the volume into a state where the dimensions of some channels don't align with the dimensions of the volume. This isn't ideal, but I didn't find this to be too disruptive in my testing. There's plenty we could do down the line to mitigate this effect, but I considered it out of scope for now.

### Screenshots

image
image

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -927,13 +927,16 @@ function onVolumeCreated(volume: Volume, isTimeSeries = false, frameNumber = 0) 
       // create the main volume and add to view (this is the only place)
       myState.volume = new Volume(myJson);
 
+      const atlasWidth = myJson.tile_width * myJson.cols;
+      const atlasHeight = myJson.tile_height * myJson.rows;
+
       // TODO: this can go in the Volume and Channel constructors!
       // preallocate some memory to be filled in later
       for (let i = 0; i < myState.volume.num_channels; ++i) {
         myState.volume.channels[i].imgData = {
-          data: new Uint8ClampedArray(myJson.atlas_width * myJson.atlas_height),
-          width: myJson.atlas_width,
-          height: myJson.atlas_height,
+          data: new Uint8ClampedArray(atlasWidth * atlasHeight),
+          width: atlasWidth,
+          height: atlasHeight,
         };
         myState.volume.channels[i].volumeData = new Uint8Array(myJson.tile_width * myJson.tile_height * myJson.tiles);
         // TODO also preallocate the Fused data texture
@@ -1100,20 +1103,15 @@ function createTestVolume() {
     height: 64,
     channels: 3,
     channel_names: ["DRAQ5", "EGFP", "SEG_Memb"],
+    // These dimensions are used to prepare the raw volume arrays for rendering as tiled texture atlases
+    // for webgl reasons, it is best for atlas width (cols*tile_width) and height (rows*tile_height)
+    // to be <= 2048 and ideally a power of 2. (adjust other dimensions accordingly)
     rows: 8,
     cols: 8,
     // tiles <= rows*cols, tiles is number of z slices
     tiles: 64,
     tile_width: 64,
     tile_height: 64,
-
-    // These dimensions are used to prepare the raw volume arrays for rendering as tiled texture atlases
-    // for webgl reasons, it is best for atlas_width and atlas_height to be <= 2048
-    // and ideally a power of 2. (adjust other dimensions accordingly)
-    // atlas_width === cols*tile_width
-    atlas_width: 512,
-    // atlas_height === rows*tile_height
-    atlas_height: 512,
 
     pixel_size_x: 1,
     pixel_size_y: 1,
--- a/public/index.ts
+++ b/public/index.ts
@@ -599,7 +599,7 @@ function updateZSliceUI(volume: Volume) {
   const zSlider = document.getElementById("zSlider") as HTMLInputElement;
   const zInput = document.getElementById("zValue") as HTMLInputElement;
 
-  const totalZSlices = volume.z;
+  const totalZSlices = volume.imageInfo.vol_size_z || volume.z;
   zSlider.max = `${totalZSlices}`;
   zInput.max = `${totalZSlices}`;
 }
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -162,7 +162,7 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
   }
 
   public updateVolumeDimensions(): void {
-    const scale = this.volume.scale;
+    const scale = this.volume.normalizedPhysicalSize;
     this.geometryMesh.scale.copy(scale);
     this.setUniform("volumeScale", scale);
     this.boxHelper.box.set(scale.clone().multiplyScalar(-0.5), scale.clone().multiplyScalar(0.5));
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -147,12 +147,9 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
     if (dirtyFlags & SettingsFlags.ROI) {
       // Normalize and set bounds
       const bounds = this.settings.bounds;
-      const boundsNormalized = {
-        bmin: new Vector3(bounds.bmin.x * 2.0, bounds.bmin.y * 2.0, bounds.bmin.z * 2.0),
-        bmax: new Vector3(bounds.bmax.x * 2.0, bounds.bmax.y * 2.0, bounds.bmax.z * 2.0),
-      };
-      this.setUniform("AABB_CLIP_MIN", boundsNormalized.bmin);
-      this.setUniform("AABB_CLIP_MAX", boundsNormalized.bmax);
+
+      this.setUniform("AABB_CLIP_MIN", bounds.bmin);
+      this.setUniform("AABB_CLIP_MAX", bounds.bmax);
       const slice = Math.floor(this.settings.zSlice);
       if (slice >= 0 && slice <= this.volume.z - 1) {
         this.setUniform("Z_SLICE", slice);
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -60,8 +60,17 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
 
     this.geometryTransformNode.add(this.boxHelper, this.geometryMesh);
 
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    const { cols, rows, tile_width, tile_height } = volume.imageInfo;
+    /* eslint-disable @typescript-eslint/naming-convention */
+    const { cols, rows, tile_width, tile_height, tiles } = volume.imageInfo;
+    const {
+      vol_size_x = tile_width,
+      vol_size_y = tile_height,
+      vol_size_z = tiles,
+      offset_x = 0,
+      offset_y = 0,
+      offset_z = 0,
+    } = volume.imageInfo;
+    /* eslint-enable @typescript-eslint/naming-convention */
     const atlasWidth = tile_width * cols;
     const atlasHeight = tile_height * rows;
 
@@ -69,6 +78,8 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
     this.setUniform("ATLAS_Y", rows);
     this.setUniform("textureRes", new Vector2(atlasWidth, atlasHeight));
     this.setUniform("SLICES", volume.z);
+    this.setUniform("SUBSET_SCALE", new Vector3(tile_width / vol_size_x, tile_height / vol_size_y, tiles / vol_size_z));
+    this.setUniform("SUBSET_OFFSET", new Vector3(offset_x / vol_size_x, offset_y / vol_size_y, offset_z / vol_size_z));
     this.setUniform("Z_SLICE", Math.floor(volume.z / 2));
 
     this.channelData = new FusedChannelData(atlasWidth, atlasHeight);
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -61,18 +61,24 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
     this.geometryTransformNode.add(this.boxHelper, this.geometryMesh);
 
     this.setUniform("Z_SLICE", Math.floor((volume.imageInfo.vol_size_z || volume.z) / 2));
-    this.initChannelData();
     this.updateVolumeDimensions();
     this.settings = settings;
     this.updateSettings(settings, SettingsFlags.ALL);
   }
 
-  private initChannelData() {
+  public updateVolumeDimensions(): void {
+    const scale = this.volume.normalizedPhysicalSize;
+    // set scale
+    this.geometryMesh.scale.copy(scale);
+    this.setUniform("volumeScale", scale);
+    this.boxHelper.box.set(scale.clone().multiplyScalar(-0.5), scale.clone().multiplyScalar(0.5));
+
     /* eslint-disable-next-line @typescript-eslint/naming-convention */
     const { cols, rows, tile_width, tile_height } = this.volume.imageInfo;
     const atlasWidth = tile_width * cols;
     const atlasHeight = tile_height * rows;
 
+    // set lots of dimension uniforms
     this.setUniform("ATLAS_X", cols);
     this.setUniform("ATLAS_Y", rows);
     this.setUniform("textureRes", new Vector2(atlasWidth, atlasHeight));
@@ -80,6 +86,7 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
     this.setUniform("SUBSET_SCALE", this.volume.contentSize);
     this.setUniform("SUBSET_OFFSET", this.volume.contentOffset);
 
+    // (re)create channel data
     if (!this.channelData || this.channelData.width !== atlasWidth || this.channelData.height !== atlasHeight) {
       this.channelData?.cleanup();
       this.channelData = new FusedChannelData(atlasWidth, atlasHeight);
@@ -161,14 +168,6 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
     }
   }
 
-  public updateVolumeDimensions(): void {
-    const scale = this.volume.normalizedPhysicalSize;
-    this.geometryMesh.scale.copy(scale);
-    this.setUniform("volumeScale", scale);
-    this.boxHelper.box.set(scale.clone().multiplyScalar(-0.5), scale.clone().multiplyScalar(0.5));
-    this.initChannelData();
-  }
-
   private createGeometry(
     uniforms: ReturnType<typeof sliceShaderUniforms>
   ): [ShapeGeometry, Mesh<BufferGeometry, Material>] {
@@ -243,7 +242,6 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
       return;
     }
     this.uniforms[name].value = value;
-    this.geometryMesh.material.needsUpdate = true;
   }
 
   public setRenderUpdateListener(_listener?: ((iteration: number) => void) | undefined) {
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -35,7 +35,7 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
   private geometryTransformNode: Group;
   private boxHelper: Box3Helper;
   private uniforms: ReturnType<typeof sliceShaderUniforms>;
-  private channelData: FusedChannelData;
+  private channelData!: FusedChannelData;
 
   /**
    * Creates a new Atlas2DSlice.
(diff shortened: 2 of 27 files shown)


public/index.ts
```diff
@@ -599,7 +599,7 @@ function updateZSliceUI(volume: Volume) {
   const zSlider = document.getElementById("zSlider") as HTMLInputElement;
   const zInput = document.getElementById("zValue") as HTMLInputElement;
 
-  const totalZSlices = volume.z;
+  const totalZSlices = volume.imageInfo.vol_size_z || volume.z;
```
frasercl, line 602: This pattern of using `z` as a fallback shows up a bunch when I need the size of the full volume in pixels. (It's required only for z, the special dimension which we currently never downsample.) It's ugly, and indicates that some `Volume` properties might need reshuffling/renaming, and/or that we ought to remove `ImageInfo` as it currently exists from `Volume`.

src/Atlas2DSlice.ts
```diff
@@ -60,16 +60,30 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
+    this.setUniform("textureRes", new Vector2(atlasWidth, atlasHeight));
+    this.setUniform("SLICES", this.volume.imageInfo.vol_size_z || this.volume.z);
+    this.setUniform("SUBSET_SCALE", this.volume.contentSize);
+    this.setUniform("SUBSET_OFFSET", this.volume.contentOffset);
+
+    if (!this.channelData || this.channelData.width !== atlasWidth || this.channelData.height !== atlasHeight) {
+      this.channelData?.cleanup();
+      this.channelData = new FusedChannelData(atlasWidth, atlasHeight);
+    }
+  }
```
frasercl, line 85: This function and an equivalent one on `RayMarchedAtlasVolume` currently run twice when the constructor is called. I should think hard about how to make this no longer the case.

src/Atlas2DSlice.ts
```diff
@@ -103,20 +117,21 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
-        new Vector3(-0.5 * scale.x, -0.5 * scale.y, -0.5 * scale.z),
-        new Vector3(0.5 * scale.x, 0.5 * scale.y, 0.5 * scale.z)
-      );
       // Set rotation and translation
       this.geometryTransformNode.position.copy(this.settings.translation);
       this.geometryTransformNode.rotation.copy(this.settings.rotation);
       this.setUniform("flipVolume", this.settings.flipAxes);
     }
 
+    if (dirtyFlags & SettingsFlags.DATA_SIZE) {
```
kmitcham, line 126: Possibly out of scope for this PR, but could this be simplified by an early return if !dirtyFlags?

kmitcham (approved):

src/View3d.ts
```diff
@@ -218,6 +218,7 @@ export class View3d {
 
   // channels is an array of channel indices for which new data just arrived.
   onVolumeData(volume: Volume, channels: number[]): void {
+    this.image?.setScale(volume.scale, volume.contentSize, volume.contentOffset);
```
frasercl, line 221: `setScale` is currently both a member of `View3d`'s public interface and is called to set the scale of the volume. The volume internally believes that it is a 1x1x1 cube, but is scaled such that its maximum dimension is 1 and its other dimensions are an appropriate ratio. This scale value is stored both as a property of `Volume` and a property of `VolumeRenderSettings`.

It's not clear to me that an external user would see what they expect if they called `setScale`. E.g., if they tried to scale z by 0.5, and the volume were less than half as large in z as in some other dimension, the volume would actually grow in z. Additionally, `VolumeRenderImpl`s use a hodgepodge of reading `scale` from settings vs. reading directly from the volume, and I've just made this worse by following the same discipline for `contentSize` and `contentOffset`.

I think it would make the most sense to consider the `scale` in settings to be a separate user-defined scale factor which is never modified internally. `VolumeRenderImpl`s would always read `Volume` scale directly from the `Volume` object, then additionally scale by this setting. If `Volume` dimensions update (which they never did before this PR), we'd use a separate pathway to update the `VolumeRenderImpl` that doesn't go through `updateSettings` at all. Does this check out? Or could `setScale` even go away completely?

src/Volume.ts
```diff
@@ -242,18 +287,19 @@ export default class Volume {
...
     }
 
+    const sizez = this.imageInfo.vol_size_z || this.z;
     const physSizeMin = Math.min(this.pixel_size[0], this.pixel_size[1], this.pixel_size[2]);
-    const pixelsMax = Math.max(this.imageInfo.width, this.imageInfo.height, this.z);
+    const pixelsMax = Math.max(this.imageInfo.width, this.imageInfo.height, sizez);
     const sx = ((this.pixel_size[0] / physSizeMin) * this.imageInfo.width) / pixelsMax;
     const sy = ((this.pixel_size[1] / physSizeMin) * this.imageInfo.height) / pixelsMax;
-    const sz = ((this.pixel_size[2] / physSizeMin) * this.z) / pixelsMax;
+    const sz = ((this.pixel_size[2] / physSizeMin) * sizez) / pixelsMax;
```
frasercl, line 295: A comment below says that these variables, which are only ever used to set `this.scale`, are the same as `this.normalizedPhysicalSize`. Some `console.log`s show that `this.scale` and `this.normalizedPhysicalSize` are indeed always the same. `scale` is predominantly used in `RayMarchedAtlasVolume` and `Atlas2DSlice`, but `PathTracedVolume` exclusively uses `normalizedPhysicalSize`. Is there a future scenario where they will be different and it will be meaningful to use one or the other? Keeping just one could simplify this function a lot.

src/View3d.ts
frasercl: `setScale` can go away completely as long as `setVoxelSize` remains exposed and works sensibly

src/Volume.ts
frasercl: `scale` can be removed if it is replaced by a unit test that confirms that this is equal

public/index.ts
frasercl: `ImageInfo` in its current form to be isolated to `JsonImageInfoLoader` only and replaced with a more sensible data structure for the modern age of this tool. This may happen in a separate PR to reduce review burden.

src/loaders/IVolumeLoader.ts
```diff
@@ -18,13 +19,79 @@ export class LoadSpec {
...
+  }
+
+  return { ...loadSpec, minx, maxx, miny, maxy, minz, maxz };
+}
+
+/** Shrinks a `LoadSpec` to fit within a provided extent */
+export function fitLoadSpecRegionToExtent<T extends Partial<LoadSpecExtent>>(
+  loadSpec: T,
+  extent: LoadSpecExtent
+): T & LoadSpecExtent {
```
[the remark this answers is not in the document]
frasercl: Having a unified subregion type makes sense and would be nice. Consider using three's `Box3` type.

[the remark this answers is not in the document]
toloudis: secondary suggestion: can keep the fields at the top level of loadspec but have a getBox3():Box3 and setFromBox3(b: Box3) to put the values into an object that you can manipulate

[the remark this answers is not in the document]
frasercl: This can also happen as a separate ticket.

src/workers/FetchZarrWorker.ts
```diff
@@ -40,13 +42,10 @@ self.onmessage = async (e: MessageEvent<FetchZarrMessage>) => {
...
   // build slice spec
-  // assuming ZYX are the last three dimensions:
   const { minx, maxx, miny, maxy, minz, maxz } = e.data.spec;
-  const sliceSpec: (Slice | number)[] = [
-    slice(minz === undefined ? null : minz, maxz),
-    slice(miny === undefined ? null : miny, maxy),
-    slice(minx === undefined ? null : minx, maxx),
-  ];
+  // assuming ZYX are the last three dimensions:
+  // TODO spatial dimension indexes are now avaliable to the worker. use those?
```
toloudis, line 47: is this still a TODO?

src/Volume.ts
```diff
@@ -9,29 +9,56 @@ import { LoadSpec } from "./loaders/IVolumeLoader";
...
-  cols: number;
-  tile_width: number;
-  tile_height: number;
-  atlas_width: number;
-  atlas_height: number;
+
+  times: number;
+  time_scale: number;
+  time_unit: string;
+
```
toloudis, line 57: the addition and removal of fields from ImageInfo has an effect on the proposed refactor to bring ImageInfo back into JsonImageInfoLoader.  Just to note that the JsonImageInfoLoader should have an ImageInfo spec that exactly matches the old json schema.

toloudis:
Now is a good time to write a loader class that can wrap an in-memory data array and return results immediately. Then you can have unit tests that really test edge cases of the subset stuff and have hardcoded confirmation that the api returns what it's supposed to.  Without having to rely on file formats and remote download.  In addition to unit testing, this loader could be used in nbvv/website-3d-cell-viewer.

public/index.ts
```diff
@@ -927,13 +927,16 @@ function onVolumeCreated(volume: Volume, isTimeSeries = false, frameNumber = 0)
       // create the main volume and add to view (this is the only place)
       myState.volume = new Volume(myJson);
 
+      const atlasWidth = myJson.tile_width * myJson.cols;
```
ShrimpCryptid, line 930: This maybe a holdover from an old design choice, but is there a reason for the image info to be named `myJson` here?

toloudis: Yes - this viewer used to only load from one single particular file format, and it was represented as a json file that mapped directly to the ImageInfo data structure.

src/loaders/IVolumeLoader.ts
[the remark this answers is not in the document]
frasercl: Tracked by #132

public/index.ts
frasercl: Tracked by #131

src/workers/FetchZarrWorker.ts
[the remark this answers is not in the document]
frasercl: Workers were formerly explicitly excluded from type checking in tsconfig.json, but this PR introduced a type in FetchZarrWorker which is imported by OmeZarrLoader, so the worker was included. I removed the exclusion and added a type fix suggested by @toloudis: `(self as unknown as Worker).postMessage(...);`

src/workers/FetchZarrWorker.ts
frasercl: Just to-did it

src/Atlas2DSlice.ts
```diff
@@ -60,16 +60,32 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
+  private initChannelData() {
+    /* eslint-disable-next-line @typescript-eslint/naming-convention */
+    const { cols, rows, tile_width, tile_height } = this.volume.imageInfo;
+    const atlasWidth = tile_width * cols;
+    const atlasHeight = tile_height * rows;
+
+    this.setUniform("ATLAS_X", cols);
+    this.setUniform("ATLAS_Y", rows);
+    this.setUniform("textureRes", new Vector2(atlasWidth, atlasHeight));
+    this.setUniform("SLICES", this.volume.imageInfo.vol_size_z || this.volume.z);
```
toloudis, line 79: do we need to worry about distinguishing vol_size_z === 0 versus undefined here?

src/Atlas2DSlice.ts
```diff
@@ -132,14 +140,12 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
-        bmax: new Vector3(bounds.bmax.x * 2.0, bounds.bmax.y * 2.0, bounds.bmax.z * 2.0),
-      };
-      this.setUniform("AABB_CLIP_MIN", boundsNormalized.bmin);
-      this.setUniform("AABB_CLIP_MAX", boundsNormalized.bmax);
+
+      this.setUniform("AABB_CLIP_MIN", bounds.bmin);
+      this.setUniform("AABB_CLIP_MAX", bounds.bmax);
       const slice = Math.floor(this.settings.zSlice);
-      if (slice >= 0 && slice <= this.volume.z - 1) {
+      const sizez = this.volume.imageInfo.vol_size_z || this.volume.z;
```
toloudis, line 147: I see this repeated a couple times - could me a member function on Volume.  getTrueNumberofZSlices bla bla bla

src/Atlas2DSlice.ts
frasercl: I thought about this - arguably yes, but if volume size is 0 you probably have bigger problems and maybe ought to fall back on data size

src/Atlas2DSlice.ts
frasercl: I commented on the repetition of this pattern above and noted that it would hopefully be addressed by #131

src/RayMarchedAtlasVolume.ts
```diff
@@ -173,10 +179,27 @@ export default class RayMarchedAtlasVolume implements VolumeRenderImpl {
...
+    // Set offset
+    this.geometryMesh.position.copy(this.volume.getContentCenter());
+    // Set scale
+    this.geometryMesh.scale.copy(contentSize).multiply(normalizedPhysicalSize);
+    this.setUniform("volumeScale", normalizedPhysicalSize);
+    this.boxHelper.box.set(
+      normalizedPhysicalSize.clone().multiplyScalar(-0.5),
+      normalizedPhysicalSize.clone().multiplyScalar(0.5)
+    );
+    // TODO do tickmarks need to be regenerated here to remain accurate?
```
toloudis, line 196: if the answer is "yes" then we have a bug?

frasercl: I thought about this since leaving this comment and I think for a well-formed dataset the answer is no. If a load changes the size of the subregion, that should have no impact on the size of the bounding box. If a load changes the size of the full volume, it should only be to switch scale levels. In that case, it should still hold the physical size and aspect ratio of the volume constant within a rounding error for all but the smallest and/or silliest datasets.

src/Volume.ts
```diff
@@ -262,15 +300,23 @@ export default class Volume {
...
-    this.scale = new Vector3(sx, sy, sz);
   }
 
   setUnitSymbol(symbol: string): void {
     this.physicalUnitSymbol = symbol;
   }
 
+  /** Computes the center of the volume subset */
+  getContentCenter(): Vector3 {
+    // center point: (contentSize / 2 + contentOffset - 0.5) * normalizedPhysicalSize;
```
toloudis, line 311: I like having the math notation in a comment above.  So much more readable than the threejs functions.

src/loaders/OmeZarrLoader.ts
```diff
@@ -319,34 +348,45 @@ class OMEZarrLoader implements IVolumeLoader {
...
     // got some data, now let's construct the volume.
-    const vol = new Volume(imgdata, loadSpec);
+    const vol = new Volume(imgdata, fullExtentLoadSpec);
     vol.imageMetadata = buildDefaultMetadata(imgdata);
     return vol;
   }
 
-  async loadVolumeData(vol: Volume, onChannelLoaded: PerChannelCallback, explicitLoadSpec?: LoadSpec): Promise<void> {
-    const loadSpec = explicitLoadSpec || vol.loadSpec;
+  loadVolumeData(vol: Volume, onChannelLoaded: PerChannelCallback, explicitLoadSpec?: LoadSpec) {
```
ShrimpCryptid, line 361: Missing return type?

src/loaders/IVolumeLoader.ts
```diff
@@ -6,6 +6,7 @@ export class LoadSpec {
   scene = 0;
   time = 0;
   // sub-region; if not specified, the entire volume is loaded
+  // specify as floats between 0 and 1
```
toloudis, line 9: Just for the record, I still feel like in a LoadSpec, this should be an explicit pixel value and not a relative fraction.

src/loaders/IVolumeLoader.ts
```diff
@@ -18,13 +19,79 @@ export class LoadSpec {
...
+  }
+  if (miny === maxy && miny < sizey) {
+    maxy += 1;
+  }
+  if (minz === maxz && minz < sizez) {
+    maxz += 1;
+  }
+
+  return { ...loadSpec, minx, maxx, miny, maxy, minz, maxz };
+}
```
toloudis, line 57: a unit test would be good to test this function against edge cases to ensure it always does what is intended

src/loaders/OmeZarrLoader.ts
```diff
@@ -172,50 +201,52 @@ async function pickLevelToLoad(
...
         console.log("ERROR: shape length " + shape.length + " does not match axes length " + axes.length);
       }
 
-      const scale5d = getScale(multiscale);
+      const scale5d = getScale(dataset);
 
       const d = new VolumeDims();
-      d.subpath = "";
-      d.shape = [1, 1, 1, 1, 1];
+      d.subpath = loadSpec.subpath;
```
toloudis, line 240: this is making me remember to note that loadSpec really needs to have a `multiresolutionLevel: number` field instead of subpath.  `subpath` should be specific to ome-zarr only and is something that is mapped from the multiresolution level.

frasercl: Worth another issue?

ShrimpCryptid (approved):
There's still a linter error in `src/test/volume.test.js` that needs to be fixed, but other than that I think this PR is good! Thanks for fixing the bug I noted with the 3D/Z view switch.

src/loaders/OmeZarrLoader.ts
toloudis: Yeah that could be a separate ticket

toloudis (approved):

src/loaders/IVolumeLoader.ts
frasercl: I'd certainly be in favor of adding a field to `LoadSpec` that specifies when the values are in pixels, but I don't know how a pixel value would work when the loader is free to choose its own scale level. Would it be based on pixels of the highest-resolution scale level?

toloudis: if a loadSpec's subregion were always in xyz pixels, it would have to always be consistent with the same loadspec's multiresolutionLevel.    Changing levels would require you to calculate the fractional values, and then rescale them to the new level's pixel dimensions.  (Similar to how you resolve from fraction to pixels in order to actually load the data)

toloudis: you could still have a "null" state for the region that just says "give us the whole extents by default"

toloudis: Maybe figuring out how to populate the initial loadspec is a thing -- you'd want to be able to ask  for some dimension info in order to know the ranges you're dealing with.

toloudis: One more thought - if a loader is auto-selecting a multiresolution level, then it has to modify the loadspec in some way before sending the load requests along to the worker.  (It doesn't have to modify it permanently, but it could!)
But the initial incoming loadspec would still specify something explicit.

PR #135, opened 2023-08-09
Refactor: update `ImageInfo`

### LINKED ISSUE #131: Refactor `ImageInfo`
## Use Case

`ImageInfo` is a legacy type whose fields are specific to a single data format and use a naming convention that we have to specifically ask the linter to ignore. As a developer, I'd prefer to work with a clearer, more general type to specify basic dimensions of a `Volume`.

## Solution

- Move the legacy `ImageInfo` into `JsonImageInfoLoader`, where it is required to load data in this format.
- Write a new, more readable, more understandable type to use with all `Volume` objects.
- `JsonImageInfoLoader` should convert to this new type when loading.

### Note

`ImageInfo` has been changed a bit by #130. It should be changed back when moving into the JSON loader to exactly match the JSON the JSON loader loads.

frasercl:
### Problem

Resolves #131: `Volume` receives a type called `ImageInfo` on construction which contains essential information about the volume's dimensions. This type is based on the JSON schema used by `JsonImageInfoLoader`, which is not particularly suited to being a general volume information format:

- It is specific to the way `JsonImageInfoLoader` expects data to be represented (as a series of pre-atlased images, each containing up to 4 volume channels in their r/g/b/a components).
- The JSON is generated from a Python script and uses snake_case. We have to ask eslint to ignore snake_case whenever it's used.
- Because it is read in from JSON directly, it cannot take advantage of Three.js's native vector types.

### Solution

- Create a new, more general `ImageInfo` type, using vectors and camelCase and names that make sense.
- Move the current `ImageInfo` into `JsonImageInfoLoader`, rename it `JsonImageInfo`, revert it to exactly represent the stored JSON, and have the loader convert between the old and new type.
- With more data being represented as vectors, I spotted some shader code that could be more vectorized, so I did that.


src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
-  channels: number;
-  channel_names: string[];
-  channel_colors?: [number, number, number][];
+  version?: string;
+
+  /** XY size of the *original* (not downsampled) volume, in pixels */
+  // If we ever allow downsampling in z, replace with Vector3
+  originalSize: Vector2;
+  /** Number of rows and columns of z-slice tiles (not pixels) in the texture atlas */
+  atlasDims: Vector2;
```
ShrimpCryptid, line 16: I like the vectorizing you did here, it packs related data together more neatly.

src/test/volume.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  expect(v.channelNames.length).to.equal(imgdata.numChannels);
+  expect(v.channels.length).to.equal(imgdata.numChannels);
+
+  const mx = Math.max(Math.max(v.normPhysicalSize.x, v.normPhysicalSize.y), v.normPhysicalSize.z);
+  expect(mx).to.equal(1.0);
+
+  // TODO cleanup?
+  // expect(v.x).to.equal(imgdata.tile_width);
+  // expect(v.y).to.equal(imgdata.tile_height);
+  // expect(v.z).to.equal(imgdata.tiles);
```
ShrimpCryptid, line 63: What's the reasoning for keeping these as comments? Is there an aspect of deleting this that's currently uncertain?

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
+  /** Size of the volume, in pixels */
+  volumeSize: Vector3;
+  /** Size of the currently loaded subregion, in pixels */
+  regionSize: Vector3;
+  /** Offset of the loaded subregion into the total volume, in pixels */
+  regionOffset: Vector3;
```
ShrimpCryptid, line 22: Appreciate the commenting here 👌

src/Volume.ts
ShrimpCryptid, line 16: Something about this comment doesn't really connect for me. 
I *think* I get it (it's how each Z-slice is laid out in the texture atlas), but I'm wondering if there's a more descriptive way to illustrate what you mean here.

src/Volume.ts
```diff
@@ -157,146 +127,110 @@ export default class Volume {
   public loadSpec: LoadSpec;
   public imageMetadata: Record<string, unknown>;
   public name: string;
-  public x: number;
-  public y: number;
-  public z: number;
+  // public x: number;
+  // public y: number;
+  // public z: number;
```
ShrimpCryptid, line 132: delete? (+other member variables below)

src/Volume.ts
```diff
@@ -157,146 +127,110 @@ export default class Volume {
...
   private loaded: boolean;
   /* eslint-disable @typescript-eslint/naming-convention */
-  public num_channels: number;
-  public channel_names: string[];
+  // public num_channels: number;
+  public channelNames: string[];
   public channel_colors_default: [number, number, number][];
-  public pixel_size: [number, number, number];
+  // public pixel_size: [number, number, number];
   /* eslint-enable @typescript-eslint/naming-convention */
```
ShrimpCryptid, line 148: This can be removed now, correct?

src/Volume.ts
```diff
@@ -157,146 +127,110 @@ export default class Volume {
...
-    this.y = this.imageInfo.tile_height;
-    this.z = this.imageInfo.tiles;
+    // TODO clean up these unused properties or replace them with whatever seems most sensible
+    // this.pixel_size = [this.imageInfo.pixel_size_x, this.imageInfo.pixel_size_y, this.imageInfo.pixel_size_z];
+    // this.x = this.imageInfo.tile_width;
+    // this.y = this.imageInfo.tile_height;
+    // this.z = this.imageInfo.tiles;
 
-    this.num_channels = this.imageInfo.channels;
+    // this.num_channels = this.imageInfo.channels;
```
ShrimpCryptid, line 173: I've noticed a lot of commented out lines and TODOs. Should this be added as a separate issue ticket?

public/index.ts
```diff
@@ -1097,40 +1105,33 @@ function goToZSlice(slice: number): boolean {
 function createTestVolume() {
   /* eslint-disable @typescript-eslint/naming-convention */
```
ShrimpCryptid, line 1106: Noted a couple of these left over, might be worth doing a pass through the whole repo to search for any that can be removed

src/constants/shaders/raymarch.frag
ShrimpCryptid, line 1: Enjoying how clean this got with vectors!

ShrimpCryptid:
image
Hey, I'm seeing some artifacting/weird slicing behavior? Is this intentional?

frasercl:
> Hey, I'm seeing some artifacting/weird slicing behavior? Is this a setting I have misconfigured or is this happening for you too?

Could you check that you have the latest version of this branch? I also saw this behavior and fixed it (or attempted to) in 4c143f9.

src/test/volume.test.ts
frasercl: Thought I'd done a thorough enough pass over this PR to clean out all the things I'd tentatively commented out. Apparently not. I'll give it another look over

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
-  /** z offset of the volume subset into the volume */
-  offset_z?: number;
-
-  channels: number;
-  channel_names: string[];
-  channel_colors?: [number, number, number][];
+  version?: string;
+
+  /** XY size of the *original* (not downsampled) volume, in pixels */
+  // If we ever allow downsampling in z, replace with Vector3
```
toloudis, line 13: I would just bump this to be a Vector3 if it is representing multiresolution level 0 dims.  ome-zarr does allow downsampling in z.  There's no guarantee that each multiresolution level will have the same Z size.  (Technically I guess one could also downsample in T and even C, although C would be very strange.)

If that breaks some assumptions elsewhere then we can make that a separate ticket/pr.

src/Volume.ts
toloudis: > Something about this comment doesn't really connect for me. I _think_ I get it (it's how each Z-slice is laid out in the texture atlas), but I'm wondering if there's a more descriptive way to illustrate what you mean here.

I think the variable name `atlasDims` is hard to understand at a glance.  This is a tricky one to name.  `numTilesXY` or somehow using the word tiles instead of atlas might help.   To me, the comment does describe what the variable is, pretty well.  Maybe it could be
```
// number of tiles(z slices) in x and y of the texture atlas
```

ShrimpCryptid:
> > Hey, I'm seeing some artifacting/weird slicing behavior? Is this a setting I have misconfigured or is this happening for you too?
> 
> Could you check that you have the latest version of this branch? I also saw this behavior and fixed it (or attempted to) in [4c143f9](https://github.com/allen-cell-animated/volume-viewer/commit/4c143f966b14fe9dd2be1c37d67febe82b9516c9).

Okay, yes, that fixed it.

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
+  /** Number of rows and columns of z-slice tiles (not pixels) in the texture atlas */
+  atlasDims: Vector2;
+  /** Size of the volume, in pixels */
+  volumeSize: Vector3;
+  /** Size of the currently loaded subregion, in pixels */
+  regionSize: Vector3;
+  /** Offset of the loaded subregion into the total volume, in pixels */
+  regionOffset: Vector3;
+  /** Size of a single *original* (not downsampled) pixel, in spatial units */
+  pixelSize: Vector3;
```
toloudis, line 24: Nit: A better name for this could be `physicalPixelSize` which is what is used in OME-XML.  It's also a bit strange that it is the size for level 0 as opposed to the currently loaded level.

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
-  offset_x?: number;
-  /** y offset of the volume subset into the volume */
-  offset_y?: number;
-  /** z offset of the volume subset into the volume */
-  offset_z?: number;
-
-  channels: number;
-  channel_names: string[];
-  channel_colors?: [number, number, number][];
+  version?: string;
```
toloudis, line 10: I have a feeling version could be removed

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
 
-/* eslint-enable @typescript-eslint/naming-convention */
+export const getDefaultImageInfo = (): ImageInfo => ({
+  name: "",
+  version: "",
+  originalSize: new Vector2(1, 1),
+  atlasDims: new Vector2(1, 1),
+  volumeSize: new Vector3(1, 1, 1),
+  regionSize: new Vector3(1, 1, 1),
+  regionOffset: new Vector3(0, 0, 0),
```
ShrimpCryptid, line 51: Nit: I wonder if references to `region` would be more descriptive as `subregion` instead, since it implies that it's a subset of the larger volume?

src/Volume.ts
```diff
@@ -102,48 +79,41 @@ interface VolumeDataObserver {
...
+ * @property {Vector2} originalSize XY size of the *original* (not downsampled) volume, in pixels
+ * @property {Vector2} atlasDims Number of rows and columns of z-slice tiles (not pixels) in the texture atlas
+ * @property {Vector3} volumeSize Size of the volume, in pixels
+ * @property {Vector3} regionSize Size of the currently loaded subregion, in pixels
+ * @property {Vector3} regionOffset Offset of the loaded subregion into the total volume, in pixels
+ * @property {Vector3} pixelSize Size of a single *original* (not downsampled) pixel, in spatial units
+ * @property {string} spatialUnit Symbol of physical spatial unit used by `pixelSize`
+ * @property {number} numChannels Number of channels
+ * @property {Array.<string>} channelNames Names of each of the channels to be rendered, in order. Unique identifier expected
+ * @property {Array.<Array.<number>>} [channelColors] Colors of each of the channels to be rendered, as an ordered list of [r, g, b] arrays
```
ShrimpCryptid, line 92: What do the brackets around the variable name here do?

frasercl: That's JSDoc for "optional parameter"

src/test/volume.test.ts
frasercl: Done! I'll just comment here and resolve comments below which this pass mopped up

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
   times: number;
-  time_scale: number;
-  time_unit: string;
+  timeScale: number;
+  timeUnit: string;
 
   transform: {
-    translation: [number, number, number];
-    rotation: [number, number, number];
+    translation: Vector3;
```
toloudis, line 37: might be good to add a comment saying what coordinate system the translation has to be in

src/Volume.ts
```diff
@@ -1,96 +1,73 @@
...
+  /** Offset of the loaded subregion into the total volume, in pixels */
+  regionOffset: Vector3;
+  /** Size of a single *original* (not downsampled) pixel, in spatial units */
+  pixelSize: Vector3;
+  /** Symbol of physical spatial unit used by `pixelSize` */
+  spatialUnit: string;
+
+  numChannels: number;
+  channelNames: string[];
+  channelColors?: [number, number, number][];
```
toloudis, line 30: Please add a comment saying whether these are supposed to be 0..255 or 0..1 numbers

src/Volume.ts
```diff
@@ -102,48 +79,41 @@ interface VolumeDataObserver {
...
+  "timeUnit": "hr",
   "transform": {
-    "translation": [5, 5, 1],
-    "rotation": [0, 3.14159, 1.57]
-  }
+    "translation": new Vector3(5, 5, 1),
+    "rotation": new Vector3(0, 3.14159, 1.57),
+  },
   };
  */
```
toloudis, line 118: Nit: could this whole docstring comment be merged/inlined with the declaration of ImageInfo above?  I am not sure if that is a limitation of jsdoc

src/Volume.ts
```diff
@@ -157,146 +127,92 @@ export default class Volume {
...
+  public normRegionOffset: Vector3;
   public physicalUnitSymbol: string;
   public tickMarkPhysicalLength: number;
   private loaded: boolean;
-  /* eslint-disable @typescript-eslint/naming-convention */
-  public num_channels: number;
-  public channel_names: string[];
+  public channelNames: string[];
+  /* eslint-disable-next-line @typescript-eslint/naming-convention */
   public channel_colors_default: [number, number, number][];
```
toloudis, line 142: rename?

src/Volume.ts
```diff
@@ -157,146 +127,92 @@ export default class Volume {
...
-
-    this.num_channels = this.imageInfo.channels;
+    this.validatePixelSize();
 
-    this.channel_names = this.imageInfo.channel_names.slice();
-    this.channel_colors_default = this.imageInfo.channel_colors
-      ? this.imageInfo.channel_colors.slice()
-      : this.channel_names.map((name, index) => getColorByChannelIndex(index));
+    // TODO do we really want this copied value?
+    this.channelNames = this.imageInfo.channelNames.slice();
```
toloudis, line 162: I have no idea why we saved a separate copy here, unless we are modifying the names?

src/Volume.ts
```diff
@@ -157,146 +127,92 @@ export default class Volume {
...
-    if (values[0] > 0) {
-      this.pixel_size[0] = values[0];
-    }
-    if (values[1] > 0) {
-      this.pixel_size[1] = values[1];
-    }
-    if (values[2] > 0) {
-      this.pixel_size[2] = values[2];
-    }
+    // TODO this indicates that maybe pixel size should stick around?
```
toloudis, line 207: I suspect this comment will need more clarification

src/VolumeRenderSettings.ts
```diff
@@ -127,10 +127,10 @@ export class VolumeRenderSettings {
...
 
   public resizeWithVolume(volume: Volume): void {
-    this.zSlice = Math.floor(volume.z / 2);
-    this.specular = new Array(volume.num_channels).fill([0, 0, 0]);
-    this.emissive = new Array(volume.num_channels).fill([0, 0, 0]);
-    this.glossiness = new Array(volume.num_channels).fill(0);
+    this.zSlice = Math.floor(volume.imageInfo.regionSize.z / 2);
+    this.specular = new Array(volume.imageInfo.numChannels).fill([0, 0, 0]);
+    this.emissive = new Array(volume.imageInfo.numChannels).fill([0, 0, 0]);
+    this.glossiness = new Array(volume.imageInfo.numChannels).fill(0);
```
toloudis, line 133: Nit: it would be cool if we could declare volume.imageInfo as read-only somehow, just to protect these deep accesses.  When I see these it makes me want to put wrapper functions so you write volume.getNumChannels()

src/Volume.ts
frasercl: Seems like it's separated because the array may be modified by the `appendEmptyChannel` method

src/Volume.ts
frasercl: This is referring to the fact that I deleted the `Volume` property that held onto pixel size, but this method is explicitly for modifying pixel size, so if we want `ImageInfo` to be a read-only structure it ought to be brought back out.

(I'm still not sure when we'd want to call this method from outside?)

src/Volume.ts
frasercl: Looks to be a relatively trivial change that also means downsampling in Z ought to "just work" should a loader ever want to support it.

src/Volume.ts
frasercl: It may just not be necessary to keep around. It isn't directly associated with an item (either the class `Volume` or the type `ImageInfo`) and doesn't seem to properly link into the description for `Volume`, which looks to be the intent. Anyone working in TypeScript will be prompted by the type checker to first create an `ImageInfo` object, then to add its required properties, at which point they'll encounter the individual property documentation above.

src/VolumeRenderSettings.ts
frasercl: Adding `readonly` before all property values or wrapping the whole type in `Readonly` (which just maps lower-case `readonly` over the whole type) accomplishes this. I did the latter and fixed the couple direct modifications it revealed. The property doesn't make it all the way in to individual vectors: it's still possible to modify individual x/y/z components or call mutating methods. Some more clever type mapping could at least protect direct component modification though, if we want that.

src/Volume.ts
toloudis: Ah, appendEmptyChannel is a remnant of an early attempt to allow augmenting the volume by loading a second volume to append to it.

src/Volume.ts
```diff
@@ -7,21 +7,23 @@ import { LoadSpec } from "./loaders/IVolumeLoader";
...
   /** XY size of the *original* (not downsampled) volume, in pixels */
   // If we ever allow downsampling in z, replace with Vector3
   originalSize: Vector2;
-  /** Number of rows and columns of z-slice tiles (not pixels) in the texture atlas */
-  atlasDims: Vector2;
+  /**
+   * XY dimensions of the texture atlas used by `RayMarchedAtlasVolume` and `Atlas2DSlice`, in number of z-slice
+   * tiles (not pixels). Chosen by the loader to lay out the 3D volume in the squarest possible 2D texture atlas.
+   */
+  atlasTileDims: Vector2;
```
ShrimpCryptid, line 18: Love the improved description here!

ShrimpCryptid (approved):
Looks good, love the really sensible changes and comments here. Thank you, Cameron!

src/VolumeRenderSettings.ts
toloudis: The C++ programmer in me always feels nervous when you have public property accesses 3+ levels deep.  🤷

toloudis (approved):
LGTM!

Issue #132, opened 2023-08-02
Subregion type


frasercl:
## Use Case

`VolumeCache` and `IVolumeLoader` both define separate types to specify a volume subregion, with different functions to operate on them. As a developer, I'd prefer to work with a single unified type in these files and anywhere else we need to deal with subregions.

## Solution

Create a single type for this purpose. Some thoughts:
- Consider using Three's native `Box3` type.
- How should this type relate to `LoadSpec`? Should `LoadSpec` extend this type, convert to this type, or hold this type?

frasercl:
Unit test operations on this type

frasercl:
Resolved by #136 and #137

PR #136, opened 2023-08-22
Subregion type part 1: `LoadSpec`

frasercl:
Addresses #132: there are currently two unconnected areas of the app that care about subregions of the volume, and 
they use unrelated types to express these subregions. Eventually we will connect these two systems, and it would be nice if they both spoke the same language when we do. This PR and its counterpart (#137) convert both to use three.js's native `Box3` type.

This PR converts the subregion fields of `LoadSpec` (`minx`, `maxx`, `miny`, etc., each of type `number`) to a single `subregion` field of type `Box3`, and deals with the consequences of this:
- Replace legacy subregion functions in IVolumeLoader.ts with native `Box3` methods where available, or new functions in VolumeLoaderUtils.ts
- Update `OmeZarrLoader` (the only loader that pays attention to subregion so far) to use `Box3`
- Update some tests for `RequestQueue`


src/loaders/OmeZarrLoader.ts
```diff
@@ -362,23 +345,28 @@ class OMEZarrLoader implements IVolumeLoader {
     }
 
     vol.loadSpec = explicitLoadSpec || vol.loadSpec;
-    const normLoadSpec = fitLoadSpecRegionToExtent(vol.loadSpec, this.maxExtent);
+    const subregion = composeSubregion(vol.loadSpec.subregion, this.maxExtent);
```
toloudis, line 348: does this modify vol.loadSpec in place?

src/loaders/OmeZarrLoader.ts
```diff
@@ -409,15 +398,18 @@ class OMEZarrLoader implements IVolumeLoader {
...
+    const offset = convertSubregionToPixels(vol.loadSpec.subregion, volSize);
     vol.imageInfo = {
       ...vol.imageInfo,
       atlasTileDims: new Vector2(nrows, ncols),
-      volumeSize: new Vector3(vx, vy, vz),
-      subregionSize: new Vector3(tw, th, tz),
-      subregionOffset: new Vector3(offset.minx, offset.miny, offset.minz),
+      volumeSize: volSize,
+      subregionSize: pxSize,
+      subregionOffset: offset.min,
```
toloudis, line 412: Nit: all 3 of the above vars are in pixels... would be good if the names all had px in them, like maybe volSizePx, subregionSizePx, offsetPx?

src/loaders/VolumeLoaderUtils.ts
```diff
@@ -104,6 +105,33 @@ export function estimateLevelForAtlas(spatialDimsZYX: number[][], maxAtlasEdge =
...
+    max.y += 1;
+  }
+  if (min.z === max.z && min.z < size.z) {
+    max.z += 1;
+  }
+
+  return new Box3(min, max);
+}
+
+/** Transform a `Box3` in the 0-1 range into the space of a containing `Box3` */
```
toloudis, line 127: I'm not totally clear on what this does.  Is the containing Box3 in the same 0-1 coordinate space or in pixels?

toloudis (approved):
overall LGTM, just a couple minor questions

src/loaders/VolumeLoaderUtils.ts
frasercl: As used, both `region` and `container` are within 0-1, though only `region` has to be. This is basically getting the subregion of `container` represented by `region`. (Perhaps that should be the comment? idk why I've had so much trouble finding a clear docstring for this function, either here or in its previous form)

src/loaders/OmeZarrLoader.ts
frasercl: Nope, `composeSubregion` returns a new `Box3` without modifying its arguments.

src/loaders/VolumeLoaderUtils.ts
frasercl: i.e. if `container`'s range along the X axis is 0-4 and `region`'s is 0.25-0.5, the result will have range 1-2.

toloudis: ``` Return the subset of container specified by assuming region contains fractional values. i.e. if container's range along the X axis is 0-4 and region's is 0.25-0.5, the result will have range 1-2.```

toloudis: (suggested docstring comment above)

src/loaders/OmeZarrLoader.ts
frasercl: As in the `ImageInfo` fields or the variables they're being set to?

src/loaders/OmeZarrLoader.ts
ShrimpCryptid: It looks like it doesn't, but that brings up that the method comment should probably specify that it returns a new Vector3?

ShrimpCryptid (approved):
This seems like a very logical change and cleans up a bunch of code. I do want to second Dan's comment on updating the method comment for composeSubregion as something to fix before merging. Nice work!

src/loaders/OmeZarrLoader.ts
toloudis: I was thinking `volSize`, `pxSize`, and `offset.min`

src/loaders/OmeZarrLoader.ts
toloudis: I was wondering because it looks like the prior code was using the returned value but the new code uses vol.loadspec after this function call.

frasercl: Ah - the previous code accepted a `LoadSpec`, since that was the region type it had, and duplicated it while modifying the region fields. Functions like `imageIndexFromLoadSpec` care about other fields of `LoadSpec` than the region, so while they could use the output from that function, they aren't worried about whether the region has been modified or not.

PR #137, opened 2023-08-22
Subregion type part 2: `VolumeCache`

frasercl:
Addresses #132: there are currently two unconnected areas of the app that care about subregions of the volume, and 
they use unrelated types to express these subregions. Eventually we will connect these two systems, and it would be nice if they both spoke the same language when we do. This PR and its counterpart (#136) convert both to use three.js's native `Box3` type.

This PR adapts the (as yet unused) `VolumeCache` class to use `Box3` rather than a custom subregion type, and updates its tests. Note that it also introduces a minor behavior change: ranges were previously treated as inclusive on both ends (i.e. a range with an x min of 1 and max of 2 would be understood as a subregion containing rows 1 and 2 of the volume), but are now inclusive in min and exclusive in max (i.e. x min of 1 and max of 2 means a volume containing only row 1).

### DIFF
--- a/src/VolumeCache.ts
+++ b/src/VolumeCache.ts
@@ -48,18 +48,20 @@ function applyDefaultsToRegion(region: Box3 | undefined, size: Vector3): Box3 {
     return new Box3(new Vector3(), size.clone());
   }
 
-  const min = new Vector3(
-    isFinite(region.min.x) ? region.min.x : 0,
-    isFinite(region.min.y) ? region.min.y : 0,
-    isFinite(region.min.z) ? region.min.z : 0
+  const { min, max } = region;
+
+  const newMin = new Vector3(
+    isFinite(min.x) && min.x >= 0 ? min.x : 0,
+    isFinite(min.y) && min.y >= 0 ? min.y : 0,
+    isFinite(min.z) && min.z >= 0 ? min.z : 0
   );
-  const max = new Vector3(
-    region.max.x > 0 ? region.max.x : size.x,
-    region.max.y > 0 ? region.max.y : size.y,
-    region.max.z > 0 ? region.max.z : size.z
+  const newMax = new Vector3(
+    isFinite(max.x) && max.x > 0 ? max.x : size.x,
+    isFinite(max.y) && max.y > 0 ? max.y : size.y,
+    isFinite(max.z) && max.z > 0 ? max.z : size.z
   );
 
-  return new Box3(min, max);
+  return new Box3(newMin, newMax);
 }
 
 const anyComponentGreater = (a: Vector3, b: Vector3) => a.x > b.x || a.y > b.y || a.z > b.z;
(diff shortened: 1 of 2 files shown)


src/VolumeCache.ts
```diff
@@ -222,15 +218,15 @@ export default class VolumeCache {
 
     // Check if entry is already in cache
     for (const existingEntry of entryList) {
-      if (extentIsEqual(existingEntry.extent, extent)) {
+      if (existingEntry.region.equals(region)) {
```
toloudis, line 221: I just thought of something related to two regions being "equal".  Do you consider two regions equal if they result in the same pixels?  Because that allows for some rounding off error with float values that land in between exact integer pixels.

toloudis (approved):
I like using Box3 and Vector3 for these 3d quantities, it's a natural fit.  I think the question of equality of two regions is something we have to have a definite answer for but it won't hold up this PR.

src/VolumeCache.ts
frasercl: For the moment, `VolumeCache` only works in pixels anyways - it should never have to scale from ratios to pixels. (The lack of clarity about when we're working in which space is potentially a pitfall of a unified type. Perhaps I should go over both PRs and make sure the documentation is clear when we expect which?)

src/VolumeCache.ts
```diff
@@ -42,24 +42,27 @@ export type CachedVolume = {
...
+
+  const min = new Vector3(
+    isFinite(region.min.x) ? region.min.x : 0,
+    isFinite(region.min.y) ? region.min.y : 0,
+    isFinite(region.min.z) ? region.min.z : 0
+  );
+  const max = new Vector3(
+    region.max.x > 0 ? region.max.x : size.x,
+    region.max.y > 0 ? region.max.y : size.y,
+    region.max.z > 0 ? region.max.z : size.z
```
ShrimpCryptid, line 59: Should you also be checking for finite-ness here too?

src/VolumeCache.ts
```diff
@@ -21,7 +21,7 @@ type CacheEntry = {
   // TODO allow more types of `TypedArray` to be stored together in the cache?
   data: Uint8Array;
   /** The subset of the volume covered by this entry */
-  extent: CacheEntryExtent;
+  region: Box3;
```
ShrimpCryptid, line 24: Region vs. subregion here? Does subregion make more sense to be consistent with its usage elsewhere?

ShrimpCryptid (approved):
Like the other PR, I like the switch the Box3! I left a few comments that should be pretty quick to address.

src/VolumeCache.ts
toloudis: it definitely helps when there's a comment saying what coordinate space your numbers are supposed to be in.

src/VolumeCache.ts
frasercl: The idea here was based on the fact that `Box3` defaults to min = `Infinity` and max = `-Infinity` for all components, to represent an empty/uninitialized box. I figured the > 0 check would capture this case for max, but I might as well make it more robust and easier to understand anyways.

PR #138, opened 2023-08-22
Fix tickmark size being overwritten after being picked

frasercl:
Corrects a tiny silly bug I introduced in #135: bounding box tick marks are currently always 1 physical unit (e.g. 1μm) regardless of the actual size of the volume, resulting in potentially useless tickmark spacings.

#### Before
image

#### After
image


toloudis (approved):
💯

ShrimpCryptid (approved):
Looks good!

PR #139, opened 2023-09-01
Region-aware loading

frasercl:
Add a basic API to `Volume` to track what data is currently needed to draw the image given the state of the viewer, and use it to enable context-aware single-slice loading (!!).

- `Volume` gets new properties: `loader`, `channelLoadCallback`, and `loadSpecRequired`. The first two of these allow `Volume`s to issue load requests all by themselves. `loadSpecRequired` is a second `loadSpec` that tracks the subset of data which is necessary to display the current image.
- `Volume` gets a new method: `updateRequiredData`. This applies changes to `loadSpecRequired`, and issues a load request when new data is needed.
- `View3d`'s `setTime` method now makes use of `updateRequiredData`
- `VolumeRenderImpl`s call `updateRequiredData` to allow single-slice loading: `Atlas2DSlice` calls it on construction and when z-slice changes to request a single z-slice, and all other implementations call it on construction to reset the required data to the full extent of the volume.
- To accommodate `Volume`'s new data loading responsibilities, `IVolumeLoader` gets a new API: callbacks for when single channels load are passed into `createVolume`, rather than `loadVolumeData`, so that `Volume` can store them. It is still possible to pass a callback to `loadVolumeData`, in which case both will run. But only the callback passed to `createVolume` will be run when a volume decides to load new data via `updateRequiredData`, which should almost always be the right behavior for such a callback.

### DIFF
--- a/src/PathTracedVolume.ts
+++ b/src/PathTracedVolume.ts
@@ -1,4 +1,5 @@
 import {
+  Box3,
   DataTexture,
   Data3DTexture,
   FloatType,
@@ -80,6 +81,7 @@ export default class PathTracedVolume implements VolumeRenderImpl {
   constructor(volume: Volume, settings: VolumeRenderSettings = new VolumeRenderSettings(volume)) {
     this.pathTracingUniforms = pathTracingUniforms();
     this.volume = volume;
+    this.volume.updateRequiredData({ subregion: new Box3(new Vector3(0, 0, 0), new Vector3(1, 1, 1)) });
     this.viewChannels = [-1, -1, -1, -1];
 
     // create volume texture
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -230,10 +230,9 @@ export class View3d {
     this.image?.onChannelAdded(newChannelIndex);
   }
 
-  setTime(volume: Volume, time: number, loader: IVolumeLoader, onChannelLoaded?: PerChannelCallback): void {
-    volume.loadSpec.time = Math.max(0, Math.min(volume.imageInfo.times, time));
+  setTime(volume: Volume, time: number): void {
+    volume.updateRequiredData({ time: Math.max(0, Math.min(time, volume.imageInfo.times)) });
     this.updateTimestepIndicator(volume);
-    loader.loadVolumeData(volume, onChannelLoaded);
   }
 
   /**
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -15,7 +15,6 @@ import { ThreeJsPanel } from "./ThreeJsPanel";
 import lightSettings from "./constants/lights";
 import VolumeDrawable from "./VolumeDrawable";
 import { Light, AREA_LIGHT, SKY_LIGHT } from "./Light";
-import { IVolumeLoader, PerChannelCallback } from "./loaders/IVolumeLoader";
 import Volume from "./Volume";
 import {
   VolumeChannelDisplayOptions,
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -3,7 +3,7 @@ import { Vector2, Vector3 } from "three";
 import Channel from "./Channel";
 import Histogram from "./Histogram";
 import { getColorByChannelIndex } from "./constants/colors";
-import { LoadSpec } from "./loaders/IVolumeLoader";
+import { IVolumeLoader, LoadSpec } from "./loaders/IVolumeLoader";
 
 export type ImageInfo = Readonly<{
   name: string;
@@ -129,6 +129,8 @@ interface VolumeDataObserver {
 export default class Volume {
   public imageInfo: ImageInfo;
   public loadSpec: LoadSpec;
+  public loader?: IVolumeLoader;
+  private loadSpecRequired: LoadSpec;
   public imageMetadata: Record<string, unknown>;
   public name: string;
 
@@ -149,11 +151,17 @@ export default class Volume {
   private volumeDataObservers: VolumeDataObserver[];
   private loaded: boolean;
 
-  constructor(imageInfo: ImageInfo = getDefaultImageInfo(), loadSpec: LoadSpec = new LoadSpec()) {
+  constructor(
+    imageInfo: ImageInfo = getDefaultImageInfo(),
+    loadSpec: LoadSpec = new LoadSpec(),
+    loader?: IVolumeLoader
+  ) {
     this.loaded = false;
     this.imageInfo = imageInfo;
     this.name = this.imageInfo.name;
     this.loadSpec = loadSpec;
+    this.loadSpecRequired = { ...loadSpec, subregion: loadSpec.subregion.clone() };
+    this.loader = loader;
     // imageMetadata to be filled in by Volume Loaders
     this.imageMetadata = {};
 
@@ -200,14 +208,27 @@ export default class Volume {
     this.normRegionOffset = subregionOffset.clone().divide(volumeSize);
   }
 
+  updateRequiredData(required: Partial<LoadSpec>) {
+    this.loadSpecRequired = { ...this.loadSpecRequired, ...required };
+    // if newly required data is not currently contained in this volume...
+    if (
+      this.loadSpecRequired.time !== this.loadSpec.time ||
+      this.loadSpecRequired.scene !== this.loadSpec.scene ||
+      !this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion)
+    ) {
+      // ...clone `loadSpecRequired` into `loadSpec` and load
+      this.loadSpec = { ...this.loadSpecRequired, subregion: this.loadSpecRequired.subregion.clone() };
+      this.loader?.loadVolumeData(this);
+    }
+  }
+
   // we calculate the physical size of the volume (voxels*pixel_size)
   // and then normalize to the max physical dimension
   setVoxelSize(size: Vector3): void {
     // only set the data if it is > 0.  zero is not an allowed value.
-    size.clampScalar(0, Infinity);
-    size.x = size.x || 1.0;
-    size.y = size.y || 1.0;
-    size.z = size.z || 1.0;
+    size.x = size.x > 0 ? size.x : 1.0;
+    size.y = size.y > 0 ? size.y : 1.0;
+    size.z = size.z > 0 ? size.z : 1.0;
     this.physicalPixelSize = size;
 
     this.physicalSize = this.imageInfo.originalSize.clone().multiply(this.physicalPixelSize);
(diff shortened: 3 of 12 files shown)


src/PathTracedVolume.ts
```diff
@@ -80,6 +81,7 @@ export default class PathTracedVolume implements VolumeRenderImpl {
   constructor(volume: Volume, settings: VolumeRenderSettings = new VolumeRenderSettings(volume)) {
     this.pathTracingUniforms = pathTracingUniforms();
     this.volume = volume;
+    this.volume.updateRequiredData({ subregion: new Box3(new Vector3(0, 0, 0), new Vector3(1, 1, 1)) });
```
toloudis, line 84: Pausing to think about this... creating a PathTracedVolume fires off a load request as a side effect.  I wonder if this should be more explicit (like call it after the ctor finishes and returns) rather than tucked inside here.

src/Volume.ts
```diff
@@ -129,6 +129,9 @@ interface VolumeDataObserver {
 export default class Volume {
   public imageInfo: ImageInfo;
   public loadSpec: LoadSpec;
+  public loader?: IVolumeLoader;
```
toloudis, line 132: what does it mean if there's no loader?

src/Volume.ts
```diff
@@ -200,14 +209,27 @@ export default class Volume {
...
+    // if newly required data is not currently contained in this volume...
+    if (
+      this.loadSpecRequired.time !== this.loadSpec.time ||
+      this.loadSpecRequired.scene !== this.loadSpec.scene ||
+      !this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion)
+    ) {
+      // ...clone `loadSpecRequired` into `loadSpec` and load
+      this.loadSpec = { ...this.loadSpecRequired, subregion: this.loadSpecRequired.subregion.clone() };
+      this.loader?.loadVolumeData(this);
+    }
```
toloudis, line 223: something feels weird about the naming of loadSpec vs loadSpecRequired.   I don't really have a good suggestion for renaming. maybe loadSpecToRequest?
Also, I know it's premature but can't help but start thinking about how this code interacts with cache.  An early naive implementation could assume that cache will only have either single slices or full volume and can just check dimensions+channel for exact matches.

src/Volume.ts
```diff
@@ -248,9 +270,8 @@ export default class Volume {
     if (this.channels.every((element) => element.loaded)) {
       this.loaded = true;
     }
-    for (let i = 0; i < this.volumeDataObservers.length; ++i) {
-      this.volumeDataObservers[i].onVolumeData(this, batch);
-    }
+    batch.forEach((channelIndex) => this.channelLoadCallback?.(this.loadSpec.url, this, channelIndex));
```
toloudis, line 273: probably should document somewhere (in IVolumeLoader) the order in which the two possible channel loaded callbacks will be called.

src/loaders/IVolumeLoader.ts
```diff
@@ -10,11 +10,11 @@ export class LoadSpec {
...
   subregion = new Box3(new Vector3(0, 0, 0), new Vector3(1, 1, 1));
+}
 
-  toString(): string {
-    const { min, max } = this.subregion;
-    return `${this.url}:${this.subpath}${this.scene}:${this.time}:x(${min.x},${max.x}):y(${min.y},${max.y}):z(${min.z},${max.z})`;
-  }
+export function loadSpecToString(spec: LoadSpec): string {
+  const { min, max } = spec.subregion;
+  return `${spec.url}:${spec.subpath}${spec.scene}:${spec.time}:x(${min.x},${max.x}):y(${min.y},${max.y}):z(${min.z},${max.z})`;
```
toloudis, line 17: is this just to keep loadSpec as a data type that doesn't have any functions as members?

src/View3d.ts
ShrimpCryptid, line 1: I like that View3D had some dependencies removed! It's nice that it's getting decoupled from IVolumeLoader, though I suppose the tradeoff is Volume and IVolumeLoader becoming tighter.

src/Volume.ts
```diff
@@ -200,14 +209,27 @@ export default class Volume {
     this.normRegionOffset = subregionOffset.clone().divide(volumeSize);
   }
 
+  updateRequiredData(required: Partial<LoadSpec>) {
```
ShrimpCryptid, line 212: This made me go down a typescript rabbit hole haha

src/Volume.ts
```diff
@@ -200,14 +209,27 @@ export default class Volume {
...
   // and then normalize to the max physical dimension
   setVoxelSize(size: Vector3): void {
     // only set the data if it is > 0.  zero is not an allowed value.
-    size.clampScalar(0, Infinity);
-    size.x = size.x || 1.0;
-    size.y = size.y || 1.0;
-    size.z = size.z || 1.0;
+    size.x = size.x > 0 ? size.x : 1.0;
+    size.y = size.y > 0 ? size.y : 1.0;
+    size.z = size.z > 0 ? size.z : 1.0;
```
ShrimpCryptid, line 232: What does this do if it receives a Vector3 with infinite values?

ShrimpCryptid (approved):
I had one question I added, but otherwise this change looks solid. Thank you!

src/PathTracedVolume.ts
frasercl: I did consider adding a method to `VolumeRenderImpl` to contain these side effects, so that it would be possible e.g. to construct an instance without that instance assuming it immediately "owns" the rendering side of its associated `Volume`. I picked "faster" over "cleaner" with this PR, but this seems like a sensible place to go back and make it nice.

src/Volume.ts
frasercl: My thought here was to preserve the ability to just directly construct a volume and add data to it via methods, in which case any side effects from the whole `updateRequiredData` system would get captured by the null check on this loader and just not happen.

src/loaders/IVolumeLoader.ts
frasercl: Right. The spreads I use to clone or partially update `LoadSpec`s (i.e. `this.loadSpecRequired = { ...this.loadSpecRequired, ...required };`) don't preserve prototypes, which was causing some tests on `RequestQueue` to fail.

src/Volume.ts
frasercl: I believe I first spotted this pattern in the signature for `setState` in React class components. The argument to that method has type `Partial<State>` in order to capture the idea that you can pass in an incomplete `State` object containing only fields which need updating. Same idea here.

src/Volume.ts
frasercl: The behavior here is hopefully identical to the previous code, just a bit more efficient. This code will leave `+Infinity` alone and turn `-Infinity` into 1.

src/PathTracedVolume.ts
toloudis: I was picturing 
```
foo = new PathTracedVolume(volume, ...);
volume.updateRequiredData(...);
```
which hopefully is not a huge difference in code.
I'm not totally 100% which is better but it seems that loosening this coupling might enable more flexibility when it comes to requesting new data from a variety of other kinds of user actions.  If you want to wait till the next phase to reconsider it that makes sense too.

frasercl: My initial concern with that is it makes it less clear which class is responsible for calling `updateRequiredData`. Currently, this method is only ever touched by `VolumeRenderImpl`s, whether during construction or in response to a state update. (The only state update that calls it for now is in `Atlas2DSlice` when the z slice changes, but there may be many more clever optimizations to make using this method in the future.) Moving the method call up splits responsibility for updating required data between `VolumeRenderImpl`s and `VolumeDrawable`.

...though I'm realizing the only further optimizations I can easily think of are awareness of clipping box and visible channels, whose handling is basically identical across `VolumeRenderImpl`s. So maybe it makes more sense than I thought to split responsibilities like that?

src/loaders/IVolumeLoader.ts
```diff
@@ -49,18 +49,21 @@ export interface IVolumeLoader {
 
   /**
    * Create an empty `Volume` from a `LoadSpec`, which must be passed to `loadVolumeData` to begin loading.
+   * Optionally pass a callback to respond whenever new channel data is loaded into the volume.
    *
-   * May cache some values for use on only the next call to `loadVolumeData`; callers should guarantee that the next
-   * call to this loader's `loadVolumeData` is made on the returned `Volume` before the volume's state is changed.
+   * Loaders are allowed to assume that they will only be called on a single data source, in order to cache
+   * information about that source. Once this method has been called, every subsequent call to it or
+   * `loadVolumeData` should reference the same source.
```
toloudis, line 56: Good to note. Interesting/scary imagining trying to enforce this in code.

toloudis (approved):
Overall LGTM!  Good progress.  One optional comment to do now or consider doing later as the code evolves / adds polish. (the side effect VolumeRenderImpl thing)

src/loaders/IVolumeLoader.ts
frasercl: This kind of dovetails with another concern I have about this PR: `loadSpecRequired` has no reason to carry around a URL. In a world where loading from one place and then another may cause problems more often than not, it might make sense to remove URL from `LoadSpec` and instead pass it into loader constructors.

toloudis: I kinda like that idea.

toloudis: That could be a separate PR and combined with changing the subpath to be multiresolution level.

src/PathTracedVolume.ts
toloudis: A few thoughts...

On state updates, I can imagine that multiple dirtyFlags could cause updateRequiredData to be called.  So one change could be to consolidate requests as the dirty flags are checked and then make sure there's nothing redundant or conflicting.  
But also, setTime causes updateRequiredData to be called currently from a much higher level. 

I think it could be much more convenient to be "allowed" to call updateRequiredData from anywhere that has the Volume.  That's assuming the drawable stuff can just "pick up" whatever data there is and draw with it -- which could be one reason to keep the updates held tightly inside the renderimpls. The reason being we don't want to draw with partially old data and partially new.  But of course there are other more explicit ways to do it.  

I generally favor more explicit and separated responsibilities.  Looser coupling, simpler code modules, more predictable behaviors.  I feel like for maintainability it's easier to deal with "do A then do B" as opposed to "do A (and B automatically happened behind the scenes)".

Bottom line, now is the time to try either way.  The code is at a reasonably simple point where it's not a huge cost to switch things around.  So take your pick!

toloudis: I also think that the next round of changes will be more the time to test these assumptions as things go up one level of complexity (caching / prefetching)

Issue #134, opened 2023-08-02
Rework `LoadSpec`


frasercl:
## Use Case

As a developer, the `LoadSpec` type doesn't capture all settings that are relevant when loading a volume.

## Solution

- Remove the `subpath` field, replace with something like `resolutionLevel`
- Consider giving extent fields in pixels at provided resolution level, rather than 0-1 floats
- Add a list of channel indexes to load

toloudis:
Consider refactor to remove url from LoadSpec and provide it to stateful IVolumeLoaders in ctor.  One goal being to help enforce that loadspecs can be the "original" load spec, and also a requested new loadspec that is intended to use the same data source.   So a possible alternative is to have two different flavors of LoadSpec.  LoadSpec and LoadSpecRequest? or even LoadSpec and LoadSpecWithUrl!

Issue #126, opened 2023-07-11
Prefetch and cache slices


toloudis:
## Use Case

Playback in XY mode is too slow if we are fetching single slices.

## Solution

We can optimize by prefetching and caching slices.  When time or z slider is moved, prefetch some range of slices (like the next 20, for example.  Or possibly previous 5 / next 15 etc).  Cancel any previous prefetching that is out of range.  XY rendering can then get its data from the local cache.

Choose a simple heuristic to get things working. We can analyze / profile / refine / improve in other tickets once any kind of prefetching is working.

toloudis:
Not sure if there are any changes required for website-3d-cell-viewer to support this.  Would be ideal if it could be handled entirely down at this level.

PR #140, opened 2023-09-13
Rework `LoadSpec` 1/Use Cache 1/Time series tweaks

frasercl:
### TL;DR:

- Removed `url` property from `LoadSpec` in favor of passing URLs into `IVolumeLoader` constructors
- Replace test viewer's outdated time series playback with newer API
  - Allow `JsonImageInfoLoader` to accept an array of URLs to support JSON time series in the new API
- Add caching to `OMEZarrLoader` (but no other loaders so far)
- Made the API changes described in the list at the bottom of this comment

<hr />

Strap in, this one got away from me a bit.

The original intent of this PR was to address a comment in #134: now that loaders are stateful and `Volume`s may submit load requests to them at will, it makes sense to remove the `url` property from `LoadSpec` and set it in loader constructors instead.

I made this change, but this presented a problem for the example viewer. It used its own ad hoc time series playback mechanism, where a collection of JSON atlas volumes are treated as frames of a single time series. (Ideally I would have removed this when I added first class support for time series in #104, but I didn't have a means to also support JSON time series then.) All these volumes were previously loaded by the same `JsonImageInfoLoader`, but under this new stateful loader paradigm each frame would have required a new loader. In order to avoid this awkward situation, I replaced the old time series mechanism with the current `setTime` API, and supported the JSON time series by allowing `JsonImageInfoLoader` to accept an array of URLs representing time series frames. This has the added bonus of enabling new region-aware loading features in zarr time series, since the old mechanism would throw out and recreate volumes.

In order to preserve as much of the original behavior of the time series JSON as possible, my original intent was to also add cache support in this PR (partially addressing #126). I began this process with `OMEZarrLoader`, but found that `JsonImageInfoLoader` would have required more intensive changes, since its data arrives in atlased batches of (up to) four channels and is never converted to a cache-compatible format in the loader. So only zarrs get cached for now, but this at least demonstrates the API I want to extend to the other loaders and allows us to try out cache performance in the test viewer. (Note though: JSON time series frames are no longer prefetched!)

<hr />

I also made the following stray API tweaks:
- Removed the `url` parameter from the `PerChannelCallback` type. All callbacks passed to loaders ignored this parameter, both in the test viewer and in website-3d-cell-viewer.
- `setTime` accepts an optional `PerChannelCallback` like a loader method call would.
- Volumes now properly mark themselves as "unloaded" when loading new data
- Rename: `CachedVolume` -> `CacheStore` for clarity

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -854,39 +854,7 @@ function loadImageData(jsonData, volumeData) {
   return vol;
 }
 
-// function cacheTimeSeriesImageData(jsonData, frameNumber) {
-//   const vol = new Volume(jsonData);
-//   myState.timeSeriesVolumes[frameNumber] = vol;
-//   return vol;
-// }
-
-function onChannelDataArrived(v: Volume, channelIndex: number, frameNumber = 0) {
-  // if (cacheForTimeSeries) {
-  //   if (v.isLoaded()) {
-  //     const currentVol = cacheTimeSeriesImageData(v.imageInfo, frameNumber);
-
-  //     console.log("currentVol with name " + v.name + " is loaded");
-  //     myState.numberOfVolumesCached++;
-  //     if (myState.numberOfVolumesCached >= myState.totalFrames) {
-  //       console.log("ALL FRAMES RECEIVED");
-  //     }
-
-  //     copyVolumeToVolume(v, currentVol);
-  //     if (frameNumber === 0) {
-  //       copyVolumeToVolume(v, myState.volume);
-  //       // has assumption that only 3 channels
-  //       view3D.onVolumeData(myState.volume, [0, 1, 2]);
-  //       view3D.updateActiveChannels(myState.volume);
-  //       view3D.updateLuts(myState.volume);
-
-  //       myState.volume.setIsLoaded(true);
-  //     } else {
-  //       //copyVolumeToVolume(v, currentVol);
-  //     }
-  //     return;
-  //   }
-  // }
-
+function onChannelDataArrived(v: Volume, channelIndex: number) {
   const currentVol = v; // myState.volume;
 
   currentVol.channels[channelIndex].lutGenerator_percentiles(0.5, 0.998);
@@ -904,58 +872,6 @@ function onChannelDataArrived(v: Volume, channelIndex: number, frameNumber = 0) 
 
 function onVolumeCreated(volume: Volume) {
   const myJson = volume.imageInfo;
-  // if (isTimeSeries) {
-  //   if (frameNumber === 0) {
-  //     // create the main volume and add to view (this is the only place)
-  //     myState.volume = new Volume(myJson);
-
-  //     const atlasWidth = myJson.subregionSize.x * myJson.atlasTileDims.x;
-  //     const atlasHeight = myJson.subregionSize.y * myJson.atlasTileDims.y;
-
-  //     // TODO: this can go in the Volume and Channel constructors!
-  //     // preallocate some memory to be filled in later
-  //     for (let i = 0; i < volume.imageInfo.numChannels; ++i) {
-  //       myState.volume.channels[i].imgData = {
-  //         data: new Uint8ClampedArray(atlasWidth * atlasHeight),
-  //         width: atlasWidth,
-  //         height: atlasHeight,
-  //       };
-  //       myState.volume.channels[i].volumeData = new Uint8Array(
-  //         myJson.subregionSize.x * myJson.subregionSize.y * myJson.subregionSize.z
-  //       );
-  //       // TODO also preallocate the Fused data texture
-  //     }
-
-  //     view3D.removeAllVolumes();
-  //     view3D.addVolume(myState.volume);
-  //     setInitialRenderMode();
-  //     // first 3 channels for starters
-  //     for (let ch = 0; ch < myState.volume.imageInfo.numChannels; ++ch) {
-  //       view3D.setVolumeChannelEnabled(myState.volume, ch, ch < 3);
-  //     }
-  //     view3D.setVolumeChannelAsMask(myState.volume, myJson.channelNames.indexOf("SEG_Memb"));
-  //     view3D.updateActiveChannels(myState.volume);
-  //     view3D.updateLuts(myState.volume);
-  //     view3D.updateLights(myState.lights);
-  //     view3D.updateDensity(myState.volume, densitySliderToView3D(myState.density));
-  //     view3D.updateExposure(myState.exposure);
-  //     // apply a volume transform from an external source:
-  //     if (myJson.transform) {
-  //       const alignTransform = myJson.transform;
-  //       view3D.setVolumeTranslation(
-  //         myState.volume,
-  //         myState.volume.voxelsToWorldSpace(alignTransform.translation.toArray())
-  //       );
-  //       view3D.setVolumeRotation(myState.volume, alignTransform.rotation.toArray());
-  //     }
-  //   }
-
-  //   updateTimeUI(myState.volume);
-  //   updateZSliceUI(myState.volume);
-  //   showChannelUI(myState.volume);
-  //   return;
-  // }
-
   myState.volume = volume;
 
   view3D.removeAllVolumes();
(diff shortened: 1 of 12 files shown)


src/loaders/IVolumeLoader.ts
```diff
@@ -35,7 +34,7 @@ export class VolumeDims {
  * @param {Volume} volume
  * @param {number} channelindex
  */
-export type PerChannelCallback = (imageurl: string, volume: Volume, channelIndex: number) => void;
+export type PerChannelCallback = (volume: Volume, channelIndex: number) => void;
```
toloudis, line 37: I think the reason the url is part of the signature here was so that the function implementation can make sure that the data is arriving for the intended target.  This was in part solving the problem of trying to load one volume and then requesting a totally different image while the first one was in flight.  Hopefully our request queue class will be the solution for this. It's definitely a scenario we need to be aware of.

src/View3d.ts
```diff
@@ -229,8 +230,9 @@ export class View3d {
     this.image?.onChannelAdded(newChannelIndex);
   }
 
-  setTime(volume: Volume, time: number): void {
-    volume.updateRequiredData({ time: Math.max(0, Math.min(time, volume.imageInfo.times)) });
+  setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): void {
+    const timeClamped = Math.max(0, Math.min(time, volume.imageInfo.times - 1));
```
toloudis, line 234: callers should also be responsible for passing correct times in; thinking about playback and looping.

src/Volume.ts
```diff
@@ -200,6 +200,18 @@ export default class Volume {
     this.volumeDataObservers = [];
   }
 
+  private setUnloaded() {
+    this.loaded = false;
+    // TODO this will cause problems once it is possible to skip loading some channels. Come back to this then.
+    this.channels.forEach((channel) => {
+      channel.loaded = false;
```
toloudis, line 207: do you want to empty out the channel's data arrays here too?

src/loaders/IVolumeLoader.ts
```diff
@@ -65,5 +64,6 @@ export interface IVolumeLoader {
   // TODO make this return a promise that resolves when loading is done?
   // TODO this is not cancellable in the sense that any async requests initiated here are not stored
   // in a way that they can be interrupted.
+  // TODO explicitly passing a `LoadSpec` is now rarely useful. Remove?
```
toloudis, line 67: very soon we should re-examine all the todos above here too

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -90,24 +90,31 @@ const convertImageInfo = (json: JsonImageInfo): ImageInfo => ({
...
-    if (!this.imageInfo) {
-      const response = await fetch(loadSpec.url);
-      const imageInfo = (await response.json()) as JsonImageInfo;
+    this.time = 0;
+  }
 
-      imageInfo.pixel_size_unit = imageInfo.pixel_size_unit || "μm";
-      this.imageInfo = imageInfo;
+  async getJsonImageInfo(loadSpec: LoadSpec): Promise<JsonImageInfo> {
+    const response = await fetch(this.urls[loadSpec.time]);
```
toloudis, line 108: I like this.  It also makes me wonder if we should support something similar for other file formats when the T samples have been split among many files.

frasercl: I mean basically the only reason I didn't do this for `TiffLoader` as well is that none of the example datasets required it.

src/View3d.ts
frasercl: Should this function throw an error or return some error value rather than quietly fixing out-of-bounds time values for the caller?

src/Volume.ts
frasercl: I think those can stay, rather than risking trying to redraw an empty channel while new data is being loaded for it.

The problem this function solves btw was with functions that trigger loading and receive a per-channel callback. It's common for those callbacks to call `Volume.isLoaded` to ensure that some code runs only once, after every channel has loaded. But if `Volume` had already been loaded into at least once, `Volume.isLoaded()` was coming back true on every callback, not just the last one.

src/View3d.ts
toloudis: I think it's ok if the responsibility is shared.  This step you have here seems ok to prevent catastrophic problems.  (Tradeoff: if you handle error gracefully, sometimes it hides that there was something actually wrong.  If you raise an exception, you clearly know something is wrong but you crash the app.)

src/VolumeCache.ts
```diff
@@ -30,14 +30,14 @@ type CacheEntry = {
   parentArr: CacheEntry[];
 };
 
-type CachedVolumeScale = {
+type CacheStoreScale = {
```
toloudis, line 33: suggested: `CacheStoreMultiscaleLevel`.  Pyramid is also a reasonable term that is used for Multiscale or Multiresolution.  Personally I prefer Multiresolution but any of them work fine

toloudis (approved):
Nothing needs urgent attention.  This is great progress, looking forward to the next iteration!

public/index.ts
```diff
@@ -996,94 +894,49 @@ function onVolumeCreated(volume: Volume, isTimeSeries = false, frameNumber = 0)
...
+    // TODO would be real nice if this were an `await`-able promise instead...
+    view3D.setTime(myState.volume, nextFrame, (vol) => {
+      if (vol.isLoaded()) {
+        myState.currentFrame = nextFrame;
+        onNewFrameCallback();
+
+        if (myState.isPlaying) {
+          const timeLoading = Date.now() - myState.lastFrameTime;
+          myState.timerId = window.setTimeout(loadNextFrame, PLAYBACK_INTERVAL - timeLoading);
+        }
```
ShrimpCryptid, line 919: Should this check happen before setting the state (or even before calling `view3d.setTime`)? If playback is stopped, you might see the frame jump suddenly.

src/Volume.ts
```diff
@@ -270,7 +283,7 @@ export default class Volume {
     if (this.channels.every((element) => element.loaded)) {
       this.loaded = true;
     }
-    batch.forEach((channelIndex) => this.channelLoadCallback?.(this.loadSpec.url, this, channelIndex));
+    batch.forEach((channelIndex) => this.channelLoadCallback?.(this, channelIndex));
```
ShrimpCryptid, line 286: Just asking for my own clarification, does the `function?.(args)` mean you're only calling the function if it exists?

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -129,17 +137,22 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
+      this.jsonInfo = await this.getJsonImageInfo(loadSpec);
+    }
     // This regex removes everything after the last slash, so the url had better be simple.
-    const urlPrefix = loadSpec.url.replace(/[^/]*$/, "");
-    this.imageArray.forEach((element) => {
-      element.name = urlPrefix + element.name;
-    });
-    JsonImageInfoLoader.loadVolumeAtlasData(vol, this.imageArray, onChannelLoaded);
+    const urlPrefix = this.urls[this.time].replace(/[^/]*$/, "");
+    const images = this.jsonInfo?.images.map((element) => ({ ...element, name: urlPrefix + element.name }));
```
ShrimpCryptid, line 151: Might be good to remove any trailing slashes before doing this?
ex: `https://some-example.com/json_file.json/`

ShrimpCryptid (approved):
Left some minor comments but otherwise looks good!

public/index.ts
frasercl: The new recursive structure of this function is a bit harder to reason about (I had to think about this comment for a second before figuring out the answer too)... The callback runs after `view3D.setTime` has completed loading in new frame data, so the part outside the check is dealing with that, and the part inside it is about whether we're still playing and should do it again.

Zooming out: while playback is happening, this recursive loop is either waiting on a timeout or loading new data. If we're waiting on a timeout, `pause` will cancel that timeout, so there's no need to check `myState.isPlaying` in the outer scope of this function. There's no meaningful way to cancel data loading at the moment (so the frame may jump by one after the user has hit pause), but we can prevent another timeout from being fired in the first place. That's what the check is doing.

src/Volume.ts
frasercl: Yep!

PR #141, opened 2023-09-25
fix memory leak

toloudis:
Fixes a memory leak.  
Reproduce:
1. select a time series and play through time (repeatedly)
2. either observe an eventual performance degradation and exception, or observe the javascript heap graph growing using devtools profiler.

The fix is to use threejs's `dispose` in key places that hold references to the data textures that contain the volume pixels.    

An alternative future code path was already presented in a TODO comment:  allocate the Materials to be used in `gpuFuse` as members of Channel, so that they can be encapsulated alongside their DataTextures and created/destroyed in the same place.

### DIFF
--- a/src/Channel.ts
+++ b/src/Channel.ts
@@ -105,20 +105,25 @@ export default class Channel {
     return this.imgData.data[offset];
   }
 
-  // give the channel fresh data and initialize from that data
-  // data is formatted as a texture atlas where each tile is a z slice of the volume
-  public setBits(bitsArray: Uint8Array, w: number, h: number): void {
-    this.imgData = { data: new Uint8ClampedArray(bitsArray.buffer), width: w, height: h };
+  private rebuildDataTexture(data: Uint8ClampedArray, w: number, h: number): void {
     if (this.dataTexture) {
       this.dataTexture.dispose();
     }
-    this.dataTexture = new DataTexture(this.imgData.data, w, h);
+    this.dataTexture = new DataTexture(data, w, h);
     this.dataTexture.format = RedFormat;
     this.dataTexture.type = UnsignedByteType;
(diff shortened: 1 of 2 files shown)


kmitcham (approved):

src/FusedChannelData.ts
```diff
@@ -143,15 +148,32 @@ export default class FusedChannelData {
...
 
     // webgl draw one mesh per channel to fuse.  clear texture to 0,0,0,0
+
+    this.fuseScene.traverse(function (node) {
+      if (node instanceof Mesh) {
+        // materials were holding references to the channel data textures
+        // causing mem leak so we must dispose before clearing the scene
+        node.material.dispose();
+      }
+    });
```
frasercl, line 158: optional nit: could be an arrow func

src/Channel.ts
```diff
@@ -109,6 +109,9 @@ export default class Channel {
   // data is formatted as a texture atlas where each tile is a z slice of the volume
   public setBits(bitsArray: Uint8Array, w: number, h: number): void {
     this.imgData = { data: new Uint8ClampedArray(bitsArray.buffer), width: w, height: h };
+    if (this.dataTexture) {
+      this.dataTexture.dispose();
+    }
     this.dataTexture = new DataTexture(this.imgData.data, w, h);
     this.dataTexture.format = RedFormat;
     this.dataTexture.type = UnsignedByteType;
```
frasercl, line 117: There's a fair bit of `dataTexture` initialization (and now destruction) code here and in `packToAtlas` that maaaaybe could be broken out to its own private method? Honestly could go either way on this

frasercl (approved):
Glad to see an easy fix to this!

PR #142, opened 2023-09-25
Jsonloader with cache

toloudis:
This is a bit "quick and dirty" but I wanted to see our timelapse volume play with caching.  That timelapse is stuck in our old json+png format. This basically uses the blueprint already set up in the Zarr loader.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -24,7 +24,7 @@ import { OpenCellLoader } from "../src/loaders/OpenCellLoader";
 import { State, TestDataSpec } from "./types";
 import { getDefaultImageInfo } from "../src/Volume";
 
-const CACHE_MAX_SIZE = 500_000_000;
+const CACHE_MAX_SIZE = 1_000_000_000;
 const PLAYBACK_INTERVAL = 80;
 
 const TEST_DATA: Record<string, TestDataSpec> = {
@@ -859,7 +859,7 @@ function onChannelDataArrived(v: Volume, channelIndex: number) {
 
   currentVol.channels[channelIndex].lutGenerator_percentiles(0.5, 0.998);
   view3D.onVolumeData(currentVol, [channelIndex]);
-  view3D.setVolumeChannelEnabled(currentVol, channelIndex, channelIndex < 3);
+  //view3D.setVolumeChannelEnabled(currentVol, channelIndex, channelIndex < 3);
 
   view3D.updateActiveChannels(currentVol);
   view3D.updateLuts(currentVol);
@@ -1005,7 +1005,7 @@ function createLoader(data: TestDataSpec): IVolumeLoader {
       for (let t = 0; t <= times; t++) {
         frameUrls.push(data.url.replace("%%", t.toString()));
       }
-      return new JsonImageInfoLoader(frameUrls);
+      return new JsonImageInfoLoader(frameUrls, volumeCache);
     default:
       throw new Error("Unknown loader type: " + data.type);
   }
--- a/src/loaders/JsonImageInfoLoader.ts
+++ b/src/loaders/JsonImageInfoLoader.ts
@@ -1,8 +1,9 @@
 import { Vector2, Vector3 } from "three";
 
 import { IVolumeLoader, LoadSpec, PerChannelCallback, VolumeDims } from "./IVolumeLoader";
-import { buildDefaultMetadata } from "./VolumeLoaderUtils";
+import { buildDefaultMetadata, convertSubregionToPixels } from "./VolumeLoaderUtils";
 import Volume, { ImageInfo } from "../Volume";
+import VolumeCache, { CacheStore } from "../VolumeCache";
 
 interface PackedChannelsImage {
   name: string;
@@ -94,7 +95,10 @@ class JsonImageInfoLoader implements IVolumeLoader {
   time: number;
   jsonInfo: JsonImageInfo | null = null;
 
-  constructor(urls: string | string[]) {
+  cache?: VolumeCache;
+  cacheStore?: CacheStore;
+
+  constructor(urls: string | string[], cache?: VolumeCache) {
     if (Array.isArray(urls)) {
       this.urls = urls;
     } else {
@@ -102,6 +106,7 @@ class JsonImageInfoLoader implements IVolumeLoader {
     }
 
     this.time = 0;
+    this.cache = cache;
   }
 
   async getJsonImageInfo(loadSpec: LoadSpec): Promise<JsonImageInfo> {
@@ -130,6 +135,10 @@ class JsonImageInfoLoader implements IVolumeLoader {
     this.jsonInfo = await this.getJsonImageInfo(loadSpec);
     const imageInfo = convertImageInfo(this.jsonInfo);
 
+    this.cacheStore = this.cache?.addVolume(imageInfo.numChannels, Math.max(imageInfo.times, this.urls.length), [
+      new Vector3(imageInfo.subregionSize.x, imageInfo.subregionSize.y, imageInfo.subregionSize.z),
+    ]);
+
     const vol = new Volume(imageInfo, loadSpec, this);
     vol.channelLoadCallback = onChannelLoaded;
     vol.imageMetadata = buildDefaultMetadata(imageInfo);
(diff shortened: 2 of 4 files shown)


public/index.ts
```diff
@@ -859,7 +859,7 @@ function onChannelDataArrived(v: Volume, channelIndex: number) {
 
   currentVol.channels[channelIndex].lutGenerator_percentiles(0.5, 0.998);
   view3D.onVolumeData(currentVol, [channelIndex]);
-  view3D.setVolumeChannelEnabled(currentVol, channelIndex, channelIndex < 3);
+  //view3D.setVolumeChannelEnabled(currentVol, channelIndex, channelIndex < 3);
```
toloudis, line 862: removing this line let me turn channels on and off during playback when new data was still arriving.

public/index.ts
```diff
@@ -24,7 +24,7 @@ import { OpenCellLoader } from "../src/loaders/OpenCellLoader";
 import { State, TestDataSpec } from "./types";
 import { getDefaultImageInfo } from "../src/Volume";
 
-const CACHE_MAX_SIZE = 500_000_000;
+const CACHE_MAX_SIZE = 1_000_000_000;
```
toloudis, line 27: MORE MORE MORE.  I just increased this so the whole time series could be cached. Didn't try to find a min value, just doubled it.

kmitcham: Is doubling enough?

public/index.ts
kmitcham: Nit: delete, or add a comment when commenting out a line.

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -187,57 +198,90 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
-          for (let ch = 0; ch < Math.min(thisbatch.length, 4); ++ch) {
-            channelsBits.push(new Uint8Array(w * h));
-          }
-          // extract the data
-          for (let j = 0; j < Math.min(thisbatch.length, 4); ++j) {
-            for (let px = 0; px < w * h; px++) {
-              channelsBits[j][px] = iData.data[px * 4 + j];
+      // if any in batch is cached then they all are...?
+      let cacheHit = false;
+      for (let j = 0; j < Math.min(batch.length, 4); ++j) {
```
kmitcham, line 209: If the first item is a hit, and the second is a miss, cacheHit is true, and no further items are checked?  This boolean flag seems odd inside the loop.  If you only want to see if at least one is a hit, break on if (cacheResult), not else.

kmitcham (approved):

public/index.ts
toloudis: enough to work to demonstrate the functionality with the sample data.  This is a tricky business because I am not aware of easy ways to reliably detect available mem resources in javascript.  We'll work out a lowest common denominator approach for production code.

src/loaders/JsonImageInfoLoader.ts
toloudis: In this particular loader, the data is requested and arrives in batches of up to 4.  The reason why this code works is because it's order-independent because of the batched nature of the data.  Either all items in the batch will be cached or they won't.  So that assumption is built-in here.
Any cache miss means we need to request the whole batch.  I could keep looping, but it's going to fetch a whole batch of data anyway.  Maybe would be easier to read if I didn't break out of the loop.
(I could also rewrite this to check cache for batch[0] only, and then loop over batches based on that result.)

toloudis: We still need the loop in order to fetch the cached items, but yes.

kmitcham: Ah.  I was focused on loop control, and missed that part.  Thanks for explaining!

public/index.ts
ShrimpCryptid: Extra nit: Add a TODO for easier searching later?

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -130,6 +135,10 @@ class JsonImageInfoLoader implements IVolumeLoader {
     this.jsonInfo = await this.getJsonImageInfo(loadSpec);
     const imageInfo = convertImageInfo(this.jsonInfo);
 
+    this.cacheStore = this.cache?.addVolume(imageInfo.numChannels, Math.max(imageInfo.times, this.urls.length), [
+      new Vector3(imageInfo.subregionSize.x, imageInfo.subregionSize.y, imageInfo.subregionSize.z),
+    ]);
+
```
ShrimpCryptid, line 141: If `createVolume` is called multiple times, will this add duplicates of the same volume to `cache`?

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -187,57 +198,90 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
-          const channelsBits: Uint8Array[] = [];
-          // allocate channels in batch
-          for (let ch = 0; ch < Math.min(thisbatch.length, 4); ++ch) {
-            channelsBits.push(new Uint8Array(w * h));
-          }
-          // extract the data
-          for (let j = 0; j < Math.min(thisbatch.length, 4); ++j) {
-            for (let px = 0; px < w * h; px++) {
-              channelsBits[j][px] = iData.data[px * 4 + j];
+      // if any in batch is cached then they all are...?
```
ShrimpCryptid, line 207: What's the uncertainty here? Is it possible for only part of a batch to be cached?

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -187,57 +198,90 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
+          volume.setChannelDataFromVolume(chindex, cacheResult);
+          onChannelLoaded?.(volume, chindex);
+          cacheHit = true;
+          console.log("cache hit " + volume.loadSpec.time + " ch " + chindex);
+        } else {
+          console.log("cache miss " + volume.loadSpec.time + " ch " + chindex);
+          break;
+        }
+      }
+      if (!cacheHit) {
```
ShrimpCryptid, line 222: Nit: I would favor reducing indentation by flipping this if statement!
```
if (cacheHit) {
continue;
}
...
```

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -187,57 +198,90 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
+                scale: 0,
+                time: volume.loadSpec.time,
+                channel: thisbatch[ch],
+              };
+              cacheStore && cache?.insert(cacheStore, volume.channels[thisbatch[ch]].volumeData, cacheInsertDims);
+
+              onChannelLoaded?.(volume, thisbatch[ch]);
+            }
+          };
+        })(batch);
```
ShrimpCryptid, line 280: Nit: The calling of this function happens so far down from the original `img.onload` setter... Would it make more sense to declare the function separately and then call it?

ShrimpCryptid (approved):
Looks good, left some nits about code styling and a few questions.

src/loaders/JsonImageInfoLoader.ts
toloudis: Hm, that might be out of scope of this PR but it's something for us to worry about for sure.

src/loaders/JsonImageInfoLoader.ts
toloudis: I think I might have a go at rewriting this after Kevin's comments about it.  It could be a bit simpler.  The question mark is just me knowing that I am baking in a very specific assumption and the logic only works because of it.

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -187,6 +198,35 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
+          cacheHit = false;
+          // we can stop checking because we know we are going to have to fetch the whole batch
+          break;
+        }
+      }
+
+      // if there were ANY cache misses then we need to request the batch
+      if (cacheHit) {
+        continue;
+      }
```
frasercl, line 228: nit: might flip the phrasing on this comment to match the flipped logic of this `if`, something like "if all channels are in cache, move on to the next image without requesting"

frasercl (approved):
Looks good!

frasercl (approved):
Looks good!

PR #144, opened 2023-09-28
allow tiffs to load from server that doesn't support range requests

toloudis:
to support Amil's use case, he is running from a http server that doesn't support range requests.  This is a stopgap to just allow the whole tiff file to be downloaded in absence of range reqs.

### DIFF
--- a/src/loaders/TiffLoader.ts
+++ b/src/loaders/TiffLoader.ts
@@ -61,7 +61,7 @@ function getOMEDims(imageEl: Element): OMEDims {
 }
 
 async function getDimsFromUrl(url: string): Promise<OMEDims> {
-  const tiff = await fromUrl(url);
+  const tiff = await fromUrl(url, { allowFullFile: true });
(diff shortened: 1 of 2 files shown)


ShrimpCryptid (approved):

toloudis:
The reason why this is ok is :
-  large tiff files are already not an efficient input format for this viewer
- small tiffs are still reasonable to load
- cloud hosts generally support range requests already
- if we want to optimize for larger tiff we can always undo this change easily as a part of that work, and reinstate this requirement (as a possible breaking change for @amilworks use case)

frasercl (approved):

amilworks:
Just to comment on a future development on my side, I most likely will be moving to support `byte-ranges`, but it will not be immediate, maybe in a few months. Most of the `TIFF` files I am serving are in the kilobyte range, hence the reason why loading the entire image is a nice feature for small file sizes. But yeah, eventually we will move the backend to `StreamResponse`. Thanks again for this and all your help!

PR #145, opened 2023-09-28
Fix zarr loading from cache

frasercl:
Fixes a bug where `Volume`s loaded from `OmeZarrLoader` would appear blank when the data was fetched from cache.

This problem was caused because the volume's `imageInfo` was not being updated until after the data had been loaded. Consider the following pseudocode:

```typescript
// ...
loadSomeData((data) => { // callback
  setData(data);
});
setImageInfo(imageInfo);
```

If `loadSomeData` is asynchronous, like it is when the data has to come from a network source, it yields control and allows `setImageInfo` to run before `setData`. BUT if it is synchronous, as is the case when data comes straight from cache, then `setData` is called immediately, before `setImageInfo` gets a chance to run.

The solution is just to move `setImageInfo` up to before `loadSomeData`. So that's what this fix does. No code behavior is actually modified besides being shuffled around.


toloudis (approved):
LGTM!

ShrimpCryptid (approved):
Thank you for the explanation! Looks good.

PR #146, opened 2023-09-28
Rework `LoadSpec`

### LINKED ISSUE #134: Rework `LoadSpec`

frasercl:
Fully resolves #134: reworks `LoadSpec` to have all the fields we want and none of the fields we don't.

- Removed field: `subpath`. No one wanted to use it.
- Added optional field: `multiscaleLevel`. This specifies which scale level to load (in zarr only). The loader will still pick the optimal level if `multiscaleLevel` is undefined or larger than the texture limits, so this is more like a "minimum downsampling level."
- Added optional field: `channels`. This lets you pick a subset of channel indexes to load.

Also, updated loaders to always set `Volume.loadSpec` to reflect the data that was *actually* loaded, rather than the `LoadSpec` that was passed in. E.g. no loader other than zarr can use `subregion`, so keep it set to the max. And since json loader has to load channels in static groups of (up to) four, when the user requests some subset of channels, also include all the channels that came along with the requested channel's images.

### DIFF
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -130,6 +130,8 @@ export default class Volume {
   public imageInfo: ImageInfo;
   public loadSpec: LoadSpec;
   public loader?: IVolumeLoader;
+  // `LoadSpec` representing the minimum data required to display what's in the viewer (subregion, channels, etc.).
+  // Used to intelligently issue load requests whenever required by a state change. Modify with `updateRequiredData`.
   private loadSpecRequired: Required<LoadSpec>;
   public channelLoadCallback?: PerChannelCallback;
   public imageMetadata: Record<string, unknown>;
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -130,7 +130,7 @@ export default class Volume {
   public imageInfo: ImageInfo;
   public loadSpec: LoadSpec;
   public loader?: IVolumeLoader;
-  private loadSpecRequired: LoadSpec;
+  private loadSpecRequired: Required<LoadSpec>;
   public channelLoadCallback?: PerChannelCallback;
   public imageMetadata: Record<string, unknown>;
   public name: string;
@@ -161,7 +161,13 @@ export default class Volume {
     this.imageInfo = imageInfo;
     this.name = this.imageInfo.name;
     this.loadSpec = loadSpec;
-    this.loadSpecRequired = { ...loadSpec, subregion: loadSpec.subregion.clone() };
+    this.loadSpecRequired = {
+      // Fill in defaults for optional properties
+      multiscaleLevel: 0,
+      channels: Array.from({ length: this.imageInfo.numChannels }, (_val, idx) => idx),
+      ...loadSpec,
+      subregion: loadSpec.subregion.clone(),
+    };
     this.loader = loader;
     // imageMetadata to be filled in by Volume Loaders
     this.imageMetadata = {};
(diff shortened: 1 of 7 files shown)


ShrimpCryptid (commented):
Could you talk a bit more about why subpath was removed?

src/Volume.ts
```diff
@@ -130,7 +130,7 @@ export default class Volume {
   public imageInfo: ImageInfo;
   public loadSpec: LoadSpec;
   public loader?: IVolumeLoader;
-  private loadSpecRequired: LoadSpec;
+  private loadSpecRequired: Required<LoadSpec>;
```
toloudis, line 133: What do you think of a comment up here saying why loadSpecRequired is different than loadSpec?  Just to show the intention...

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -231,60 +249,62 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
-          for (let ch = 0; ch < Math.min(thisbatch.length, 4); ++ch) {
-            channelsBits.push(new Uint8Array(w * h));
-          }
-          // extract the data
-          for (let j = 0; j < Math.min(thisbatch.length, 4); ++j) {
-            for (let px = 0; px < w * h; px++) {
-              channelsBits[j][px] = iData.data[px * 4 + j];
-            }
+
+      img.onload = (event) => {
```
toloudis, line 257: This is a big improvement!

src/loaders/OpenCellLoader.ts
```diff
@@ -48,13 +48,14 @@ class OpenCellLoader implements IVolumeLoader {
     };
 
     // got some data, now let's construct the volume.
-    const vol = new Volume(imgdata, loadSpec, this);
+    // This loader uses no fields from `LoadSpec`. Iinitialize volume with defaults.
```
toloudis, line 51:
```suggestion
    // This loader uses no fields from `LoadSpec`. Initialize volume with defaults.
```

toloudis (approved):
Couple small comments but the functional stuff looks good to me!

src/loaders/JsonImageInfoLoader.ts
frasercl: I (mercifully) never had a reason to use that make-a-function-that-returns-a-function-then-immediately-call-the-outer-function hack in the era before arrow functions, and I don't envy folks who had to write software professionally back when that was an accepted solution.

frasercl:
> Could you talk a bit more about why subpath was removed?

As I understand it, the only thing `subpath` might have been used for is what `multiscaleLevel` does now: picking different multiscale levels, by name rather than index (if they were different - I think that was rare anyways). It was never really used anywhere and didn't have a clear reason to keep existing. @toloudis might know more though.

toloudis:
> > Could you talk a bit more about why subpath was removed?
> 
> As I understand it, the only thing `subpath` might have been used for is what `multiscaleLevel` does now: picking different multiscale levels, by name rather than index (if they were different - I think that was rare anyways). It was never really used anywhere and didn't have a clear reason to keep existing. @toloudis might know more though.

Yes - in zarr metadata, a multiscale level is really pointing to a folder that contains zarr arrays. And in practice, often the folder name is the same as the integer index.   So subpath (aka multiscale level as string) was an early way to make it work but it's only relevant for Zarr.  An integer level index is much more sensible and general - and in the Zarr case, it is used to find the subpath to the zarr data for that level.

ShrimpCryptid (approved):
Thanks for clarifying, looks good!

PR #149, opened 2023-10-11
Feature/Zarr loader caches chunks

frasercl:
Changes the relationship between `OMEZarrLoader` and the cache: rather than caching whatever arbitrary subsets of the volume are requested, the loader caches the actual array chunks loaded by the underlying zarr library. This has some significant benefits:

- **All data loaded from the server is cached, not just the data placed in the volume.** For example, the chunks in our test volumes commonly cross four z-slices. Currently in z-slice mode, we cache only the slice that is displayed, and have to re-request data when the user changes slices. With this change, we get those three extra slices in the cache for free.
- **The cache may provide the same data for multiple requests with different regions.** Even if all the data required for a request has been previously loaded, if that request's region doesn't exactly match a previously cached request's region the current behavior is to reload the requested data and re-cache it with a different region. With this change, the same data can be applied to multiple distinct but intersecting regions, and the loader may even source data from a combination of cache and remote requests.

I did this by writing a new `Store`, which is zarr.js's tiny abstraction around anything that can read and write data to a `string` path (the user's local storage, a server, etc.). `ChunkCachingStore` wraps any other store and adds cache queries and inserts. This approach is great in that it allows us to call regular zarr load methods without having to worry about the cache at all, but it also has some important implications:
- Because `Store`s are so low-level, data is now cached before decompression and before possible conversion of the buffer to `Uint8Array`. The performance overhead of doing these extra steps on cache hits seems to be minimal, though not entirely negligible. At least we save the user some memory though.
- I had to make some changes to how `VolumeCache` works. Most notably, because it's impractical and useless to convert from chunk coordinates to pixel subregions all the time, the cache is now indexed by chunks. The cache also now works with raw `ArrayBuffer`s (because entries haven't necessarily been type-converted) and does not sanity check buffer length against region extent (because entries may be compressed and don't have a consistent size).
- With accesses to a single shared cache now integrated into all our interactions with the zarr library, it becomes impractical to hand off each channel load to its own web worker. (It's certainly not impossible though, if we want to keep it.) All the testing I did indicates the performance difference between the old and new loaders is basically a toss-up, but this is still a bummer. (It seemed like the old loader was delayed by a long compile step for all its workers? If we put workers back, it might be worth the effort to do it with a sort of thread pool approach utilizing a fixed number of persistent workers.)

One more thing: the loader now does its various metadata fetches on construction, rather than with every call to a loading method. This ought to save some unnecessary requests, but it did require replacing the constructor with an async static method.


src/loaders/OmeZarrLoader.ts
```diff
@@ -81,89 +83,125 @@ type OMEZarrMetadata = {
...
-      console.log("ERROR: UNRECOGNIZED AXIS in zarr: " + axis.name);
-    }
+  setItem(item: string, value: ArrayBuffer): Promise<boolean> {
+    return this.baseStore.setItem(item, value);
   }
-  if (axisTCZYX[2] === -1 || axisTCZYX[3] === -1 || axisTCZYX[4] === -1) {
-    console.log("ERROR: zarr loader expects a z, y, and x axis.");
+
+  deleteItem(item: string): Promise<boolean> {
+    return this.baseStore.deleteItem(item);
```
toloudis, line 173: This looks scary.  Consider never deleting from baseStore and console log instead?

src/loaders/OmeZarrLoader.ts
```diff
@@ -81,89 +83,125 @@ type OMEZarrMetadata = {
...
-      axisTCZYX[2] = i;
-    } else if (axis.name === "y") {
-      axisTCZYX[3] = i;
-    } else if (axis.name === "x") {
-      axisTCZYX[4] = i;
-    } else {
-      console.log("ERROR: UNRECOGNIZED AXIS in zarr: " + axis.name);
-    }
+  setItem(item: string, value: ArrayBuffer): Promise<boolean> {
+    return this.baseStore.setItem(item, value);
```
toloudis, line 169: this looks scary - consider never overwriting anything in the baseStore from our code?

public/index.ts
```diff
@@ -991,12 +993,12 @@ function createTestVolume() {
...
 }
 
-function createLoader(data: TestDataSpec): IVolumeLoader {
+async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
   switch (data.type) {
     case "opencell":
       return new OpenCellLoader();
     case "omezarr":
-      return new OMEZarrLoader(data.url, volumeCache);
+      return await OMEZarrLoader.createLoader(data.url, 0, volumeCache);
```
toloudis, line 1001: Thinking about how to keep this a bit more polymorphic across all loaders, there are a couple of things you could do.
1. make a factory method in VolumeLoaderUtils that figures out file type from url and creates the loader. (more or less replicating this function) - and then always create loaders through that function, even in w3cv
2. make all loaders constructors private and make them all have a static createLoader.
3. let the constructors stay trivial and add an async init call to be called right after construction
Some mix and match of those would keep this code more uniform in appearance.  I slightly favor #1 if it works in w3cv and here.  You might have to pass in more info than url in order for it to detect properly.

src/loaders/OmeZarrLoader.ts
```diff
@@ -325,102 +425,63 @@ class OMEZarrLoader implements IVolumeLoader {
...
+        }
+      });
+
+      const result = (await level.getRaw(sliceSpec)) as RawArray;
+      const u8 = convertChannel(result.data, result.dtype);
+      vol.setChannelDataFromVolume(ch, u8);
+      onChannelLoaded?.(vol, ch);
+    });
+
+    await Promise.all(channelPromises);
```
toloudis, line 484: what would happen if you didn't await here?

frasercl: This is just an add-on to make the function return a promise that resolves when all channels are loaded - I don't think this method is ever actually `await`ed. (Come to think of it, this does give us the opportunity to transition this method from callback-based to promise-based with minimal changes, if we want.) As written, I don't believe it'll block anything.

src/loaders/OmeZarrLoader.ts
frasercl: I believe this and your below comment is prevented from happening by the `mode: "r"` (read-only) argument passed into `openArray`/`openGroup` and the fact that we never call any functions that begin with "set," but I also felt safe doing this because `deleteItem` is unimplemented by `HTTPStore`. Worth at least adding warnings though.

public/index.ts
frasercl: I'll give 1 a try, it's also my favorite

src/VolumeCache.ts
```diff
@@ -19,9 +19,9 @@ type MaybeCacheEntry = CacheEntry | null;
 type CacheEntry = {
   /** The data contained in this entry */
   // TODO allow more types of `TypedArray` to be stored together in the cache?
-  data: Uint8Array;
-  /** The subset of the volume covered by this entry, in pixels */
-  subregion: Box3;
+  data: ArrayBuffer;
+  /** The chunk of the volume covered by this entry, in pixels */
```
ShrimpCryptid, line 23: Does the comment need to be updated since it's not using pixel indexing anymore?

frasercl: Whoops, yep!

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -136,7 +136,7 @@ class JsonImageInfoLoader implements IVolumeLoader {
     const imageInfo = convertImageInfo(this.jsonInfo);
 
     this.cacheStore = this.cache?.addVolume(imageInfo.numChannels, Math.max(imageInfo.times, this.urls.length), [
-      new Vector3(imageInfo.subregionSize.x, imageInfo.subregionSize.y, imageInfo.subregionSize.z),
+      new Vector3(1, 1, 1),
```
ShrimpCryptid, line 139: Why does this get converted to 1,1,1 here? Is it because chunks don't need to be scaled, or...?

src/loaders/OmeZarrLoader.ts
```diff
@@ -81,89 +83,125 @@ type OMEZarrMetadata = {
...
-    console.log("ERROR: no coordinate transformations for scale level");
-    return [1, 1, 1, 1, 1];
-  }
+/** Turns `axisTCZYX` into the number of dimensions in the array */
+const getDimensionCount = ([t, c]: [number, number, number, number, number]) => 3 + Number(t > -1) + Number(c > -1);
+
+class ChunkCachingStore implements AsyncStore<ArrayBuffer, RequestInit> {
+  // Required by `AsyncStore`
+  listDir?: (path?: string) => Promise<string[]>;
+  rmDir?: (path?: string) => Promise<boolean>;
```
ShrimpCryptid, line 92: What does rmDir stand for here? Remove directory?

src/loaders/OmeZarrLoader.ts
```diff
@@ -178,139 +216,201 @@ function pickLevelToLoad(loadSpec: LoadSpec, multiscaleDims: number[][], dimInde
   }
 }
 
+function convertChannel(channelData: TypedArray, dtype: string): Uint8Array {
+  if (dtype === "|u1") {
```
ShrimpCryptid, line 220: Should this be a constant somewhere?

src/loaders/OmeZarrLoader.ts
```diff
@@ -178,139 +216,201 @@ function pickLevelToLoad(loadSpec: LoadSpec, multiscaleDims: number[][], dimInde
...
 
-    const axisTCZYX = remapAxesToTCZYX(axes);
+    return new OMEZarrLoader(scaleLevels, multiscale, metadata.omero, axisTCZYX);
+  }
 
-    // Assume all axes have the same units - we have no means of storing per-axis unit symbols
-    const spaceUnitName = axes[axisTCZYX[4]].unit;
+  private getUnitSymbols(): [string, string] {
+    // Assume all spatial axes have the same units - we have no means of storing per-axis unit symbols
+    const xi = this.axesTCZYX[4];
```
ShrimpCryptid, line 300: Okay this is maybe a silly suggestion by my brain keeps thinking it would be really nice to have an enum for TCZYX in case this ever changes

src/test/VolumeCache.test.ts
```diff
@@ -28,14 +28,14 @@ describe("VolumeCache", () => {
...
-        region: new Box3(new Vector3(0, 0, 1), new Vector3(2, 2, 2)),
+        chunk: new Vector3(0, 0, 1),
       };
       const insertionResult = cache.insert(vol, new Uint8Array(4), extent);
       expect(insertionResult).to.be.true;
       expect(cache.get(vol, 1, extent)).to.deep.equal(new Uint8Array(4));
     });
 
-    it("defaults to the first channel, first timestep, and maximum extent when unspecified", () => {
+    it("defaults to the first channel, first timestep, and first chunk when unspecified", () => {
```
ShrimpCryptid, line 38: Does zarr use 1-based indexing? Is there a chunk w/ coordinates 0,0,0?

ShrimpCryptid (approved):
I looked through the PR and left a few comments. I think I'm still struggling a bit to understand how all the pieces fit together, which may just be me having an off day today, but I don't want to halt any dev work! Thanks for showing me the working demo too :)

src/loaders/index.ts
```diff
@@ -0,0 +1,57 @@
...
+type CreateLoaderOptions = CreateZarrLoaderOptions | CreateJsonLoaderOptions | CreateTiffLoaderOptions;
+
+async function createVolumeLoader<T extends VolumeFileFormat>(
+  fileType: T,
+  options?: CreateLoaderOptionsMap[T]
+): Promise<IVolumeLoader> {
+  switch (fileType) {
+    case VolumeFileFormat.ZARR:
+      const zOptions = options as CreateZarrLoaderOptions;
+      return await OMEZarrLoader.createLoader(zOptions.path, zOptions.scene, zOptions.cache);
```
toloudis, line 35: What happens if loader creation fails?  If one of the constructors throws, do we get a rejected promise?

src/loaders/index.ts
```diff
@@ -0,0 +1,57 @@
...
+export {
+  LoadSpec,
+  loadSpecToString,
+  OMEZarrLoader,
+  JsonImageInfoLoader,
+  TiffLoader,
+  VolumeDims,
+  VolumeFileFormat,
+  CreateLoaderOptions,
+  createVolumeLoader,
```
toloudis, line 56: Nit: These types of internal index.ts files always feel confusing to me because they obscure where the imports are really coming from.  The top-level index.ts for volume-viewer could be exporting these from their sources instead.

src/loaders/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type CreateLoaderOptions = CreateZarrLoaderOptions | CreateJsonLoaderOptions | CreateTiffLoaderOptions;
+
+async function createVolumeLoader<T extends VolumeFileFormat>(
+  fileType: T,
+  options?: CreateLoaderOptionsMap[T]
+): Promise<IVolumeLoader> {
+  switch (fileType) {
+    case VolumeFileFormat.ZARR:
+      const zOptions = options as CreateZarrLoaderOptions;
```
toloudis, line 34: "unexpected lexical declaration in case block" might need curly brackets around the block

src/loaders/index.ts
frasercl: Yes - `async` functions transform runtime errors into rejected promises, and `await`ing a rejected promise throws a runtime error.

src/loaders/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const enum VolumeFileFormat {
+  ZARR = "zarr",
+  JSON = "json",
+  TIFF = "tiff",
+}
+
+type CreateZarrLoaderOptions = { path: string; cache?: VolumeCache; scene?: number };
+type CreateJsonLoaderOptions = { path: string | string[]; cache?: VolumeCache };
+type CreateTiffLoaderOptions = { path: string };
```
toloudis, line 16: For simplicity, why not just make one generic union type that has path, cache, and scene? (Do we even use scene? I think not, for zarr?).   Then you can just pass LoaderOptions right in to the factory function

src/loaders/index.ts
toloudis: so the place where we `await` this function's results need to be ready to catch things (thinking about w3cv)

src/loaders/index.ts
frasercl: Scene is added as an option to `OMEZarrLoader` by this PR. My thought was, in a loaders-as-handles world it makes more sense to pick a scene at construction time rather than allow scenes with potentially different dimensions to be selected and reselected at load time.

toloudis: I agree that scene should be selected at load time.  But I'm semi-proposing that we remove scene for now, as it is a higher-level detail on how zarr data is organized, and we really don't have a good "multi-scene" example to work with, if I remember correctly.  At some level a scene is just a whole zarr file.

src/loaders/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const forceString = (value: string | string[]): string => (typeof value === "object" ? value[0] : value);
+
+async function createVolumeLoader(fileType: VolumeFileFormat, options: CreateLoaderOptions): Promise<IVolumeLoader> {
+  switch (fileType) {
+    case VolumeFileFormat.ZARR:
+      return await OMEZarrLoader.createLoader(forceString(options.path), options.scene, options.cache);
+    case VolumeFileFormat.JSON:
+      return new JsonImageInfoLoader(options.path, options.cache);
+    case VolumeFileFormat.TIFF:
```
toloudis, line 28: I have a vague suspicion that w3cv is going to have a kind of switch statement that guesses the VolumeFileFormat from the file extension, which might feel redundant with this code.   Not sure what would be cleanest, but maybe the type could be used as a hint and if omitted, then we make a guess by looking at the end of the string, or if it's an array.  Then fileType could become an optional member of CreateLoaderOptions, even.

toloudis (approved):
Not sure why there's a broken unit test but once fixed, this LGTM.    I recommend one minor tweak to the loader factory, but it's not a showstopper.

PR #153, opened 2023-10-25
Fix/persistent channel enabled state

toloudis:

- [ ] Name your pull request _your-development-type/short-description_. Ex: _feature/read-tiff-files_
- [ ] Link to any relevant issue in the PR description. Ex: _Resolves [gh-##], adds tiff file format support_
- [ ] Provide description and context of changes.
- [ ] Provide relevant tests for your feature or bug fix.
- [ ] Provide or update documentation for any feature added by your pull request.

Thanks for contributing!

### DIFF
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -163,9 +163,14 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
       this.setUniform("iResolution", this.settings.resolution);
     }
 
-    if (dirtyFlags & SettingsFlags.MASK) {
+    if (dirtyFlags & SettingsFlags.MASK_ALPHA) {
       this.setUniform("maskAlpha", this.settings.maskAlpha);
-      this.setUniform("maskAlpha", this.settings.maskAlpha);
+    }
+    if (dirtyFlags & SettingsFlags.MASK_DATA) {
+      this.channelData.setChannelAsMask(
+        this.settings.maskChannelIndex,
+        this.volume.getChannel(this.settings.maskChannelIndex)
+      );
     }
   }
 
@@ -211,6 +216,7 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
 
     this.channelData.gpuFuse(canvas.renderer);
     this.setUniform("textureAtlas", this.channelData.getFusedTexture());
+    this.setUniform("textureAtlasMask", this.channelData.maskTexture);
 
     this.geometryTransformNode.updateMatrixWorld(true);
 
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -793,7 +793,8 @@ export class View3d {
    * @param {boolean} enabled
    */
   setVolumeChannelEnabled(volume: Volume, channel: number, enabled: boolean): void {
-    this.image?.setVolumeChannelEnabled(channel, enabled);
+    // TODO: should this become the primary way channel settings are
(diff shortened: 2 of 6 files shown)


src/loaders/OmeZarrLoader.ts
```diff
@@ -234,16 +234,21 @@ function convertChannel(channelData: TypedArray, dtype: string): Uint8Array {
...
-  });
+  for (let i = 0; i < channelData.length; i++) {
+    const val = channelData[i];
+    if (val < min) {
+      min = val;
+    }
+    if (val > max) {
+      max = val;
+    }
+  }
```
ShrimpCryptid, line 245: forEach also lets you pass the index into the callback; any reason to prefer a `for` loop over it?

ShrimpCryptid: Answered my own question, `for` loops are more efficient, but there's some debate over code clarity

src/Atlas2DSlice.ts
```diff
@@ -163,9 +163,14 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
       this.setUniform("iResolution", this.settings.resolution);
     }
 
-    if (dirtyFlags & SettingsFlags.MASK) {
-      this.setUniform("maskAlpha", this.settings.maskAlpha);
-      this.setUniform("maskAlpha", this.settings.maskAlpha);
```
ShrimpCryptid, line 168: oops this was definitely a bug before, nice catch

src/View3d.ts
```diff
@@ -793,7 +793,11 @@ export class View3d {
    * @param {boolean} enabled
    */
   setVolumeChannelEnabled(volume: Volume, channel: number, enabled: boolean): void {
-    this.image?.setVolumeChannelEnabled(channel, enabled);
+    // NOTE: explicitly calling `image?.setVolumeChannelEnabled` allowed the enabled/disabled state of the channel to
+    //   be inconsistent across loads (incl. on time or z-slice changes) because channel state is restored on reload
+    //   from the object managed by `setChannelOptions`.
```
ShrimpCryptid, line 798: Does this note still make sense to leave here now that the offending line is removed? Does it make sense to leave it in but commented-out?

ShrimpCryptid (approved):
Left a few comments but looks good!

src/loaders/OmeZarrLoader.ts
frasercl: We found yesterday that using `forEach` was adding ~40ms to data loads from the cache (it would probably add more for volumes with more channels). Loads already (I think) feel a bit hitchy when trying to move quickly through a time series. I believe I wrote it this way from an earlier version that actually used `for` loops, so this is my mistake being corrected :)

src/View3d.ts
frasercl: I wrote this more as an explanatory comment on the TODO below, so that there's a history that motivates why we ought to examine how state is managed within channels. So I suppose it might be more at home in a GitHub issue?

frasercl (approved):
Might be wise to move the comment Peyton flagged into an issue before merging (if we agree that's an issue worth tracking), but all the code here looks solid!

src/loaders/OmeZarrLoader.ts
toloudis: Yeah for looping over large numeric data arrays, there is a tangible performance gain to doing it this way.

Issue #154, opened 2023-10-25
Allow new loads to pick up on old requests in progress


frasercl:
## Use Case

With `RequestQueue` added to the zarr load path, if `OMEZarrLoader.loadVolumeData` is called and then called again before chunk requests generated by the first call can complete, we should be able to cancel the requests made by the previous call. However, it is possible that the new loads want the same chunks! We should be able to transfer ownership of in-progress requests to the new load, rather than send lots of different requests for the same data.

## Solution

Enable either `RequestQueue` or a new wrapper class to pick cancelled requests back up if they were already in progress. Make promises returned by this queue unique to a single request *and* a single subscriber, and reject only the promise associated with a single subscriber without trying to halt a request in progress. Use this new behavior to implement the behavior described above.

Issue #143, opened 2023-09-28
tweakpane leaks memory


toloudis:
## Description

Run a time series playthrough in the test viewer.  In Chrome devtools memory profiler you can see heap allocations every time new volume data is swapped in.  The retained objects are nearly all from tweakpane.

## Expected Behavior

Playing through time series should not cause the js heap graph to rise.
 
Possible workaround: do we even need to rebuild the tweakpane every time data arrives?  Best solution would be to be able to properly clean up the tweakpane, or understand why we can not.

PR #161, opened 2023-10-30
Use request queue

### LINKED ISSUE #151: Integrate `RequestQueue`
## Use Case

We have a `RequestQueue` object for limiting concurrent requests and allowing them to be cancelled. It needs integrating into our loading path before we can start using the capabilities it gives us.

## Solution

Integrate `RequestQueue` into `OmeZarrLoader` (and other loaders?). Double-check that zarr's under-the-hood request queue (p-queue) isn't also trying to do the same thing.

There should be no noticeable difference in behavior after this change.

frasercl:
Resolves #151: `OMEZarrLoader` uses `RequestQueue` in the most basic possible way, while taking advantage of two new abilities it provides: cancelling requests and limiting concurrency.

Also does a bit of cleanup and fixing to `RequestQueue` itself and to the `createVolumeLoader` convenience function.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -18,6 +18,7 @@ import {
 } from "./VolumeLoaderUtils";
 
 const MAX_ATLAS_DIMENSION = 2048;
+const CHUNK_REQUEST_CANCEL_REASON = "chunk request cancelled";
 
 type CoordinateTransformation =
   | {
@@ -474,7 +475,7 @@ class OMEZarrLoader implements IVolumeLoader {
 
   async loadVolumeData(vol: Volume, explicitLoadSpec?: LoadSpec, onChannelLoaded?: PerChannelCallback): Promise<void> {
     // First, cancel any pending requests for this volume
-    this.requestQueue.cancelAllRequests();
+    this.requestQueue.cancelAllRequests(CHUNK_REQUEST_CANCEL_REASON);
 
     vol.loadSpec = explicitLoadSpec ?? vol.loadSpec;
     const maxExtent = this.maxExtent ?? new Box3(new Vector3(0, 0, 0), new Vector3(1, 1, 1));
@@ -519,10 +520,17 @@ class OMEZarrLoader implements IVolumeLoader {
         }
       });
 
-      const result = (await level.getRaw(sliceSpec, { concurrencyLimit: Infinity })) as RawArray;
-      const u8 = convertChannel(result.data, result.dtype);
-      vol.setChannelDataFromVolume(ch, u8);
-      onChannelLoaded?.(vol, ch);
+      try {
+        const result = (await level.getRaw(sliceSpec, { concurrencyLimit: Infinity })) as RawArray;
+        const u8 = convertChannel(result.data, result.dtype);
+        vol.setChannelDataFromVolume(ch, u8);
+        onChannelLoaded?.(vol, ch);
+      } catch (e) {
+        // TODO: verify that cancelling requests in progress doesn't leak memory
+        if (e !== CHUNK_REQUEST_CANCEL_REASON) {
+          throw e;
+        }
+      }
     });
 
     await Promise.all(channelPromises);
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -5,7 +5,8 @@ import { AsyncStore } from "zarr/types/storage/types";
 import { Slice } from "zarr/types/core/types";
 
 import Volume, { ImageInfo } from "../Volume";
-import VolumeCache, { CacheStore } from "../VolumeCache";
+import VolumeCache, { CacheStore, DataArrayExtent } from "../VolumeCache";
+import RequestQueue from "../utils/RequestQueue";
 import { IVolumeLoader, LoadSpec, PerChannelCallback, VolumeDims } from "./IVolumeLoader";
 import {
   buildDefaultMetadata,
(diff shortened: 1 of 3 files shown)


src/loaders/index.ts
```diff
@@ -15,33 +15,29 @@ export type CreateLoaderOptions = {
...
-        throw new Error(`Unknown file type: ${options.fileType}`);
-    }
-  } else {
-    const pathString = forceString(path);
-    if (pathString.endsWith(".json")) {
+  const pathString = typeof path === "object" ? path[0] : path;
+
+  switch (options?.fileType) {
+    case VolumeFileFormat.ZARR:
+      return await OMEZarrLoader.createLoader(pathString, options.scene, options.cache);
```
toloudis, line 29: should use concurrencylimit here?

src/loaders/OmeZarrLoader.ts
```diff
@@ -440,6 +474,9 @@ class OMEZarrLoader implements IVolumeLoader {
   }
 
   async loadVolumeData(vol: Volume, explicitLoadSpec?: LoadSpec, onChannelLoaded?: PerChannelCallback): Promise<void> {
+    // First, cancel any pending requests for this volume
+    this.requestQueue.cancelAllRequests(CHUNK_REQUEST_CANCEL_REASON);
```
toloudis, line 478: yay!

src/loaders/OmeZarrLoader.ts
```diff
@@ -254,6 +276,10 @@ class OMEZarrLoader implements IVolumeLoader {
   omeroMetadata: OmeroTransitionalMetadata;
   axesTCZYX: [number, number, number, number, number];
 
+  // TODO: should we be able to share `RequestQueue`s between loaders?
+  // TODO: store sets of requests per volume via some unique key
+  requestQueue: RequestQueue;
```
toloudis, line 281: Similar to the cache, the RequestQueue can be something of a global to the whole viewer, and created once at init time but used for any and all volumes and volumeloader instances.  I would think that the concurrencylimit should be global.
The string keys could be used with wildcards to cancel things for specific volumes (if keyed with a unique volume prefix, for example)
`cancelByPrefix(prefix: string) { for all keys if key startswith prefix then cancel } `

toloudis (approved):
LGTM, one minor cleanup thing with the loader factory.

src/loaders/OmeZarrLoader.ts
toloudis: Does this actually work?  Can you easily test this where you call loadVolumeData and it cancels previous ones without actually trying to render the stale data?

src/loaders/index.ts
frasercl: yup - whoops!

src/loaders/OmeZarrLoader.ts
```diff
@@ -483,10 +520,17 @@ class OMEZarrLoader implements IVolumeLoader {
...
+        const result = (await level.getRaw(sliceSpec, { concurrencyLimit: Infinity })) as RawArray;
+        const u8 = convertChannel(result.data, result.dtype);
+        vol.setChannelDataFromVolume(ch, u8);
+        onChannelLoaded?.(vol, ch);
+      } catch (e) {
+        // TODO: verify that cancelling requests in progress doesn't leak memory
+        if (e !== CHUNK_REQUEST_CANCEL_REASON) {
+          throw e;
+        }
+      }
```
ShrimpCryptid, line 533: When/how will you verify this?

ShrimpCryptid (approved):
Looks good! Thank you for the fixes in RequestQueue, excited to see this getting integrated! Agree with Dan's comment about moving RequestQueue to be global though.

src/loaders/OmeZarrLoader.ts
frasercl: My plan is to verify this as part of work on #154, which will add the new layer of per-subscriber `Promise` wrapping we talked about earlier. The issue here has to do with whether or not cancelling requests means all associated promises are either rejected (easy to verify) or lose all references and are garbage collected (hard to verify). I believe we can get close to certain on this with the devtools profiler, though it might be helpful to remedy our other known memory leak (#143/#156) first.

src/loaders/OmeZarrLoader.ts
frasercl: An issue I'm anticipating/thinking about here is that we don't have a great way of knowing at the `SmartStoreWrapper` level which requests are associated with which load calls. All we can get out of the `Store` is chunk paths, and all we have to roughly associate them with a single `getRaw` call is the time at which they were requested. So anything we do in the future that relies on grouping chunk requests into coherent batches will run into at least some issues. Prefetching may be particularly tricky, since we have to determine based on which chunks are requested which additional chunks to load into the cache. What happens e.g. when we try to load into two volumes at the same time? How do we know when all chunk loads associated with a single `loadVolumeData` call have been queued and we can start confidently determining which chunks to prefetch?

All this makes me wonder how hard it would be to extract and modify some minimal subset of zarr.js ourselves...

toloudis: Does it seem like just making the request keys have extra information would solve this issue you're talking about?  It still sounds like something that can be dealt with by special string prefixes (or suffixes) on the request keys, and those prefixes are known by the smartstore (or the loader).  Example: instead of "path/to/chunk", use "url?path/to/chunk".  The higher level code can know how to decode that but the requestqueue doesn't care. At the very least this solves the "requested the same chunks but for separate volumes" problem.

PR #162, opened 2023-10-31
Simplify Cache

### LINKED ISSUE #152: Make `VolumeCache` more generic
## Use Case

`VolumeCache` was originally written to store arrays representing arbitrary volume subsets, but is now used to represent a fixed set of zarr chunks. It no longer needs to know the dimensions of the volume it contains, or even that its entries represent volume subsets at all.

## Solution

Design some changes to make `VolumeCache` more generic, implement them, update relevant tests, and use the new API in loaders.

There should be no noticeable difference in behavior after this change.

frasercl:
Resolves #152: `VolumeCache` used to be set up to store arbitrary volume subsets, but is now used to represent a fixed set of volume chunks. We're not likely to ever use the more "advanced" dimension-aware indexing again, so there's no point keeping it around.

This PR changes `VolumeCache` to just index entries by string keys, updates loaders to get and insert using string keys, and updates tests.

### DIFF
--- a/src/VolumeCache.ts
+++ b/src/VolumeCache.ts
@@ -12,8 +12,6 @@ type CacheEntry = {
   key: string;
 };
 
-// Entries are indexed by T and C, then stored in lists of ZYX chunks
-// type CacheStore = CacheEntry[][][][];
 export type CacheStore = Map<string, CacheEntry>;
 
 /** Default: 250MB. Should be large enough to be useful but safe for most any computer that can run the app */
@@ -40,12 +38,12 @@ export default class VolumeCache {
   }
 
   // Hide these behind getters so they're definitely never set from the outside
-  /** The size of all data arrays currently stored in this cache, in bytes */
+  /** The size of all data arrays currently stored in this cache, in bytes. */
   public get size() {
     return this.currentSize;
   }
 
-  /** The number of entries currently stored in this cache */
+  /** The number of entries currently stored in this cache. */
   public get numberOfEntries() {
     return this.currentEntries;
   }
@@ -80,7 +78,7 @@ export default class VolumeCache {
     }
   }
 
-  /** Adds an entry which is *not currently in the list* to the front of the list */
+  /** Adds an entry which is *not currently in the list* to the front of the list. */
   private addEntryAsFirst(entry: CacheEntry): void {
     if (this.first) {
       this.first.prev = entry;
@@ -92,14 +90,14 @@ export default class VolumeCache {
     this.first = entry;
   }
 
-  /** Moves an entry which is *currently in the list* to the front of the list */
+  /** Moves an entry which is *currently in the list* to the front of the list. */
   private moveEntryToFirst(entry: CacheEntry): void {
     if (entry === this.first) return;
     this.removeEntryFromList(entry);
     this.addEntryAsFirst(entry);
   }
 
-  /** Evicts the least recently used entry from the cache */
+  /** Evicts the least recently used entry from the cache. */
   private evictLast(): void {
     if (!this.last) {
       console.error("VolumeCache: attempt to evict last entry from cache when no last entry is set");
@@ -114,15 +112,15 @@ export default class VolumeCache {
     this.last = this.last.prev;
   }
 
-  /** Evicts a specific entry from the cache */
+  /** Evicts a specific entry from the cache. */
   private evict(entry: CacheEntry): void {
     this.removeEntryFromStore(entry);
     this.removeEntryFromList(entry);
   }
 
   /**
-   * Add a new array to the cache (representing a subset of a channel's extent at a given time and scale)
-   * @returns {boolean} a boolean indicating whether the insertion succeeded
+   * Add a new entry to the cache.
+   * @returns {boolean} a boolean indicating whether the insertion succeeded.
    */
   public insert(parentStore: CacheStore, key: string, data: ArrayBuffer): boolean {
     if (data.byteLength > this.maxSize) {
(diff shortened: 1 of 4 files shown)


src/loaders/JsonImageInfoLoader.ts
```diff
@@ -216,15 +214,13 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
-      // construct cache query
-      const cacheQueryDims = { time: volume.loadSpec.time, scale: 0 };
       // Because the data is fetched such that one fetch returns a whole batch,
       // if any in batch is cached then they all should be. So if any in batch is NOT cached,
       // then we will have to do a batch request. This logic works both ways because it's all or nothing.
       let cacheHit = true;
       for (let j = 0; j < Math.min(batch.length, 4); ++j) {
         const chindex = batch[j];
-        const cacheResult = cacheStore && cache?.get(cacheStore, chindex, cacheQueryDims);
+        const cacheResult = cacheStore && cache?.get(cacheStore, `${url}/${chindex}`);
```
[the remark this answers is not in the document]
frasercl: There is as it's currently used, and I suppose I can't think of a reason we'll want to break that relationship

src/VolumeCache.ts
```diff
@@ -139,69 +108,35 @@ export default class VolumeCache {
...
     }
 
     // Check if entry is already in cache
-    for (const existingEntry of entryList) {
-      if (existingEntry.chunk.equals(chunk)) {
-        existingEntry.data = data;
-        this.moveEntryToFirst(existingEntry);
-        return true;
-      }
+    // This will move the entry to the front of the LRU list, if present
```
toloudis, line 128: Not suggesting you change this but it could be interesting if we needed an option to not do this. Like getEntry(key, doNotUpdateUsage)  My thought is that maybe you want to update LRU only on usage, not on insertion.  Totally hypothetical.  I suppose that will be more apparent in the future tickets.

src/VolumeCache.ts
```diff
@@ -210,67 +145,29 @@ export default class VolumeCache {
...
-          channel.forEach((entry) => {
-            this.evict(entry);
-          });
-        });
-      });
-    });
+  public clearWithPrefix(prefix: string): void {
+    for (const [key, entry] of this.entries.entries()) {
+      if (key.startsWith(prefix)) {
+        this.evict(entry);
```
toloudis, line 165: Minor nit: I'm assuming speed is not a concern here.  But I wonder if it might be slightly more efficient to get all the key names first, and then do the deletions? something like: this.entries.keys.reduce(...startsWith...).map(evict(key))

toloudis: or, filter instead of reduce

src/loaders/OmeZarrLoader.ts
```diff
@@ -105,84 +105,47 @@ class SmartStoreWrapper implements AsyncStore<ArrayBuffer, RequestInit> {
...
-    const scale = scaleNames.indexOf(pathElems.shift() as string);
-    // The remaining elements should be chunk coordinates, and should be numbers
-    const coordsLength = getDimensionCount(axesTCZYX);
-    const chunkCoords = pathElems.slice(-coordsLength).map((s) => parseInt(s, 10));
-    if (chunkCoords.length < coordsLength || chunkCoords.some(isNaN) || scale < 0) {
-      console.log("zarr store wrapper: unexpected item path: " + item);
-      return this.baseStore.getItem(item, opts);
-    }
+    const keyPrefix = (this.baseStore as HTTPStore).url || "";
+    const key = keyPrefix + item;
```
toloudis, line 136: Nit: I wonder if a separator character might make it more human readable, like `?`, `/` or `::`

toloudis (approved):
I love these changes, and the cache feels much more like a really good reusable array container now.

src/VolumeCache.ts
```diff
@@ -210,67 +145,29 @@ export default class VolumeCache {
...
-      scale.data.forEach((time) => {
-        time.forEach((channel) => {
-          channel.forEach((entry) => {
-            this.evict(entry);
-          });
-        });
-      });
-    });
+  public clearWithPrefix(prefix: string): void {
+    for (const [key, entry] of this.entries.entries()) {
```
ShrimpCryptid, line 163: Missing a method comment?

frasercl: No longer! Thanks!

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -287,16 +279,10 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
-
-          const cacheInsertDims = {
-            scale: 0,
-            time: volume.loadSpec.time,
-            channel: batch[ch],
-          };
-          cacheStore && cache?.insert(cacheStore, volume.channels[batch[ch]].volumeData.buffer, cacheInsertDims);
-
-          onChannelLoaded?.(volume, batch[ch]);
+          const chindex = batch[ch];
```
ShrimpCryptid, line 282: Nit: Maybe spell out `channelIndex`, or use capitalization for `chIndex`?

ShrimpCryptid (approved):
Wow, this is so much cleaner! I had a much easier time reading the new changes, super happy to see this.

src/VolumeCache.ts
frasercl: Is your concern over deleting from a collection while also iterating over it?

src/loaders/JsonImageInfoLoader.ts
frasercl: This is matching a naming convention above, around line 218

src/VolumeCache.ts
toloudis: Yes but it's not a significant concern, just musing about alternate ways.

PR #163, opened 2023-11-06
Load single slices at higher resolution

### LINKED ISSUE #157: Reload slice if it will load at a higher resolution level
## Description

Currently, when the user switches from 3D to z-slice mode in a zarr image, the image remains at the scale level chosen for 3d rather than loading a higher scale level for 2D.

## Expected Behavior

The viewer should recognize when a new load would fetch a different scale level and issue a new load.

## Solution

This is due to the check made by `Volume.updateRequiredData`: it determines that since the time has not changed and the 2d slice is contained within the 3d volume, it does not need to load new data.

Part of the solution could come from adapting the loader interface method `loadDims`. This method is currently unused in the app, and the type it returns, `VolumeDims`, has remained constant through five or six revisions of the similar `LoadSpec` type. This seems like the appropriate method to ask a loader which scale level it would load with a given `LoadSpec`.

This is also a light form of #115, since the slice will be immediately displayed at 3D resolution first before new data is loaded. Future work might be to check if the data required by a load is present in the cache at a lower scale level, so that it can be loaded immediately while new data is fetched from the server.

frasercl:
Resolves #157: If a 3d volume is loaded, the user switches to z-slice mode, and a higher resolution level is available to load, reload at that higher resolution.

This is accomplished using the `loadDims` method on `IVolumeLoader` (unused until now), which loads the dimensions of a volume without loading any data. I made the following changes to make `loadDims` a good solution to this issue:
- `loadDims` returns an array of `VolumeDims` objects for each available scale level. I added the property `canLoad: boolean` to this type to indicate whether a level can be loaded given the specifications.
- Because `loadDims` is `async`, `updateRequiredData` must now be `async` too. This is a bummer, so to compensate, I tried to ensure that `loadDims` rarely or never is `async` in practice.

I also got rid of the `scene` property of `LoadSpec`, given that the only loader that supports scenes right now (`OmeZarrLoader`) must be bound to a single scene on construction.

### DIFF
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -229,13 +229,22 @@ export default class Volume {
     this.normRegionOffset = subregionOffset.clone().divide(volumeSize);
   }
 
-  updateRequiredData(required: Partial<LoadSpec>, onChannelLoaded?: PerChannelCallback) {
+  async updateRequiredData(required: Partial<LoadSpec>, onChannelLoaded?: PerChannelCallback): Promise<void> {
     this.loadSpecRequired = { ...this.loadSpecRequired, ...required };
+    const containsSubregion = this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion);
+    let subregionRequiresReload = !containsSubregion;
+    if (containsSubregion) {
+      const dims = await this.loader?.loadDims(this.loadSpecRequired);
+      const currentScale = this.loadSpecRequired.multiscaleLevel;
+      if (dims && currentScale > 0 && dims[currentScale - 1].canLoad) {
+        subregionRequiresReload = true;
+      }
+    }
     // if newly required data is not currently contained in this volume...
     if (
       this.loadSpecRequired.time !== this.loadSpec.time ||
       this.loadSpecRequired.scene !== this.loadSpec.scene ||
-      !this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion)
+      subregionRequiresReload
     ) {
       // ...clone `loadSpecRequired` into `loadSpec` and load
       this.setUnloaded();
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -240,7 +240,6 @@ export default class Volume {
     this.loadSpecRequired = { ...this.loadSpecRequired, ...required };
     let noReload =
       this.loadSpec.time === this.loadSpecRequired.time &&
-      this.loadSpec.scene === this.loadSpecRequired.scene &&
       this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion);
 
     // An update to `subregion` should trigger a reload when the new subregion is not contained in the old one
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -4,6 +4,7 @@ import Channel from "./Channel";
 import Histogram from "./Histogram";
 import { getColorByChannelIndex } from "./constants/colors";
 import { IVolumeLoader, LoadSpec, PerChannelCallback } from "./loaders/IVolumeLoader";
+import { estimateLevelForAtlas } from "./loaders/VolumeLoaderUtils";
 
 export type ImageInfo = Readonly<{
   name: string;
@@ -247,10 +248,13 @@ export default class Volume {
     if (noReload && !this.loadSpec.subregion.equals(this.loadSpecRequired.subregion)) {
       const currentScale = this.imageInfo.multiscaleLevel;
       // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
-      const minScale = this.loadSpec.multiscaleLevel ?? 0;
+      const minLevel = this.loadSpec.multiscaleLevel ?? 0;
       // Loaders should cache loaded dimensions so that this call blocks no more than once per valid `LoadSpec`.
       const dims = await this.loader?.loadDims(this.loadSpecRequired);
-      noReload = !dims || currentScale <= minScale || !dims[currentScale - 1].canLoad;
+      if (dims) {
+        const loadableLevel = estimateLevelForAtlas(dims.map(({ shape }) => [shape[2], shape[3], shape[4]]));
+        noReload = !dims || currentScale <= Math.max(loadableLevel, minLevel);
+      }
     }
 
     // if newly required data is not currently contained in this volume...
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -235,28 +235,29 @@ export default class Volume {
     this.normRegionOffset = subregionOffset.clone().divide(volumeSize);
   }
 
+  /** Call on any state update that may require new data to be loaded (subregion, enabled channels, time, etc.) */
   async updateRequiredData(required: Partial<LoadSpec>, onChannelLoaded?: PerChannelCallback): Promise<void> {
     this.loadSpecRequired = { ...this.loadSpecRequired, ...required };
+    let noReload =
+      this.loadSpec.time === this.loadSpecRequired.time &&
+      this.loadSpec.scene === this.loadSpecRequired.scene &&
+      this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion);
+
     // An update to `subregion` should trigger a reload when the new subregion is not contained in the old one
     // OR when the new subregion is smaller than the old one by enough that we can load a higher scale level.
-    let subregionRequiresReload = !this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion);
-    if (!subregionRequiresReload) {
+    if (noReload && !this.loadSpec.subregion.equals(this.loadSpecRequired.subregion)) {
       const currentScale = this.imageInfo.multiscaleLevel;
       // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
       const minScale = this.loadSpec.multiscaleLevel ?? 0;
-      // Loaders should cache loaded dimensions such that this call blocks no more than once per valid `LoadSpec`.
+      // Loaders should cache loaded dimensions so that this call blocks no more than once per valid `LoadSpec`.
       const dims = await this.loader?.loadDims(this.loadSpecRequired);
       if (dims && currentScale > minScale && dims[currentScale - 1].canLoad) {
-        subregionRequiresReload = true;
+        noReload = false;
       }
     }
 
     // if newly required data is not currently contained in this volume...
-    if (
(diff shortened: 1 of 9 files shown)


src/Volume.ts
```diff
@@ -229,14 +235,28 @@ export default class Volume {
...
+      this.loadSpec.time === this.loadSpecRequired.time &&
+      this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion);
+
+    // An update to `subregion` should trigger a reload when the new subregion is not contained in the old one
+    // OR when the new subregion is smaller than the old one by enough that we can load a higher scale level.
+    if (noReload && !this.loadSpec.subregion.equals(this.loadSpecRequired.subregion)) {
+      const currentScale = this.imageInfo.multiscaleLevel;
+      // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
+      const minScale = this.loadSpec.multiscaleLevel ?? 0;
+      // Loaders should cache loaded dimensions so that this call blocks no more than once per valid `LoadSpec`.
```
toloudis, line 251: just wondering, why not pre-load all level dims the first time the loader is instantiated?

src/Volume.ts
```diff
@@ -41,6 +40,11 @@ export type ImageInfo = Readonly<{
   /** Symbol of temporal unit used by `timeScale`, e.g. "hr" */
   timeUnit: string;
 
+  /** Number of scale levels available for this volume */
+  numMultiscaleLevels: number;
+  /** The scale level from which this image was loaded, between `0` and `numMultiscaleLevels` */
```
toloudis, line 45:
```suggestion
  /** The scale level from which this image was loaded, between `0` and `numMultiscaleLevels-1` */
```

src/loaders/IVolumeLoader.ts
```diff
@@ -13,17 +12,17 @@ export class LoadSpec {
...
 
 export class VolumeDims {
-  subpath = "";
   // shape: [t, c, z, y, x]
   shape: number[] = [0, 0, 0, 0, 0];
   // spacing: [t, c, z, y, x]; generally expect 1 for non-spatial dimensions
   spacing: number[] = [1, 1, 1, 1, 1];
   spaceUnit = "μm";
   timeUnit = "s";
+  canLoad = true;
```
toloudis, line 25: I suggest an explanatory comment here, something like:  
```
// dims are small enough to fit within our preferred memory footprint (see `pickLevelToLoad`)
```

src/Volume.ts
```diff
@@ -229,14 +235,28 @@ export default class Volume {
...
+
+    // An update to `subregion` should trigger a reload when the new subregion is not contained in the old one
+    // OR when the new subregion is smaller than the old one by enough that we can load a higher scale level.
+    if (noReload && !this.loadSpec.subregion.equals(this.loadSpecRequired.subregion)) {
+      const currentScale = this.imageInfo.multiscaleLevel;
+      // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
+      const minScale = this.loadSpec.multiscaleLevel ?? 0;
+      // Loaders should cache loaded dimensions so that this call blocks no more than once per valid `LoadSpec`.
+      const dims = await this.loader?.loadDims(this.loadSpecRequired);
+      if (dims && currentScale > minScale && dims[currentScale - 1].canLoad) {
```
toloudis, line 253: I wonder if the canLoad could be a dynamic check here?  I don't think it's too slow to call pickLevelToLoad if it doesn't require network fetches but I forget how frequently this gets called.

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -104,24 +106,29 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
 
-  async getJsonImageInfo(loadSpec: LoadSpec): Promise<JsonImageInfo> {
-    const response = await fetch(this.urls[loadSpec.time]);
+  private async getJsonImageInfo(time: number): Promise<JsonImageInfo> {
+    const cachedInfo = this.jsonInfo[time];
+    if (cachedInfo) {
+      return cachedInfo;
+    }
+
+    const response = await fetch(this.urls[time]);
```
ShrimpCryptid, line 119: I like this optimization!

ShrimpCryptid (commented):
Agree with Dan that I'm not totally sold on `VolumeDims` tracking whether it can be loaded? It seems like that should be tracked by a system external to it, but I'm fine to talk about tradeoffs too.

src/Volume.ts
frasercl: I guess I felt like the decision of which level to load was the loader's responsibility, since `pickLevelToLoad` and the `MAX_ATLAS_DIMENSION` constant live in the same file as `OMEZarrLoader`. I kind of agree though that this ought to be treated more like a global limit. Do we make that official by moving those items into VolumeLoaderUtils and importing/calling `pickLevelToLoad` here? This is as opposed to calling it internally in `loadDims`, which is what is done now.

`pickLevelToLoad` itself doesn't require network fetches, it just takes dimensions and performs a bit of logic. Those dimensions, however, need to be fetched. This is the problem I referred to in the PR description and in the comment on line 251 above: loaders need to be smart enough to ensure that providing dimensions rarely requires a fetch in practice. The behavior in this PR is:
- In `OMEZarrLoader`, all dims are fetched before loader construction and `loadDims` will never block
- In `TiffLoader`, metadata is fetched on the first call to `loadDims` or `loadVolumeData` and cached (so if the volume already has data, `loadVolumeData` has been called and `loadDims` won't block in practice)
- In `JsonImageInfoLoader`, metadata is fetched on the first call to `loadDims` or `loadVolumeData` *per timestep* and cached. In practice, if this function calls `loadVolumeData` after calling `loadDims`, the first `loadDims` call will have cached the metadata that the `loadVolumeData` would have required so checking dims first adds no additional work.

Moving the `pickLevelToLoad` call won't change these rules on when this function has to wait for the network.

toloudis: Yea I basically agree with everything you said here.  
* pickleveltoload can be a utils-like shared function, or even a IVolumeLoader method that returns 0 by default and only does something nontrivial in omezarrloader.
* all dims could be fetched at or close to loader creation time.  Then they will just be available and no call to loadDims would have to fetch anything (though they could if they wanted?)

src/Volume.ts
```diff
@@ -247,11 +248,12 @@ export default class Volume {
...
       // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
-      const minScale = this.loadSpec.multiscaleLevel ?? 0;
+      const minLevel = this.loadSpec.multiscaleLevel ?? 0;
       // Loaders should cache loaded dimensions so that this call blocks no more than once per valid `LoadSpec`.
       const dims = await this.loader?.loadDims(this.loadSpecRequired);
-      if (dims && currentScale > minScale && dims[currentScale - 1].canLoad) {
-        noReload = false;
+      if (dims) {
+        const loadableLevel = estimateLevelForAtlas(dims.map(({ shape }) => [shape[2], shape[3], shape[4]]));
+        noReload = !dims || currentScale <= Math.max(loadableLevel, minLevel);
```
ShrimpCryptid, line 256: `dims` is guaranteed to be truthy here already, so `!dims` always evaluates to false...
```suggestion
        noReload = currentScale <= Math.max(loadableLevel, minLevel);
```

ShrimpCryptid:
I tried running this to test it, and I ran into an error?

Repro:
1. Run the project (`npm run start`)
2. Switch to OME Zarr (NucMorph)
3. Switch to Z viewer
```
Uncaught Runtime Errors:
ERROR
chunk request cancelled
    at handleError (http://localhost:9020/volume-viewer-ui.bundle.js:25109:58)
    at http://localhost:9020/volume-viewer-ui.bundle.js:25132:7
ERROR
chunk request cancelled
    at handleError (http://localhost:9020/volume-viewer-ui.bundle.js:25109:58)
    at http://localhost:9020/volume-viewer-ui.bundle.js:25132:7
ERROR
chunk request cancelled
    at handleError (http://localhost:9020/volume-viewer-ui.bundle.js:25109:58)
    at http://localhost:9020/volume-viewer-ui.bundle.js:25132:7
```

It's possible this is related to the Vast outage today. Will check again later.

ShrimpCryptid:
> I tried running this to test it, and I ran into an error?
> 
> Repro:
> 
> 1. Run the project (`npm run start`)
> 2. Switch to OME Zarr (NucMorph)
> 3. Switch to Z viewer
> 
> ```
> Uncaught Runtime Errors:
> ERROR
> chunk request cancelled
>     at handleError (http://localhost:9020/volume-viewer-ui.bundle.js:25109:58)
>     at http://localhost:9020/volume-viewer-ui.bundle.js:25132:7
> ERROR
> chunk request cancelled
>     at handleError (http://localhost:9020/volume-viewer-ui.bundle.js:25109:58)
>     at http://localhost:9020/volume-viewer-ui.bundle.js:25132:7
> ERROR
> chunk request cancelled
>     at handleError (http://localhost:9020/volume-viewer-ui.bundle.js:25109:58)
>     at http://localhost:9020/volume-viewer-ui.bundle.js:25132:7
> ```
> 
> It's possible this is related to the Vast outage today. Will check again later.

Still getting this bug even on VPN.

This is all I'm getting from the console:
```
index.ts:872 currentVol with name AICS-12_881 is loaded
index.ts:872 currentVol with name P13-C4 is loaded
localhost/:1 Uncaught (in promise) chunk request cancelled
index.ts:872 currentVol with name P13-C4 is loaded
```

toloudis:
If I have any coding time today I'll try to take a look at this.  This is an important change to get in soon.

toloudis:
The error here is indeed an uncaught promise exception that occurs when we cancel a request.
This pull request : https://github.com/gzuidhof/zarr.js/pull/145 actually seems to fix the issue.  The problem seems to be that our promises get stuck in the zarr.js plumbing because of their internal logic in between our proxy store's SmartStoreWrapper.getRaw and the resulting multiple calls to the true store's getItem.

You can see some hints of this by changing
`      return this.requestQueue.addRequest(key, () => this.getAndCacheItem(item, key, opts));`
to
```
return this.requestQueue
        .addRequest(key, () => this.getAndCacheItem(item, key, opts))
        .catch((e) => {
          if (e !== CHUNK_REQUEST_CANCEL_REASON) {
            throw e; // this will be unhandled!!!
          }
          // if cancelled, we still need to return an arraybuffer of the expected size,
          // or else handle this one - returning this buffer still breaks things.
          return new ArrayBuffer(0);
        });
```

PR #164, opened 2023-11-08
Subscribable request queue

frasercl:
Progress on #154: adds `SubscribableRequestQueue`, which wraps `RequestQueue` and adds the concept of "subscribers." Subscribers can cancel their subscriptions to a request without cancelling the underlying request.

Does NOT add `SubscribableRequestQueue` to `OMEZarrLoader` yet.

### DIFF
--- a/src/test/SubscribableRequestQueue.test.ts
+++ b/src/test/SubscribableRequestQueue.test.ts
@@ -53,6 +53,19 @@ describe("SubscribableRequestQueue", () => {
       expect(result).to.equal("foo");
     });
 
+    it("does not queue a request if the subscriber id is invalid", () => {
+      const queue = new SubscribableRequestQueue();
+      expect(() => queue.addRequestToQueue("test", -1, () => delay(TIMEOUT, "foo"))).to.throw();
+      expect(() => queue.addRequestToQueue("test", 0, () => delay(TIMEOUT, "foo"))).to.throw();
+    });
+
+    it("does not queue a request if the subscriber has been deleted", () => {
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+      queue.removeSubscriber(id);
+      expect(() => queue.addRequestToQueue("test", id, () => delay(TIMEOUT, "foo"))).to.throw();
+    });
+
     it("creates unique promises without duplicating work when two subscribers queue the same key", async () => {
       const queue = new SubscribableRequestQueue();
       const id1 = queue.addSubscriber();
--- a/src/test/SubscribableRequestQueue.test.ts
+++ b/src/test/SubscribableRequestQueue.test.ts
@@ -244,8 +244,8 @@ describe("SubscribableRequestQueue", () => {
       const id2 = queue.addSubscriber();
 
       const promise1 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "foo"));
-      const promise2 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "bar"));
-      const promise3 = queue.addRequestToQueue("test", id2, () => delay(TIMEOUT, "baz"));
+      const promise2 = queue.addRequestToQueue("test2", id1, () => delay(TIMEOUT, "bar"));
+      const promise3 = queue.addRequestToQueue("test", id2, () => delay(TIMEOUT, "foo"));
       expect(queue.hasRequest("test")).to.be.true;
       expect(queue.isSubscribed(id1, "test")).to.be.true;
       expect(queue.isSubscribed(id2, "test")).to.be.true;
@@ -253,6 +253,7 @@ describe("SubscribableRequestQueue", () => {
       queue.removeSubscriber(id1);
       expect(queue.hasRequest("test")).to.be.true;
       expect(queue.isSubscribed(id1, "test")).to.be.false;
+      expect(queue.isSubscribed(id1, "test2")).to.be.false;
       expect(await isRejected(promise1)).to.be.true;
       expect(await isRejected(promise2)).to.be.true;
       expect(await isRejected(promise3)).to.be.false;
--- a/src/test/SubscribableRequestQueue.test.ts
+++ b/src/test/SubscribableRequestQueue.test.ts
@@ -168,9 +168,11 @@ describe("SubscribableRequestQueue", () => {
       expect(queue.hasRequest("test")).to.be.true;
       expect(queue.isSubscribed(id, "test")).to.be.true;
 
-      queue.cancelRequest("test", id);
+      const cancelResult = queue.cancelRequest("test", id);
+      expect(cancelResult).to.be.true;
       expect(queue.isSubscribed(id, "test")).to.be.false;
       expect(await isRejected(promise)).to.be.true;
+      expect(queue.cancelRequest("test", id)).to.be.false;
     });
 
     it("does not cancel an underlying request if it is running", async () => {
--- a/src/test/SubscribableRequestQueue.test.ts
+++ b/src/test/SubscribableRequestQueue.test.ts
@@ -1,0 +1,248 @@
+import { expect } from "chai";
+
+import SubscribableRequestQueue from "../utils/SubscribableRequestQueue";
+
+const TIMEOUT = 10;
+const LONG_TIMEOUT = 20;
+
+/** Returns a promise that resolves after `timeoutMs` milliseconds with value `result`. */
+function delay<T>(timeoutMs: number, result: T): Promise<T> {
+  return new Promise((resolve) => setTimeout(() => resolve(result), timeoutMs));
+}
+
+/** Returns a promise that rejects after `timeoutMs` milliseconds with reason `reason`. */
+function delayReject(timeoutMs: number, reason: unknown): Promise<unknown> {
+  return new Promise((_, reject) => setTimeout(() => reject(reason), timeoutMs));
+}
+
+/** Determines whether a promise has rejected. */
+// From https://gist.github.com/tyru/29360dfa475d2fefaf6c4655a93c2cb0
+function isRejected(promise: Promise<unknown>): Promise<boolean> {
+  return Promise.race([
+    delay(0, false),
+    promise.then(
+      () => false,
+      () => true
+    ),
+  ]);
+}
+
+describe("SubscribableRequestQueue", () => {
+  describe("addSubscriber", () => {
+    it("adds a subscriber", () => {
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+      expect(queue.hasSubscriber(id)).to.be.true;
+    });
+
+    it("returns unique ids", () => {
+      const queue = new SubscribableRequestQueue();
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+      expect(id1).to.not.equal(id2);
+    });
+  });
+
+  describe("addRequestToQueue", () => {
+    it("queues a request", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+      const promise = queue.addRequestToQueue("test", id, () => delay(TIMEOUT, "foo"));
+      expect(queue.hasRequest("test")).to.be.true;
+      const result = await promise;
+      expect(result).to.equal("foo");
+    });
+
+    it("creates unique promises without duplicating work when two subscribers queue the same key", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test", id2, () => delay(TIMEOUT, "bar"));
+      expect(queue.hasRequest("test")).to.be.true;
+      expect(promise1).to.not.equal(promise2);
+
+      const [result1, result2] = await Promise.all([promise1, promise2]);
+      expect(result1).to.equal("foo");
+      expect(result2).to.equal("foo");
+    });
+
+    it("creates different requests when one subscriber queues two keys", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test1", id, () => delay(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test2", id, () => delay(TIMEOUT, "bar"));
+      expect(queue.hasRequest("test1")).to.be.true;
+      expect(queue.hasRequest("test2")).to.be.true;
+      expect(promise1).to.not.equal(promise2);
+
+      const [result1, result2] = await Promise.all([promise1, promise2]);
+      expect(result1).to.equal("foo");
+      expect(result2).to.equal("bar");
+    });
+
+    it("rejects and reissues a request when one subscriber queues the same key twice", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id, () => delay(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test", id, () => delay(TIMEOUT, "bar"));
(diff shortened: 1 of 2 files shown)


src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * An extension of `RequestQueue` that adds a concept of "subscribers," which may share references to a single request
+ * or cancel their subscription without disrupting the request for other subscribers.
+ */
+export default class SubscribableRequestQueue {
+  private queue: RequestQueue;
+
+  /** Number of subscribers, used for generating unique subscriber IDs. */
+  private subscriberIds: number;
```
ShrimpCryptid, line 21: Nit: `numSubscribers` might be clearer, or `numSubscriberIds`? The comment is useful but when I see this variable name I expect it to be an array of subscriber IDs or something.

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,171 @@
...
+  private requests: Map<string, RequestSubscription[]>;
+
+  constructor(maxActiveRequests?: number) {
+    this.queue = new RequestQueue(maxActiveRequests);
+    this.subscriberIds = 0;
+    this.subscribers = new Map();
+    this.requests = new Map();
+  }
+
+  /** Resolves all subscriptions to request `key` with `value */
```
ShrimpCryptid, line 34:
```suggestion
  /** Resolves all subscriptions to request `key` with `value` */
```

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,171 @@
...
+    // Create single underlying request if it does not yet exist
+    if (!this.queue.hasRequest(key)) {
+      this.queue
+        .addRequest(key, requestAction, delayMs)
+        .then((value) => this.resolveAll(key, value))
+        .catch((reason) => this.rejectAll(key, reason));
+    }
+    if (!this.requests.has(key)) {
+      this.requests.set(key, []);
+    }
```
ShrimpCryptid, line 82: Would there ever be an instance where the keys of `queue` and `request` become desynced?

ShrimpCryptid: (I don't think there would be, looking at the code? This just feels funny to me but doesn't do anything bad)

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,171 @@
...
+    const existingRequest = subscriber.get(key);
+    if (existingRequest) {
+      this.rejectSubscription(key, existingRequest, "SubscribableRequestQueue: request re-queued while running");
+    }
+
+    // Create promise and add to list of requests
+    return new Promise<T>((resolve, reject) => {
+      this.requests.get(key)?.push({ resolve, reject, subscriberId });
+      this.subscribers.get(subscriberId)?.set(key, reject);
+    });
```
ShrimpCryptid, line 101: Does rejecting and then re-adding the subscription have any consequences for the caller? Would it be better to just return the same promise as before?

ShrimpCryptid: Probably worth documenting this behavior in the method comment as well

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  constructor(maxActiveRequests?: number) {
+    this.queue = new RequestQueue(maxActiveRequests);
+    this.subscriberIds = 0;
+    this.subscribers = new Map();
+    this.requests = new Map();
+  }
+
+  /** Resolves all subscriptions to request `key` with `value` */
+  private resolveAll<T>(key: string, value: T) {
```
ShrimpCryptid, line 35: I'm surprised the linter doesn't require you to add explicit void return types on these. Maybe the Nucmorph-Colorizer project has different settings.

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    });
+
+    it("cancels all subscriptions from a removed subscriber", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "bar"));
+      const promise3 = queue.addRequestToQueue("test", id2, () => delay(TIMEOUT, "baz"));
```
ShrimpCryptid, line 248: Should `promise1` and `promise2` have different keys here? `promise1` would get rejected right away by the second `addRequestToQueue` call, which means you'd only be testing that one subscription is cancelled? (vs. all subscriptions getting cancelled)

ShrimpCryptid (approved):
Looks good, really sensible tests and the code is well-commented! Left a few questions/comments here and there.

src/test/SubscribableRequestQueue.test.ts
frasercl: The perils of co-authoring your tests with Copilot 🙃. Great catch, thank you!

src/utils/SubscribableRequestQueue.ts
frasercl: Yeah, we should never see only one or the other of these `if`s evaluate to true, but I still separated them for extra robustness.

ShrimpCryptid (approved):
LGTM!

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,175 @@
+import RequestQueue from "./RequestQueue";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Resolver = (value?: any) => void;
```
toloudis, line 4: could you use unknown instead of any here, to quiet the linter?

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,175 @@
...
+
+type RequestSubscription = {
+  subscriberId: number;
+  resolve: Resolver;
+  reject: Rejecter;
+};
+
+/**
+ * An extension of `RequestQueue` that adds a concept of "subscribers," which may share references to a single request
+ * or cancel their subscription without disrupting the request for other subscribers.
```
toloudis, line 15: just to refresh my memory, does this mean that each subscriber gets a unique promise instead of having one shared promise for each unique data request?

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+export default class SubscribableRequestQueue {
+  private queue: RequestQueue;
+
+  /** Number of subscribers, used for generating unique subscriber IDs. */
+  private numSubscribers: number;
+  /** Map keyed by subscriber ID. Subscribers are only useful for cancelling early, so we only store rejecters here. */
```
toloudis, line 22: the comment doesn't make it clear what the Map<string, Rejecter> is for. We're keying by subscriberid but what is the value?

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,175 @@
...
+    }
+    subscriber?.delete(key);
+  }
+
+  /** Removes a subscriber and cancels its remaining subscriptions. */
+  removeSubscriber(subscriberId: number, cancelReason?: unknown): void {
+    const subscriptions = this.subscribers.get(subscriberId);
+    if (subscriptions) {
+      for (const [key, reject] of subscriptions.entries()) {
+        this.rejectSubscription(key, reject, cancelReason);
```
toloudis, line 150: is this modifying the list while iterating it?

src/utils/SubscribableRequestQueue.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    // Remove the underlying request if there are no more subscribers and the request is not already running
+    if (subscriptions.length < 1 && !this.queue.requestRunning(key)) {
+      this.queue.cancelRequest(key, cancelReason);
+      this.requests.delete(key);
+    }
+  }
+
+  /** Cancels a request subscription, and cancels the underlying request if it is no longer subscribed or running. */
+  cancelRequest(key: string, subscriberId: number, cancelReason?: unknown): void {
```
toloudis, line 136: is there any need for a cancelRequest that is independent of subscribers (e.g. `cancelForAllSubscribers(key, reason)`)?  Or does it always need to be initiated by an individual subscriber?

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/** Returns a promise that rejects after `timeoutMs` milliseconds with reason `reason`. */
+function delayReject(timeoutMs: number, reason: unknown): Promise<unknown> {
+  return new Promise((_, reject) => setTimeout(() => reject(reason), timeoutMs));
+}
+
+/** Determines whether a promise has rejected. */
+// From https://gist.github.com/tyru/29360dfa475d2fefaf6c4655a93c2cb0
+function isRejected(promise: Promise<unknown>): Promise<boolean> {
```
toloudis, line 20: wow.

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+      expect(queue.hasSubscriber(id)).to.be.true;
+    });
+
+    it("returns unique ids", () => {
+      const queue = new SubscribableRequestQueue();
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+      expect(id1).to.not.equal(id2);
```
toloudis, line 42: I'd add one more test to remove one of id1 or id2, and then add an id3, and confirm it's also unique.

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      const queue = new SubscribableRequestQueue();
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test", id2, () => delay(TIMEOUT, "bar"));
+      expect(queue.hasRequest("test")).to.be.true;
+      expect(promise1).to.not.equal(promise2);
+
+      const [result1, result2] = await Promise.all([promise1, promise2]);
```
toloudis, line 79: this test makes me think that subscriber 1 wants to do the "foo" result and subscriber2 wants "bar" when the same data request arrives.  Can two subscribers use the same key and do different things when the request completes?

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    });
+
+    it("rejects and reissues a request when one subscriber queues the same key twice", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id, () => delay(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test", id, () => delay(TIMEOUT, "bar"));
+      expect(queue.hasRequest("test")).to.be.true;
+      expect(await isRejected(promise1)).to.be.true;
```
toloudis, line 106: Would it be a plausible alternative to just drop the reference to the "foo" result and simply replace it with the "bar" result, without rejecting anything?

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      expect(result2).to.equal("foo");
+    });
+
+    it("passes request rejections to all subscribers", async () => {
+      const queue = new SubscribableRequestQueue();
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id1, () => delayReject(TIMEOUT, "foo"));
+      const promise2 = queue.addRequestToQueue("test", id2, () => delay(TIMEOUT, "bar"));
```
toloudis, line 118: this one is also slightly confusing to me - same request key, different subscribers.  Why does "foo" win instead of "bar"?

src/test/SubscribableRequestQueue.test.ts
toloudis: I think my confusion was in forgetting that the last arg is the whole process and not the "then" part.   It seems odd to add requests consecutively in this way, when they don't know about each other and are passing in totally different functions.

src/test/SubscribableRequestQueue.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      const id1 = queue.addSubscriber();
+      const id2 = queue.addSubscriber();
+
+      const promise1 = queue.addRequestToQueue("test", id1, () => delay(TIMEOUT, "foo"));
+      expect(queue.hasRequest("test")).to.be.true;
+      expect(queue.requestRunning("test")).to.be.true;
+      expect(queue.isSubscribed(id1, "test")).to.be.true;
+      expect(queue.isSubscribed(id2, "test")).to.be.false;
+
+      queue.cancelRequest("test", id1);
```
toloudis, line 177: maybe this could return a bool if it couldn't actually cancel anything?

src/utils/SubscribableRequestQueue.ts
frasercl: Tried this, and it seems like `unknown` works for `Rejecter` and not `Resolver` because the argument for `Rejecter` never has a narrower type than `unknown`. But putting `unknown` in `Resolver` causes an error when we try to put the actual resolve function from a promise into a field of type `Resolver`:

```
let resolver: Resolver;
new Promise<T>((resolve) => resolver = resolve); // error: type 'T' is not assignable to type 'unknown'
```

I tried making `Resolver` generic as well, but then we can't store `Resolvers` of lots of different types in one map without using `any`:

`let resolvers: Map<string, Resolver<???>>;`

Either way, `any` does seem to be the only way out here.

src/utils/SubscribableRequestQueue.ts
frasercl: Right, this class owns a `RequestQueue` and that underlying queue still creates promises that run a task and resolve when complete. This class is just about managing a set of subscriptions (which are themselves promises) to those promises and saying "when the queued promise resolves with a value, resolve all the subscribing promises with that value."

src/utils/SubscribableRequestQueue.ts
frasercl: I can't think of a use for one, but it would be simple enough to add since it just has to call the equivalent method on `RequestQueue`.

This does remind me of thoughts I've been having around how this class might relate to `RequestQueue`: would it be useful to allow `SubscribableRequestQueue` to be treated as a wrapper around `RequestQueue`? (i.e., allow passing in a `RequestQueue` to the constructor.) I could imagine, for example, using the bare `RequestQueue` for queueing prefetches on simple loaders, e.g. JSON time series frames, and only using the subscription abstraction for `OMEZarrLoader` (or any hypothetical future loader where data is chunked along more than one dimension).

src/test/SubscribableRequestQueue.test.ts
frasercl: Right, the "bar" task never starts executing. The queue assumes that clients will ensure that requests with the same key, even when initiated by different subscribers, do effectively the same thing. For example, a request keyed by a specific chunk URL will always be a request to fetch that chunk. Given that assumption, it makes sense for a new subscription to an existing key to join the work in progress rather than interrupt it and start over. That's what this test is about.

src/test/SubscribableRequestQueue.test.ts
frasercl: Answered above

src/test/SubscribableRequestQueue.test.ts
toloudis: Okay.  So my problem is that the api is weird because the caller is expected to pass in what might be a totally unique task function and it's possible that it will never ever run.  How can the caller receive that information (that their task isn't going to happen because someone else already submitted a task for this)?  (Or should the caller just not care?)

PR #165, opened 2023-11-15
better single slice render

toloudis:
Some small improvements in this PR:

1. Single-slice rendering was a bit too dim. When you switch modes, the colors were really muted.  This was because of remnants of the volumetric shader still in the single-slice shader altering the alpha value.  I simplified the single slice shader considerably.
2. The z slider was initialized to 0 even though the volume is initialized to the middle slice.  Slider now inits to middle value.
3. Add z+1 and z-1 buttons to test caching when chunks consist of more than one slice
4. Remember pathtrace vs raymarch when switching between z and 3d mode.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -586,6 +586,9 @@ function updateZSliceUI(volume: Volume) {
   const totalZSlices = volume.imageInfo.volumeSize.z;
   zSlider.max = `${totalZSlices}`;
   zInput.max = `${totalZSlices}`;
+
+  zInput.value = `${Math.floor(totalZSlices / 2)}`;
+  zSlider.value = `${Math.floor(totalZSlices / 2)}`;
 }
 
 function showChannelUI(volume: Volume) {
@@ -1186,8 +1189,20 @@ function main() {
   });
 
   // Set up Z-slice UI
+  const zforwardBtn = document.getElementById("zforwardBtn");
+  const zbackBtn = document.getElementById("zbackBtn");
   const zSlider = document.getElementById("zSlider") as HTMLInputElement;
   const zInput = document.getElementById("zValue") as HTMLInputElement;
+  zforwardBtn?.addEventListener("click", () => {
+    if (goToZSlice(zSlider?.valueAsNumber + 1)) {
+      //
+    }
+  });
+  zbackBtn?.addEventListener("click", () => {
+    if (goToZSlice(zSlider?.valueAsNumber - 1)) {
+      //
+    }
+  });
   zSlider?.addEventListener("change", () => {
     goToZSlice(zSlider?.valueAsNumber);
   });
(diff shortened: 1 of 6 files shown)


src/View3d.ts
```diff
@@ -831,19 +831,24 @@ export class View3d {
    * Switch between single pass ray-marched volume rendering and progressive path traced rendering.
    * @param {number} mode 0 for single pass ray march, 1 for progressive path trace
    */
-  setVolumeRenderMode(mode: number): void {
+  setVolumeRenderMode(mode: RenderMode.PATHTRACE | RenderMode.RAYMARCH): void {
```
frasercl, line 834: Slightly odd to just expose these two options at this level... makes me think that, if other orthographic viewing angles get limited to a single slice, we could combine camera mode and render mode into a single concept (volumetric, pathtrace, slice x/y/z)

src/loaders/OmeZarrLoader.ts
```diff
@@ -147,6 +148,7 @@ class SmartStoreWrapper implements AsyncStore<ArrayBuffer, RequestInit> {
     if (this.requestQueue) {
       return this.requestQueue.addRequest(key, () => this.getAndCacheItem(item, key, opts));
     } else {
+      // Should we ever hit this code?  We should always have a request queue.
```
frasercl, line 151: Currently, we don't, since the zarr loader builds its own queue. In the future, we could end up sharing a queue by passing it into loader constructors, as currently happens with the cache. The cache is currently optional and null-checked, since we can get on just fine without it, and since accepting it via the constructor means we rely on the client to build one themselves first. This is following the same logic.

frasercl (approved):
I left several mostly nitpicky comments, but these all look like good changes overall!

public/index.ts
```diff
@@ -1186,8 +1195,20 @@ function main() {
...
+  zforwardBtn?.addEventListener("click", () => {
+    if (goToZSlice(zSlider?.valueAsNumber + 1)) {
+      //
+    }
+  });
+  zbackBtn?.addEventListener("click", () => {
+    if (goToZSlice(zSlider?.valueAsNumber - 1)) {
+      //
+    }
+  });
```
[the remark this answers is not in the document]
toloudis: good question, I can clean that up I think.

src/View3d.ts
toloudis: Sadly, the enum is overloaded here. `setVolumeRenderMode` would never have any other values than these - it's only setting how the 3d volumetric rendering mode should behave.  The max-projection flag is also slightly confusingly blended in here.  It can be improved but at least this code change clarifies the contract and intention a little better.

If anything, I think this should be broken out into a separate enum that says what algorithm/shader to use in 3d mode. So you'd have an enum for "slice or volumetric" and if volumetric, there's a different enum for "raymarch or pathtrace".. something like that.  And then raymarch has a flag for whether it's doing max projection or not... kinda sorta maybe

PR #167, opened 2023-11-21
Fix error when cancelling zarr requests

toloudis:
The fix here is a vendored (unreleased) version of the zarr.js library from a pull request that contains the fix.  We should remove this file if the fix ever lands in a published npm package.

I added a readme to explain the provenance of this extra file.

This PR addresses the issue mentioned here:
https://github.com/allen-cell-animated/volume-viewer/pull/163#issuecomment-1809157741


frasercl (approved):
As long as we don't depend on doing this for too long

frasercl:
I think I may have a more sustainable solution in #168, if you'd like to take a look

toloudis:
closing in favor of #168

PR #168, opened 2023-11-22
Use zarrita

frasercl:
Zarrita.js is a slightly more modular zarr library which works in a similar way to zarr.js, but fixes a number of problems we've been facing with the current library:

- Most immediately relevant: errors caused by cancelled requests are properly caught (this seems to have been caused by zarr.js's use of p-queue for limiting concurrent requests; zarrita uses a dummy queue by default). This should remove the need for the hacky fix in #167
- It allows us to pass down arbitrary options to its version of `Store` (called `Readable`), which solves a huge problem we've been facing with tracking prefetch requests

This PR switches us from zarr to zarrita. Remarkably little actually changes about how we interact with the library. The biggest change is converting the store wrapper from an instance of `Store`, which requires lots of methods that look like interacting with a filesystem, to an instance of `Readable`, which requires only one method: `get`.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -186,7 +186,7 @@ function pickLevelToLoad(loadSpec: LoadSpec, spatialDimsZYX: [number, number, nu
   return Math.max(optimalLevel, loadSpec.multiscaleLevel ?? 0);
 }
 
-function convertChannel(channelData: zarr.TypedArray<"uint8" | "uint16">): Uint8Array {
+function convertChannel(channelData: zarr.TypedArray<zarr.NumberDataType>): Uint8Array {
   if (channelData instanceof Uint8Array) {
     return channelData as Uint8Array;
   }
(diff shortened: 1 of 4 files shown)


src/loaders/OmeZarrLoader.ts
```diff
@@ -213,8 +184,8 @@ function pickLevelToLoad(loadSpec: LoadSpec, spatialDimsZYX: [number, number, nu
   return Math.max(optimalLevel, loadSpec.multiscaleLevel ?? 0);
 }
 
-function convertChannel(channelData: TypedArray, dtype: string): Uint8Array {
-  if (dtype === "|u1") {
+function convertChannel(channelData: zarr.TypedArray<"uint8" | "uint16">): Uint8Array {
```
toloudis, line 187: let's make this work with more types.

src/loaders/OmeZarrLoader.ts
```diff
@@ -1,8 +1,9 @@
 import { Box3, Vector3 } from "three";
-import { HTTPStore, TypedArray, ZarrArray, openArray, openGroup, slice } from "zarr";
-import { RawArray } from "zarr/types/rawArray";
-import { AsyncStore } from "zarr/types/storage/types";
-import { Slice } from "zarr/types/core/types";
+
+// TODO: fix "could not find a declaration file for module 'zarrita'." It has a .d.ts, so why isn't it being found?
+//   For now, I pinkie swear I'm using its types right
+import * as zarr from "zarrita";
+import { AbsolutePath, AsyncReadable, Readable } from "@zarrita/storage/dist/src/types";
```
toloudis, line 6: if there's any way to make TS happier that would be good

toloudis (approved):

PR #170, opened 2023-11-29
fix how pathtracer receives updates to colors and lut

toloudis:
Pathtracer's main updating function for color and LUT updates was updateActiveChannels.  This function short-circuits based only on changes to the set of enabled channels, so if you didn't enable/disable a channel, then nothing else would update.

I am fixing this by using our settings MATERIAL flag to let the path tracer receive these updates.  Part of the fix is to add the per-channel color to our volume render settings.  Over time, we should add more and more per-channel settings to this structure and more and more will be able to use our updateSettings code path.

### DIFF
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -427,10 +427,12 @@ export default class VolumeDrawable {
 
   updateMaterial(): void {
     this.volumeRendering.updateActiveChannels(this.fusion, this.volume.channels);
+    this.volumeRendering.updateSettings(this.settings, SettingsFlags.MATERIAL);
   }
 
   updateLuts(): void {
     this.volumeRendering.updateActiveChannels(this.fusion, this.volume.channels);
+    this.volumeRendering.updateSettings(this.settings, SettingsFlags.MATERIAL);
   }
 
   setVoxelSize(values: Vector3): void {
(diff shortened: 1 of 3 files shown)


meganrm (approved):
looks good to me

src/VolumeDrawable.ts
```diff
@@ -427,10 +427,12 @@ export default class VolumeDrawable {
 
   updateMaterial(): void {
     this.volumeRendering.updateActiveChannels(this.fusion, this.volume.channels);
+    this.volumeRendering.updateSettings(this.settings, SettingsFlags.MATERIAL);
   }
 
   updateLuts(): void {
     this.volumeRendering.updateActiveChannels(this.fusion, this.volume.channels);
+    this.volumeRendering.updateSettings(this.settings, SettingsFlags.MATERIAL);
   }
```
frasercl, line 436: Bit odd that these are identical, I hadn't noticed that before

toloudis:
I forgot to say, this bug came about when we first implemented the updateSettings code path, and the path tracer was ignored / under-tested.

frasercl (approved):
Clearly other per-channel settings have already been in `VolumeRenderSettings` (`specular`, `emissive`, etc.), but I was kind of imagining that long term we'd have a separate settings structure and method to update a settings object for a single channel? Something like `updateChannelSettings(index: number, settings: ChannelRenderSettings): void { ... }`. This is a fine change for now, but worth thinking about for any future cleanup/clarification work we do here.

PR #171, opened 2023-12-05
add a raw array volume loader

### LINKED ISSUE #133: Create new loader for raw data
## Use Case

- As a developer, I'd like to easily create `Volume`s as mocks to unit test both `Volume` and the loader API.
- As a developer, I'd like to use the same API to load raw data arrays (in nbvv, e.g.) as remote, formatted data.

## Solution

Add a new raw data loader which implements `IVolumeLoader`. We should not have to modify `LoadSpec` to accomplish this. The loader should receive a `TypedArray` (or an array of `TypedArrays` representing one channel, or an array of arrays representing time points and channels) when `createVolume` is called, and hold on to it for when `loadVolumeData` is called.

Write good unit tests by using this new loader to create mocks.

Future work: update website-3d-cell-viewer, nbvv to load raw data via this path.

toloudis:
resolves #133 

This allows the "raw array" volume data to be loaded in the same way that files are loaded.  This code path is used by `nbvv` to pass numpy arrays directly into the viewer from jupyter.  The new types RawArrayData and RawArrayInfo introduced here will be exported by website-3d-cell-viewer and used as props to that component, and passed through to this loader.

To test/demo: select the "procedural" volume from the dropdown in the test app.  It should show a 3 channel volume with one torus, one sphere and one cone.


src/loaders/RawArrayLoader.ts
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    const imageInfo = vol.imageInfo;
+    const cache = this.cache;
+    for (let chindex = 0; chindex < imageInfo.numChannels; ++chindex) {
+      if (requestedChannels && requestedChannels.length > 0 && !requestedChannels.includes(chindex)) {
+        continue;
+      }
+      // TODO this loader is its own cache and need not even use the cache?
+      const cacheResult = cache?.get(`${imageInfo.name}/${chindex}`);
```
ShrimpCryptid, line 129: It seems like this would cause the memory footprint of this loader to double if it was loading from a URL resource? Or would it just ignore it if the provided URL isn't a real URL...?

src/loaders/RawArrayLoader.ts
```diff
@@ -0,0 +1,143 @@
...
+      if (cacheResult) {
+        vol.setChannelDataFromVolume(chindex, new Uint8Array(cacheResult));
+        onChannelLoaded?.(vol, chindex);
+      } else {
+        const volsize = this.data.shape[3] * this.data.shape[2] * this.data.shape[1]; // x*y*z pixels * 1 byte/pixel
+        vol.setChannelDataFromVolume(chindex, new Uint8Array(this.data.buffer.buffer, chindex * volsize, volsize));
+        cache?.insert(`${imageInfo.name}/${chindex}`, vol.channels[chindex].volumeData.buffer);
+        onChannelLoaded?.(vol, chindex);
```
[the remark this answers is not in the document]
toloudis: no, line 135 copies a subset of the data buffer into the channel.  the data buffer is all channels concatenated into one big array

toloudis:
> I'm noticing a bug I think is unrelated to your changes?
> 
> 1. Switch to OME ZARR (IDR 1)
> 2. Quickly switch to another loader (ex, Procedural)
> 

This is a good bug that we have to fix!   We have to cancel all outstanding requests when a new loader/volume is added

src/loaders/RawArrayLoader.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.data = rawData;
+    // check consistent dims
+    if (
+      this.data.shape[0] !== this.jsonInfo.sizeC ||
+      this.data.shape[1] !== this.jsonInfo.sizeZ ||
+      this.data.shape[2] !== this.jsonInfo.sizeY ||
+      this.data.shape[3] !== this.jsonInfo.sizeX
+    ) {
+      throw new Error("RawArrayLoader: data shape does not match metadata");
+    }
```
meganrm, line 83: could this ever happen in the process of switching data? Or does the metadata always get updated first?

src/loaders/RawArrayLoader.ts
```diff
@@ -0,0 +1,143 @@
...
+  },
+
+  userData: json.userData,
+});
+
+class RawArrayLoader implements IVolumeLoader {
+  // one array per channel
+  data: RawArrayData;
+  jsonInfo: RawArrayInfo;
+
```
meganrm, line 69: nit: extra space here?

src/loaders/index.ts
```diff
@@ -2,20 +2,26 @@ import { IVolumeLoader } from "./IVolumeLoader";
...
 
+// superset of all necessary loader options
 export type CreateLoaderOptions = {
   fileType?: VolumeFileFormat;
   cache?: VolumeCache;
   scene?: number;
   concurrencyLimit?: number;
+  // optional raw image data for VolumeFileFormat.DATA
+  imageData?: RawArrayData;
+  imageDataInfo?: RawArrayInfo;
```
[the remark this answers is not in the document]
meganrm: I know it could be overkill, but it just looks weird to have optional things that then throw an error if they don't exist. but I get that they're only required in that one case

ShrimpCryptid (approved):
Tested locally and everything seems to be working, outside of the bug that I described. Thanks for answering my questions!

src/loaders/index.ts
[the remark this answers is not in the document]
frasercl: I wrote the first version of this file and did initially try to write this type as a union of option objects for each `fileType`, but gave up (motivated in part by some consulation with Dan) after realizing that all the options were optional anyways. But I definitely agree that this type shouldn't allow you to produce errors by not providing these options when they're needed - perhaps by unioning one options object for `VolumeFileFormat.DATA` and one for `Exclude<VolumeFileFormat, VolumeFileFormat.DATA>`.

In fact, should `createVolumeLoader` let you create a data loader in the first place? Basically the primary convenience it provides is accepting a URL to a data source of undetermined format and working out which kind of loader to hand back. I can't think of a likely scenario where a client might not know whether a volume is coming out of a local array or a remote source, and having to pass in an unused `path` argument is a bit clumsy anyways.

src/loaders/RawArrayLoader.ts
frasercl: Agree with the TODO that this loader probably doesn't need access to the cache. Though I could see wanting to let the cache count it towards its maximum size in order to keep memory use under control? In that case, though, I'd just insert the whole contents of `data` as one entry - it doesn't seem all that worth it to avoid the work of splitting the array into different channels.

src/loaders/index.ts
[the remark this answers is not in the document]
toloudis: Do you want a `CreateLoaderOptionsFile` and `CreateLoaderOptionsRawArray`?  I could try to do that.  What would satisfy this PR comment?

public/index.ts
```diff
@@ -967,40 +969,49 @@ function goToZSlice(slice: number): boolean {
...
+  const result = new Uint8Array(totalLength);
+  let offset = 0;
+  for (const arr of arrays) {
+    result.set(arr, offset);
+    offset += arr.length;
+  }
+  return result;
+}
+
 function createTestVolume() {
```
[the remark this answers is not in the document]
toloudis: sure - why not?  The test viewer code has been allowed to be more sloppy than the actual library, but it doesn't need to be.

toloudis:
putting this PR in draft mode as it needs reworking due to lots of volume loader changes.

toloudis:
This PR is re-done to be up-to-date with the current loader api.

src/loaders/RawArrayLoader.ts
```diff
@@ -0,0 +1,138 @@
...
+    translation: new Vector3(0, 0, 0),
+    rotation: new Vector3(0, 0, 0),
+  },
+
+  userData: json.userData,
+});
+
+class RawArrayLoader extends ThreadableVolumeLoader {
+  // one array per channel
+  data: RawArrayData;
```
ShrimpCryptid, line 78: Is this comment still correct? It seems like `RawArrayData` has one array that merges all of the data across channels, not sure if I'm reading that incorrectly.

src/loaders/RawArrayLoader.ts
```diff
@@ -0,0 +1,138 @@
...
+    loadSpec: LoadSpec,
+    onData: RawChannelDataCallback
+  ): Promise<{ loadSpec?: LoadSpec }> {
+    const requestedChannels = loadSpec.channels;
+
+    for (let chindex = 0; chindex < imageInfo.numChannels; ++chindex) {
+      if (requestedChannels && requestedChannels.length > 0 && !requestedChannels.includes(chindex)) {
+        continue;
+      }
+      const volsize = this.data.shape[3] * this.data.shape[2] * this.data.shape[1]; // x*y*z pixels * 1 byte/pixel
```
ShrimpCryptid, line 122: very tiny nit: camelCase? Also, could include units in variable name (`volSizeBytes`)

public/index.ts
```diff
@@ -976,40 +980,49 @@ function goToZSlice(slice: number): boolean {
...
-    volumeData: channelVolumes,
+    metadata: imgData,
+    data: {
+      // expected to be "uint8" always
+      dtype: "uint8",
+      // [c,z,y,x]
+      shape: [channelVolumes.length, sizeZ, sizeY, sizeX],
+      // the bits (assumed uint8!!)
+      buffer: new DataView(alldata.buffer),
+    } as RawArrayData,
```
ShrimpCryptid, line 1025: Do you need the type coercion here for `RawArrayData`? I would expect Typescript to infer the type because of the function's return type; if `RawArrayData` ever changes, the mismatch would not get highlighted here.

ShrimpCryptid (approved):
LGTM! Couple small comments.

meganrm (approved):

PR #172, opened 2023-12-07
experimenting with the prefetch branch

toloudis:
some various tweaks while testing/experimenting @frasercl

### DIFF
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -926,7 +926,7 @@ export class View3d {
     Object.assign(pane.element.style, paneStyle);
 
     const cacheSettings = {
-      concurrency: 2,
+      concurrency: 4,
       size: 1000000,
     };
     const caching = pane.addFolder({ title: "Cache" });
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -925,6 +925,20 @@ export class View3d {
     };
     Object.assign(pane.element.style, paneStyle);
 
+    const cacheSettings = {
+      concurrency: 2,
+      size: 1000000,
+    };
+    const caching = pane.addFolder({ title: "Cache" });
+    caching.addInput(cacheSettings, "concurrency").on("change", ({ value }) => {
+      // TODO could/should we treat the queue as more of a global object?
+      this.image?.volume.loader?.setConcurrencyLimit(value);
+    });
+    caching.addInput(cacheSettings, "size").on("change", ({ value }) => {
+      // TODO could/should we treat the cache as more of a global object?
+      this.image?.volume.loader?.setCacheSize(value);
+    });
+
     // LIGHTS
     const lights = pane.addFolder({ title: "Lights (isosurface)" });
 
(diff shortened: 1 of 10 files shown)


ShrimpCryptid (approved):
Changes all seem reasonable! Opened it up to play with it a bit and it's nice having the debug messages for context.

src/VolumeCache.ts
```diff
@@ -100,6 +100,7 @@ export default class VolumeCache {
       return;
     }
 
+    console.log("cache eviction " + this.last.key);
```
meganrm, line 103: is this left over from debugging or do you need it here?

src/loaders/OmeZarrLoader.ts
```diff
@@ -519,7 +554,7 @@ class OMEZarrLoader implements IVolumeLoader {
         // For nearby timesteps: duplicate only the load requests made on the current timestep on this one too
         for (const coord of chunkCoords) {
           const [c, z, y, x] = coord.slice(1);
-          this.prefetchChunk(scaleLevel.path, [t, c, z, y, x], subscriber);
+          this.prefetchChunk(scaleLevel.path, [t, c, z, y, x], subscriber, Math.abs(t - activeTimestep) * 2000);
```
meganrm, line 557: just for readably: could define this as `const delayMs = Math.abs(t - activeTimestep) * 2000`. If I was looking at this in VScode I would be able to see the the inputs to `prefetchChunk`, but here it's hard to tell what this argument is

src/View3d.ts
```diff
@@ -925,6 +925,20 @@ export class View3d {
...
+    const cacheSettings = {
+      concurrency: 4,
+      size: 1000000,
+    };
+    const caching = pane.addFolder({ title: "Cache" });
+    caching.addInput(cacheSettings, "concurrency").on("change", ({ value }) => {
+      // TODO could/should we treat the queue as more of a global object?
+      this.image?.volume.loader?.setConcurrencyLimit(value);
+    });
+    caching.addInput(cacheSettings, "size").on("change", ({ value }) => {
```
meganrm, line 937: do these only change when there is a new image or are there other conditions where these settings would change?

src/loaders/OmeZarrLoader.ts
toloudis: agree.. probably will make the 2000 a named constant too.  For now this is just being tweaked to see what delay values work well for what kinds of data sizes.

src/View3d.ts
toloudis: Adding them here is just to make them tweakable during testing/debugging to help find optimal settings.   A regular user would not have a reason to change these, typically.
  
IMO the cache size and the max fetch concurrency should be thought of as global values for the whole viewer instance, but for now they are fairly tightly associated with a IVolumeLoader and they really only matter for zarr data.

toloudis:
This branch will stay alive for a while for experimentation but is a candidate to be closed within a month or so

Issue #158, opened 2023-10-26
Don't load disabled channels


frasercl:
## Use Case

Currently, we always load all channels any time we need new volume data, despite the fact that lots of use cases keep most channels of a volume disabled. (e.g. viewing the Variance dataset in CFE, whose segmentation and contour channels - a majority of total channels - are disabled by default.)

## Solution

We have a way to tell loaders which channels to load and a place to trigger new loads if we need new data. All we need to do is use them for channels as well!

As a bonus optimization, if the only change in required data is channels going from disabled to enabled, we should only load those channels, since we won't want any more data in any other channel. (This could get subtle with subregions - e.g. if enabled channels change _and the subregion also shrinks_, but the shrinkage won't result in a change of scale level per #157, load only the newly-enabled channels _at the old subregion_ to avoid dimension mismatches between old and new channels.)

PR #173, opened 2024-01-08
Only load enabled channels

frasercl:
Addresses #158: don't load disabled channels.

### To verify:
1. Open test app.
2. Switch volume to "OME Zarr (Nucmorph)"
3. Disable either channel.
4. Step through time. *Each time point should load faster than normal.*
5. Re-enable the disabled channel. *The channel should load into its current timepoint.*

### What this PR does NOT do:
This is a relatively small change to address the "letter" of the linked issue, which falls just a bit short of the "spirit" of the issue for the following reasons:

- Channels are initialized as enabled and only disabled after loading has begun, so this feature gives no benefit to the initial load of a volume. As a consequence, because the test viewer does not have a JSON volume that includes both multiple time points and enough channels to require more than one atlas image, I had trouble testing how the JSON atlas loader behaves when loading a subset of channels.
- Though I spent a long time trying to clean up how channel enabled/disabled state gets updated to fix a bug, which turned out to be caused by something unrelated and silly, I didn't manage to find a change that I could reasonably earmark onto this PR. Until channel state is a bit more reliable, this feature may be a bit fiddly, though the point above is the only behavior I've seen that I'd describe as a "bug."
- This PR doesn't include the optimization of only loading newly-enabled channels when possible. I'd rather address the subtleties that come with that optimization (described in the linked issue) when state updates are a bit more coherent. (Caching may make the benefits of this optimization minimal anyways.)

### DIFF
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -497,15 +497,11 @@ export default class VolumeDrawable {
     // flip the color to the "null" value
     this.fusion[channelIndex].rgbColor = enabled ? this.channelColors[channelIndex] : 0;
     // if all are nulled out, then hide the volume element from the scene.
-    if (this.fusion.every((elem) => elem.rgbColor === 0)) {
-      this.settings.visible = false;
-    } else {
-      this.settings.visible = true;
-    }
+    this.settings.visible = !this.fusion.every((elem) => elem.rgbColor === 0);
     this.volumeRendering.updateSettings(this.settings, SettingsFlags.VIEW);
 
     // add or remove this channel from the list of required channels to load
-    const { channels } = this.volume.loadSpec;
+    const { channels } = this.volume.loadSpecRequired;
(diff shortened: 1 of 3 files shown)


src/VolumeDrawable.ts
```diff
@@ -513,6 +518,7 @@ export default class VolumeDrawable {
   // Set the color for a channel
   // @param {Array.<number>} colorrgb [r,g,b]
   updateChannelColor(channelIndex: number, colorrgb: [number, number, number]): void {
+    console.log("update color", channelIndex, colorrgb);
```
ShrimpCryptid, line 521: Was this intentionally left in?

ShrimpCryptid (approved):
LGTM!

toloudis (approved):

src/VolumeDrawable.ts
toloudis: Yup looks like this could be removed

PR #176, opened 2024-01-12
Zarr prefetching

### LINKED ISSUE #126: Prefetch and cache slices

frasercl:
Adds configurable prefetching behavior to `OmeZarrLoader`. Should (at last!) resolve #126. I think I have this feature working pretty well now, but please pull it and try for yourself!

### How it works for users:

- When any zarr volume with a time series loads, the loader begins prefetching adjacent time steps. Moving to adjacent time steps gets faster because those steps may already have been requested and received.
- When the user switches to z-slice mode and the chunking of the volume affords loading a subset of the volume at higher resolution, the loader begins prefetching adjacent chunks in z. Moving to adjacent z-slices gets faster too.
- If both of the above are true, it does both at the same time!

### How it works for developers:

- `RequestQueue` now has two "lanes:" regular and low-priority.
  - The zarr loader uses the regular lane for normal loads and the low-priority lane for prefetches.
  - All regular requests must be started before any low-priority requests get to go.
  - The low-priority lane has its own concurrency limit, which may be set lower (but cannot be set higher!) than the limit for the regular lane. This allows the user to ensure that the queue always leaves some space open for new regular requests to begin immediately.
- `OmeZarrLoader` now properly uses `SubscribableRequestQueue`, which has new method arguments to use `RequestQueue`'s new priority features.
- `OmeZarrLoader` has two new options to configure prefetching behavior:
  - `maxPrefetchDistance` is a `Vector4` representing the maximum number of chunks to prefetch outwards in both directions of a given dimension. (X and Y are of course not currently used in practice, but they're still in there for just in case.) For instance, if a load requests chunks with z coords 3 and 4 and `maxPrefetchDistance.z` is 2, the loader will prefetch similar chunks with z coords 1, 2, 5, and 6. The `w` component is time.
  - `maxPrefetchChunks` is the max number of total prefetch requests that can be made after any load. The loader will always stop issuing prefetch requests at this number even if `maxPrefetchDistance` would allow loading more.

### Other changes:

- OmeZarrLoader.ts was getting long and cumbersome, and began with a growing string of items that made the actual beginning of the `OmeZarrLoader` class declaration hard to find. I moved some of these items into their own files in a new /zarr_utils subdirectory:
  - WrappedStore.ts for the object that lets us link into zarrita's low-level chunk request behavior.
  - types.ts for stray shared types and the big declaration of the OME-Zarr JSON metadata format.
- /zarr_utils also has one brand new item: `ChunkPrefetchIterator` encapsulates the somewhat complicated process of identifying which chunks can and should be prefetched and in what order.
- Did a bit of refactoring in `OmeZarrLoader` to make all of the above easier, notably adding a couple methods for convenient reshuffling of dimension-ordered arrays.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -127,12 +127,28 @@ class OMEZarrLoader implements IVolumeLoader {
 
   static async createLoader(
     url: string,
+    scene?: number,
+    cache?: VolumeCache,
+    queue?: SubscribableRequestQueue
+  ): Promise<OMEZarrLoader>;
+  static async createLoader(
+    url: string,
+    scene?: number,
+    cache?: VolumeCache,
+    concurrencyLimit?: number,
+    prefetchConcurrencyLimit?: number
+  ): Promise<OMEZarrLoader>;
+  static async createLoader(
+    url: string,
     scene = 0,
     cache?: VolumeCache,
-    concurrencyLimit = 10
+    queue?: SubscribableRequestQueue | number,
+    prefetchConcurrencyLimit?: number
   ): Promise<OMEZarrLoader> {
-    // Setup: create queue and store, get basic metadata
-    const queue = new SubscribableRequestQueue(concurrencyLimit, 5);
+    // Setup queue and store, get basic metadata
+    if (typeof queue === "number" || queue === undefined) {
+      queue = new SubscribableRequestQueue(queue, prefetchConcurrencyLimit);
+    }
     const store = new WrappedStore<RequestInit>(new FetchStore(url), cache, queue);
     const root = zarr.root(store);
     const group = await zarr.open(root, { kind: "group" });
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -22,6 +22,8 @@ import {
 
 const CHUNK_REQUEST_CANCEL_REASON = "chunk request cancelled";
 
+const PREFETCH_TIME_MARGIN = 8;
+
 type TCZYX<T> = [T, T, T, T, T];
 type SubscriberId = ReturnType<SubscribableRequestQueue["addSubscriber"]>;
 
@@ -96,6 +98,7 @@ const getDimensionCount = ([t, c, z]: TCZYX<number>) => 2 + Number(t > -1) + Num
 type WrappedStoreOpts<Opts> = {
   options?: Opts;
   subscriber: SubscriberId;
+  isPrefetch?: boolean;
 };
 
 /**
@@ -125,6 +128,9 @@ class WrappedStore<Opts, S extends Readable<Opts> = Readable<Opts>> implements A
     const ZARR_EXTS = [".zarray", ".zgroup", ".zattrs", "zarr.json"];
     if (!this.cache || ZARR_EXTS.some((s) => key.endsWith(s))) {
       return this.baseStore.get(key, opts?.options);
+    }
+    if (opts?.isPrefetch) {
+      console.log("prefetch: ", key);
     }
 
     let keyPrefix = (this.baseStore as FetchStore).url ?? "";
@@ -220,15 +226,16 @@ function convertChannel(channelData: zarr.TypedArray<zarr.NumberDataType>): Uint
 type NumericZarrArray = zarr.Array<zarr.NumberDataType, WrappedStore<RequestInit>>;
 
 class OMEZarrLoader implements IVolumeLoader {
+  /** Hold one optional subscriber ID per timestep, each defined iff a batch of prefetches is waiting for that frame */
+  private prefetchSubscribers: (SubscriberId | undefined)[];
+  /** The ID of the subscriber responsible for "actual loads" (non-prefetch requests) */
+  private loadSubscriber: SubscriberId | undefined;
+
   // TODO: this property should definitely be owned by `Volume` if this loader is ever used by multiple volumes.
   //   This may cause errors or incorrect results otherwise!
   private maxExtent?: Box3;
 
-  private prefetchSubscribers: (SubscriberId | undefined)[];
-  private loadSubscriber: SubscriberId | undefined;
-
   private constructor(
-    private store: WrappedStore<RequestInit>,
     private scaleLevels: NumericZarrArray[],
     private multiscaleMetadata: OMEMultiscale,
     private omeroMetadata: OmeroTransitionalMetadata,
@@ -264,9 +271,10 @@ class OMEZarrLoader implements IVolumeLoader {
     // Open all scale levels of multiscale
     const scaleLevelPromises = multiscale.datasets.map(({ path }) => zarr.open(root.resolve(path), { kind: "array" }));
     const scaleLevels = await Promise.all(scaleLevelPromises);
+    scaleLevels.forEach((level, i) => console.log(i, level.shape, level.chunks));
     const axisTCZYX = remapAxesToTCZYX(multiscale.axes);
 
-    return new OMEZarrLoader(store, scaleLevels as NumericZarrArray[], multiscale, metadata.omero, axisTCZYX, queue);
+    return new OMEZarrLoader(scaleLevels as NumericZarrArray[], multiscale, metadata.omero, axisTCZYX, queue);
   }
 
   private getUnitSymbols(): [string, string] {
@@ -442,6 +450,71 @@ class OMEZarrLoader implements IVolumeLoader {
     vol.channelLoadCallback = onChannelLoaded;
     vol.imageMetadata = buildDefaultMetadata(imgdata);
     return vol;
+  }
+
+  /** Reads a list of chunk keys requested by a `loadVolumeData` call and sets up appropriate prefetch requests. */
+  private beginPrefetch(keys: string[], scaleLevel: NumericZarrArray): void {
+    const numDims = getDimensionCount(this.axesTCZYX);
+
+    // Convert keys to arrays of coords
+    const chunkCoords = keys.map((key) => {
+      const coordsInDimensionOrder = key
+        .trim()
+        .split("/")
+        .slice(-numDims)
+        .filter((s) => s !== "")
+        .map((s) => parseInt(s, 10));
+      return this.orderByTCZYX(coordsInDimensionOrder, 0);
+    });
+
+    // Get number of chunks per dimension in `scaleLevel`
+    const chunkDimsUnordered = scaleLevel.shape.map((dim, idx) => Math.ceil(dim / scaleLevel.chunks[idx]));
+    const chunkDims = this.orderByTCZYX(chunkDimsUnordered, 1);
+
+    // Get all channels involved in this request
+    const channels = new Set<number>();
+    for (const coord of chunkCoords) {
+      channels.add(coord[1]);
+    }
+
+    // Clear out any existing prefetches (requests already in flight will be picked up by new prefetches if useful)
+    for (const subscriber of this.prefetchSubscribers) {
+      if (subscriber !== undefined) {
+        this.requestQueue.removeSubscriber(subscriber, CHUNK_REQUEST_CANCEL_REASON);
+      }
+    }
+
+    // Now let's get to prefetching!
+    // TODO: consider cache size and something like average chunk size when deciding how much to prefetch
+    const activeTimestep = chunkCoords[0][0];
+    const tmin = Math.max(0, activeTimestep - PREFETCH_TIME_MARGIN);
+    const tmax = Math.min(chunkDims[0], activeTimestep + PREFETCH_TIME_MARGIN);
+    for (let t = tmin; t < tmax; t++) {
+      const subscriber = this.requestQueue.addSubscriber();
+      this.prefetchSubscribers[t] = subscriber;
+
+      if (t === activeTimestep) {
+        // For the current timestep: get all chunks in ZYX for all channels involved in this request
+        // TODO: it is dangerous to assume that we can reasonably prefetch the entire volume at the current timestep.
+        //   This behavior shouldn't be allowed into production if we expect anyone to deal with large enough datasets.
+        for (const c of channels) {
+          for (let z = 0; z < chunkDims[2]; z++) {
+            for (let y = 0; y < chunkDims[3]; y++) {
+              for (let x = 0; x < chunkDims[4]; x++) {
+                // This will call the store, which will handle caching and queueing
+                scaleLevel.getChunk(this.orderByDimension([t, c, z, y, x]), { subscriber, isPrefetch: true });
+              }
+            }
+          }
+        }
+      } else {
+        // For nearby timesteps: duplicate only the load requests made on the current timestep on this one too
+        for (const coord of chunkCoords) {
+          const [c, z, y, x] = coord.slice(1);
+          scaleLevel.getChunk(this.orderByDimension([t, c, z, y, x]), { subscriber, isPrefetch: true });
+        }
+      }
+    }
   }
 
   async loadVolumeData(vol: Volume, explicitLoadSpec?: LoadSpec, onChannelLoaded?: PerChannelCallback): Promise<void> {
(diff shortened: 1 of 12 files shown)


src/utils/RequestQueue.ts
```diff
@@ -30,13 +30,24 @@ interface RequestItem<V> {
  * while the original request is still in the queue.
  */
 export default class RequestQueue {
-  /** The maximum number of requests that can be handled concurently.
-    Once reached, additional requests will be queued up to run once a running request completes.*/
+  /**
+   * The maximum number of requests that can be handled concurently.
```
toloudis, line 34:
```suggestion
   * The maximum number of requests that can be handled concurrently.
```

src/utils/RequestQueue.ts
```diff
@@ -30,13 +30,24 @@ interface RequestItem<V> {
...
+   * Once reached, additional requests will be queued up to run once a running request completes.
+   */
   private maxActiveRequests: number;
 
+  /**
+   * The maximum number of requests that can be handled concurrently for low-priority requests to be queued. Equal to
+   * `maxActiveRequests` by default, but may be set lower to ensure the queue leaves space for high-priority requests.
```
toloudis, line 41: seems like the default should always be to allow high priority requests and keep this one more conservative?

src/utils/RequestQueue.ts
```diff
@@ -119,23 +140,31 @@ export default class RequestQueue {
...
+        this.addRequestToQueue(key, lowPriority);
+      }
+    } else {
+      const lowPriorityIndex = this.queueLowPriority.indexOf(key);
+      if (lowPriorityIndex > -1 && !lowPriority) {
+        // This request is registered and queued, but is now being requested with high priority.
+        // Promote it to high priority.
+        this.queueLowPriority.splice(lowPriorityIndex, 1);
         this.addRequestToQueue(key);
+      } else if (delayMs <= 0) {
```
frasercl, line 163: Might be worth thinking more deeply about how "promote to high priority" and "cancel delay and execute immediately" might interact here.

src/loaders/OmeZarrLoader.ts
```diff
@@ -424,10 +354,63 @@ class OMEZarrLoader implements IVolumeLoader {
...
+    // Get number of chunks per dimension in `scaleLevel`
+    const chunkDimsUnordered = scaleLevel.shape.map((dim, idx) => Math.ceil(dim / scaleLevel.chunks[idx]));
+    const chunkDims = this.orderByTCZYX(chunkDimsUnordered, 1);
+
+    const subscriber = this.requestQueue.addSubscriber();
+    // `ChunkPrefetchIterator` yields chunk coordinates in order of roughly how likely they are to be loaded next
+    const prefetchIterator = new ChunkPrefetchIterator(
+      chunkCoords,
+      this.fetchOptions.maxPrefetchDistance,
+      new Vector4(chunkDims[4], chunkDims[3], chunkDims[2], chunkDims[0])
```
frasercl, line 394: Consider: should we use some other type than `Vector4`?

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+      }
+    }
+
+    // Create `PrefetchDirectionState`s for each direction and fill them with chunks at the borders of the fetched set
+    const directions: PrefetchDirectionState[] = extrema.map((start, direction) => {
+      const end =
+        direction & 1
+          ? Math.min(start + maxOffsetArr[direction >> 1], chunkSizeArr[direction >> 1] - 1)
+          : Math.max(start - maxOffsetArr[direction >> 1], 0);
```
frasercl, line 64: clean up and/or document this! it's dense and hard to read this way

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
frasercl, line 1: unit test this!

src/loaders/zarr_utils/types.ts
```diff
@@ -0,0 +1,69 @@
...
+export type TCZYX<T> = [T, T, T, T, T];
+export type SubscriberId = ReturnType<SubscribableRequestQueue["addSubscriber"]>;
+
+export type OMECoordinateTransformation =
+  | {
+      type: "identity";
+    }
+  | {
+      type: "translation";
+      translation: number[];
```
ShrimpCryptid, line 12: Do these arrays have specific lengths? (ZYX?) Not sure if it's useful to capture in the type, just thinking.

src/utils/SubscribableRequestQueue.ts
```diff
@@ -27,8 +27,14 @@ export default class SubscribableRequestQueue {
...
-  constructor(maxActiveRequests?: number) {
-    this.queue = new RequestQueue(maxActiveRequests);
+  constructor(maxActiveRequests?: number, maxLowPriorityRequests?: number);
+  constructor(inner: RequestQueue);
+  constructor(maxActiveRequests?: number | RequestQueue, maxLowPriorityRequests?: number) {
+    if (typeof maxActiveRequests === "number" || maxActiveRequests === undefined) {
+      this.queue = new RequestQueue(maxActiveRequests, maxLowPriorityRequests);
+    } else {
+      this.queue = maxActiveRequests;
+    }
```
ShrimpCryptid, line 37: First time I've seen an overload like this, interesting. Maybe useful to add a docstring comment to explain that it can take in an existing RequestQueue as a parameter

src/loaders/zarr_utils/types.ts
frasercl: OME-Zarr technically allows any number of dimensions in any order. (Many of the hoops this loader has to jump through have to do with converting arbitrary dimension orderings from metadata into something consistent - see `orderByDimension` and `orderByTCZYX` below. I mostly try to type that converted, consistent representation as a fixed-length array.)

src/test/RequestQueue.test.ts
```diff
@@ -218,12 +218,79 @@ describe("test RequestQueue", () => {
         { key: "b", requestAction: work },
       ];
       const start = Date.now();
-      const promises = rq.addRequests(requests, delayMs);
+      const promises = rq.addRequests(requests, false, delayMs);
```
ShrimpCryptid, line 221: Nit: Seeing this makes me think it would be nice to have an enum for the request type (HIGH or LOW), since there's no context for it when seeing it without Intellisense... Very likely unnecessarily verbose though.

src/test/RequestQueue.test.ts
```diff
@@ -218,12 +218,79 @@ describe("test RequestQueue", () => {
...
+      const prom1 = rq.addRequest("a", work);
+      const prom2 = rq.addRequest("b", work, true);
+      expect(rq.requestRunning("a")).to.be.true;
+      expect(rq.requestRunning("b")).to.be.false;
+      rq.addRequest("b", work);
+      expect(rq.requestRunning("b")).to.be.true;
+      await prom1;
+      await prom2;
+      expect(count).to.equal(2);
+    });
```
ShrimpCryptid, line 293: I know the current behavior for RequestQueue requires all high priority tasks to be complete before moving onto the low priority tasks... Would this ever cause problematic stalling?

Like, say a high priority task is stuck due to a network timeout or something. This would prevent all low priority tasks for being executed, even if there is capacity available to handle them.

ShrimpCryptid: Okay, correction. Took a closer look at RequestQueue, and it allows low priority requests to be started once the total number of active requests dips below the max low priority count and if the high priority queue is empty. Question answered 👍

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -0,0 +1,119 @@
...
+  directions: PrefetchDirectionState[];
+
+  constructor(chunks: TCZYX<number>[], tzyxMaxOffset: TZYX, tzyxChunkSize: TZYX) {
+    // Get max and min chunk coordinates for T/Z/Y/X
+    const extrema = [Infinity, -Infinity, Infinity, -Infinity, Infinity, -Infinity, Infinity, -Infinity];
+
+    for (const chunk of chunks) {
+      for (let j = 0; j < 4; j++) {
+        const val = chunk[skipC(j)];
+        const extremaIdx = j << 1;
```
ShrimpCryptid, line 54: Nit: More readable to as `j * 2`?

src/loaders/OmeZarrLoader.ts
```diff
@@ -215,44 +103,62 @@ function convertChannel(channelData: zarr.TypedArray<zarr.NumberDataType>): Uint
...
+  /** The max. number of total chunks that can be prefetched after any load. */
+  maxPrefetchChunks: number;
+};
 
-class OMEZarrLoader implements IVolumeLoader {
-  scaleLevels: NumericZarrArray[];
-  multiscaleMetadata: OMEMultiscale;
-  omeroMetadata: OmeroTransitionalMetadata;
-  axesTCZYX: [number, number, number, number, number];
+const DEFAULT_FETCH_OPTIONS = {
```
ShrimpCryptid, line 126:
```suggestion
const DEFAULT_FETCH_OPTIONS: ZarrLoaderFetchOptions = {
```

ShrimpCryptid (approved):
LGTM!

src/loaders/zarr_utils/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  unit?: string;
+};
+
+export type OMEDataset = {
+  path: string;
+  coordinateTransformations?: OMECoordinateTransformation[];
+};
+
+// https://ngff.openmicroscopy.org/latest/#multiscale-md
+export type OMEMultiscale = {
```
toloudis, line 35: Great job on making these types.  Kinda surprising that there isn't a thirdparty js lib that captures all these :). Also it's unfortunate that the ome spec is a good deal more general than the sorts of data we hope to present in this viewer. (example - axis names can be just about anything)

src/loaders/zarr_utils/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  id: number;
+  name: string;
+  version: string;
+  channels: {
+    active: boolean;
+    coefficient: number;
+    color: string;
+    family: string;
+    inverted: boolean;
+    label: string;
```
toloudis, line 56: Sadly, this "transitional" data structure is also the only official vehicle for getting channel names in.

src/test/ChunkPrefetchIterator.test.ts
```diff
@@ -0,0 +1,164 @@
+import { expect } from "chai";
+
+import { TCZYX } from "../loaders/zarr_utils/types";
+import ChunkPrefetchIterator from "../loaders/zarr_utils/ChunkPrefetchIterator";
+
+const EXPECTED_3X3X3X3 = [
```
toloudis, line 6: it's hard to understand why this is called 3x3x3 when it has all of t,z,y,x and the arrays are 5 elements each.

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
toloudis: agree

toloudis: Nit: I get what this is doing, but this code is a bit hard to parse when reading it.   I feel like it might be even more readable if the loop were unrolled and/or the extrema array were more structured, since the dims are always t,z,y,x.   It's a 4D bounding box!  The skipC function is hard to understand too.
pseudocode : 
for each chunk:  
   update min/max of T value
   update min/max of Z value
   update min/max of Y value
   update min/max of X value

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Since the user is most likely to want nearby data (in space or time) first, we should prefetch those chunks first.
+ *
+ * Given a list of just-loaded chunks and some bounds, `ChunkPrefetchIterator` iterates evenly outwards in T/Z/Y/X.
+ */
+// NOTE: Assumes `chunks` form a rectangular prism! Will create gaps otherwise! (in practice they always should)
+export default class ChunkPrefetchIterator {
+  directions: PrefetchDirectionState[];
+
+  constructor(chunks: TCZYX<number>[], tzyxMaxOffset: TZYX, tzyxChunkSize: TZYX) {
```
toloudis, line 47: is tzyxMaxOffset just the highest index for chunks in those dims?

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -0,0 +1,119 @@
...
+      this.directions = this.directions.filter((dir) => {
+        if (dir.direction & 1) {
+          return dir.start + offset <= dir.end;
+        } else {
+          return dir.start - offset >= dir.end;
+        }
+      });
+
+      // Yield chunks one chunk farther out in every remaining direction
+      for (const dir of this.directions) {
```
toloudis, line 108: to make sure I understand, is there currently an implied preferential order in T,Z,Y,X ?

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
frasercl: `tzyxChunkSize` is what you're describing; `tzyxMaxOffset` is the maximum number of chunks to move outward in both directions, so e.g. if you're at T = 30 and `tzyxMaxOffset` in T is 2, we prefetch T = 28, 29, 31, 32. (These are both bad names and I'm open to suggestions. And they ought to be documented better regardless!)

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
frasercl: Yes

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        if (val > extrema[extremaIdx + 1]) {
+          extrema[extremaIdx + 1] = val;
+        }
+      }
+    }
+
+    // Create `PrefetchDirectionState`s for each direction
+    const directions: PrefetchDirectionState[] = extrema.map((start, direction) => {
```
toloudis, line 67: Just writing this to help myself understand.  So the directions are either toward the max or toward the min for each of 4 dims. In my head I am turning the bit operations into little functions.  (`isPositive(direction)`, `getAxisDimension(direction)` )

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
toloudis: I think something like `tzyxMaxChunkIndexOffset` is slightly more explicit

toloudis: or tzyxPrefetchOffsets

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
ShrimpCryptid: Now that you've said it I'd love for these to be functions instead of bit operations 🥺

src/loaders/OmeZarrLoader.ts
```diff
@@ -215,44 +103,62 @@ function convertChannel(channelData: zarr.TypedArray<zarr.NumberDataType>): Uint
...
+  maxPrefetchDistance: [5, 5, 5, 5] as [number, number, number, number],
+  maxPrefetchChunks: 30,
```
toloudis, line 128: Idea/ Future suggestion: make this match the defaults explicitly passed in by the test app, but allow them to be tweaked at runtime using the advanced settings gui

toloudis (approved):
No show stoppers for this PR, just some suggestions for clarification/readability.

PR #177, opened 2024-01-18
Feature/loader worker

frasercl:
Adds the ability to offload most volume loaders to a separate persistent worker, to avoid blocking updates and rendering on the "main thread."

This PR is a child of #176 and includes all its commits. Review that one first! I'll keep this one as a draft until that one merges.

### How it works:
Currently, the easiest and most generic way to create a loader is to call `createVolumeLoader`, which accepts a path and an options object and returns a loader of the appropriate type. The procedure to create a loader on a separate worker is quite similar, but includes a couple extra steps:
1. Create a new `LoadWorker`. This accepts arguments for initializing a cache and queue. Internally, `LoadWorker` spins up a new webworker and tells it to get a cache and queue going.
2. Before using the `LoadWorker`, await `LoadWorker.onOpen()`, which returns a promise that resolves when the above initialization step completes and the worker reports that it's ready.
3. Call `LoadWorker.createLoader`. This method has almost identical arguments as `createVolumeLoader` (it just doesn't accept a cache and queue, since loaders on the worker will use the worker's shared cache and queue instances) and returns a `WorkerLoader` which acts as a handle to the loader on the worker and can go wherever a normal `IVolumeLoader` can.
4. More loaders can be created via this method, but currently _only one loader can run on the worker at a time_, so any previous `WorkerLoader` will be invalidated and begin throwing errors when any of its methods are called. (There's no technical reason we can't run multiple loaders on the same worker, in fact we'd get some benefits by automatically sharing cache and queue, we're just not yet likely enough to use that functionality in practice to justify the effort of implementing it.)

### Worker implementation keyfiles
- src/workers/LoadWorkerHandle.ts includes most of the important stuff: `LoadWorker`, `WorkerLoader`, and all of the associated message tracking and bookkeeping.
- src/workers/types.ts rigorously lays out the contract for what messages to/from the loader look like
- src/workers/VolumeLoadWorker.ts contains the worker code, which is not all that interesting: just a map to various message handlers, most of which just call loader methods, and a bit of error handling glue.

### Changes to existing loaders
I had to make some changes to existing volume loaders to support working across the boundary between the main thread and the worker. Most significantly, most loaders are now implemented around a new abstract class called **`ThreadableVolumeLoader`**.

The most important methods of `IVolumeLoader` (the shared interface which all volume loaders implement) want to work directly with `Volume`s: `createVolume` builds and returns a `Volume`, and `loadVolumeData` accepts a `Volume` as an argument and modifies it directly. But transferring full `Volume`s to/from a worker would be impractical for a bunch of reasons. `ThreadableVolumeLoader` is an abstract class which introduces equivalent methods - `createImageInfo` and `loadRawChannelData`, respectively - that work with the `ImageInfo` and `LoadSpec` objects that describe a `Volume`. `LoadWorker` calls these lower-level methods and transfers the resulting changes to `ImageInfo` and/or `LoadSpec` back to the main thread, where direct changes can be made to the relevant `Volume`. `ThreadableVolumeLoader` also provides default implementations of `createVolume` and `loadVolumeData` around these new abstract methods, allowing a loader on the main thread to still use the old API without duplicated code. This has the added benefit of reducing a bit of boilerplate between loaders, since it turns out that the manipulations they were doing to `Volume` were all basically the same.

Additionally, `JsonImageInfoLoader` loaded images by creating an off-screen  tag and waiting for it to load. But since DOM nodes can't be created on a worker, I rewrote this section of the code to use the fetch API.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -9,7 +9,6 @@ import {
   JsonImageInfoLoader,
   View3d,
   Volume,
-  VolumeCache,
   VolumeMaker,
   Light,
   AREA_LIGHT,
@@ -17,7 +16,6 @@ import {
   RENDERMODE_RAYMARCH,
   SKY_LIGHT,
   VolumeFileFormat,
-  createVolumeLoader,
 } from "../src";
 // special loader really just for this demo app but lives with the other loaders
 import { OpenCellLoader } from "../src/loaders/OpenCellLoader";
--- a/public/index.ts
+++ b/public/index.ts
@@ -21,7 +21,7 @@ import {
 import { OpenCellLoader } from "../src/loaders/OpenCellLoader";
 import { State, TestDataSpec } from "./types";
 import { getDefaultImageInfo } from "../src/Volume";
-import LoadWorker from "../src/workers/LoadWorkerHandle";
+import VolumeLoaderContext from "../src/workers/LoadWorkerHandle";
 
 const CACHE_MAX_SIZE = 1_000_000_000;
 const CONCURRENCY_LIMIT = 8;
@@ -86,7 +86,7 @@ const TEST_DATA: Record<string, TestDataSpec> = {
 
 let view3D: View3d;
 
-const loadWorker = new LoadWorker(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
+const loadWorker = new VolumeLoaderContext(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
 
 const myState: State = {
   file: "",
--- a/public/index.ts
+++ b/public/index.ts
@@ -1,5 +1,5 @@
 import "regenerator-runtime/runtime";
-import { Vector2, Vector3, Vector4 } from "three";
+import { Vector2, Vector3 } from "three";
 import * as dat from "dat.gui";
 
 import {
--- a/public/index.ts
+++ b/public/index.ts
@@ -23,6 +23,7 @@ import {
 import { OpenCellLoader } from "../src/loaders/OpenCellLoader";
 import { State, TestDataSpec } from "./types";
 import { getDefaultImageInfo } from "../src/Volume";
+import LoadWorker from "../src/workers/LoadWorkerHandle";
 
 const CACHE_MAX_SIZE = 1_000_000_000;
 const PLAYBACK_INTERVAL = 80;
@@ -83,7 +84,8 @@ const TEST_DATA: Record<string, TestDataSpec> = {
 
 let view3D: View3d;
 
-const volumeCache = new VolumeCache(CACHE_MAX_SIZE);
+// const volumeCache = new VolumeCache(CACHE_MAX_SIZE);
+const loadWorker = new LoadWorker(CACHE_MAX_SIZE);
 
 const myState: State = {
   file: "",
@@ -1005,6 +1007,8 @@ function createTestVolume() {
 }
 
 async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
+  await loadWorker.onOpen();
+  console.log(new Vector2(0, 1));
   if (data.type === "opencell") {
     return new OpenCellLoader();
   }
@@ -1018,7 +1022,7 @@ async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
     }
   }
 
-  return await createVolumeLoader(path, { cache: volumeCache });
+  return await loadWorker.createLoader(path);
 }
 
 async function loadVolume(loadSpec: LoadSpec, loader: IVolumeLoader): Promise<void> {
--- a/public/index.ts
+++ b/public/index.ts
@@ -86,7 +86,7 @@ const TEST_DATA: Record<string, TestDataSpec> = {
 
 let view3D: View3d;
 
-const loadWorker = new VolumeLoaderContext(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
+const loaderContext = new VolumeLoaderContext(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
 
 const myState: State = {
   file: "",
@@ -1012,7 +1012,7 @@ async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
     return new OpenCellLoader();
   }
 
-  await loadWorker.onOpen();
+  await loaderContext.onOpen();
 
   let path: string | string[] = data.url;
   if (data.type === VolumeFileFormat.JSON) {
@@ -1023,7 +1023,7 @@ async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
     }
   }
 
-  return await loadWorker.createLoader(path, {
+  return await loaderContext.createLoader(path, {
     fetchOptions: { maxPrefetchDistance: PREFETCH_DISTANCE, maxPrefetchChunks: MAX_PREFETCH_CHUNKS },
   });
 }
--- a/public/index.ts
+++ b/public/index.ts
@@ -86,8 +86,7 @@ const TEST_DATA: Record<string, TestDataSpec> = {
 
 let view3D: View3d;
 
-// const volumeCache = new VolumeCache(CACHE_MAX_SIZE);
-const loadWorker = new LoadWorker(CACHE_MAX_SIZE);
+const loadWorker = new LoadWorker(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
 
 const myState: State = {
   file: "",
@@ -1010,7 +1009,6 @@ function createTestVolume() {
 
 async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
   await loadWorker.onOpen();
-  console.log(new Vector2(0, 1));
   if (data.type === "opencell") {
     return new OpenCellLoader();
   }
@@ -1024,7 +1022,9 @@ async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
     }
   }
 
-  return await loadWorker.createLoader(path);
+  return await loadWorker.createLoader(path, {
+    fetchOptions: { maxPrefetchDistance: PREFETCH_DISTANCE, maxPrefetchChunks: MAX_PREFETCH_CHUNKS },
+  });
 }
 
 async function loadVolume(loadSpec: LoadSpec, loader: IVolumeLoader): Promise<void> {
--- a/public/index.ts
+++ b/public/index.ts
@@ -1008,10 +1008,11 @@ function createTestVolume() {
 }
 
 async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
-  await loadWorker.onOpen();
   if (data.type === "opencell") {
     return new OpenCellLoader();
   }
+
+  await loadWorker.onOpen();
 
   let path: string | string[] = data.url;
   if (data.type === VolumeFileFormat.JSON) {
--- a/src/loaders/IVolumeLoader.ts
+++ b/src/loaders/IVolumeLoader.ts
@@ -43,17 +43,9 @@ export type RawChannelDataCallback = (ch: number, data: Uint8Array, atlasDims?: 
  * Loaders may keep state for reuse between volume creation and volume loading, and should be kept alive until volume
  * loading is complete. (See `createVolume`)
  */
-export abstract class IVolumeLoader {
+export interface IVolumeLoader {
   /** Use VolumeDims to further refine a `LoadSpec` for use in `createVolume` */
-  abstract loadDims(loadSpec: LoadSpec): Promise<VolumeDims[]>;
-
-  abstract createImageInfo(loadSpec: LoadSpec): Promise<[ImageInfo, LoadSpec]>;
-
-  abstract loadRawChannelData(
-    imageInfo: ImageInfo,
-    loadSpec: LoadSpec,
-    onData: RawChannelDataCallback
-  ): Promise<[ImageInfo | undefined, LoadSpec | undefined]>;
+  loadDims(loadSpec: LoadSpec): Promise<VolumeDims[]>;
 
   /**
    * Create an empty `Volume` from a `LoadSpec`, which must be passed to `loadVolumeData` to begin loading.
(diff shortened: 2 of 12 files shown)


public/index.ts
```diff
@@ -88,8 +86,7 @@ const TEST_DATA: Record<string, TestDataSpec> = {
 
 let view3D: View3d;
 
-const volumeCache = new VolumeCache(CACHE_MAX_SIZE);
-const requestQueue = new SubscribableRequestQueue(CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
+const loadWorker = new LoadWorker(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
```
toloudis, line 89: this feels a bit funky to me.  LoadWorker looks like something that should not be exposed to outside clients - it should get an API name that sounds like some important and official part of the volume-viewer,  like VolumeLoaderSubsystem or VolumeLoaderFactory?  I'll read further and digest some more and maybe come back and edit this comment...

public/index.ts
```diff
@@ -1011,6 +1008,7 @@ function createTestVolume() {
 }
 
 async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
+  await loadWorker.onOpen();
```
toloudis, line 1011: because OpenCell is such a special case, maybe it should be checked first before even touching loadWorker?

src/workers/types.ts
```diff
@@ -0,0 +1,65 @@
...
+  CREATE_VOLUME,
+  LOAD_DIMS,
+  LOAD_VOLUME_DATA,
+}
+
+export const enum WorkerResponseKind {
+  SUCCESS,
+  ERROR,
+  EVENT,
+}
```
ShrimpCryptid, line 17: Nit: Kind vs. Type?

src/workers/VolumeLoadWorker.ts
```diff
@@ -0,0 +1,93 @@
...
+  let message: WorkerResponse<T>;
+
+  try {
+    const response = await messageHandlers[type](payload);
+    message = { responseKind: WorkerResponseKind.SUCCESS, msgId, type, payload: response };
+  } catch (e) {
+    message = { responseKind: WorkerResponseKind.ERROR, msgId, type, payload: (e as Error).message };
+  }
+  self.postMessage(message);
+};
```
ShrimpCryptid, line 93: This is shockingly clean, wow

src/workers/VolumeLoadWorker.ts
```diff
@@ -0,0 +1,93 @@
...
+      (channelIndex, data, atlasDims) => {
+        const message: WorkerResponse<WorkerMsgType> = {
+          responseKind: WorkerResponseKind.EVENT,
+          loaderId,
+          loadId,
+          channelIndex,
+          data,
+          atlasDims,
+        };
+        (self as unknown as Worker).postMessage(message, copyOnLoad ? [] : [data.buffer]);
```
ShrimpCryptid, line 76: Why are two type casts (to `unknown` and then to `Worker`) needed here?

frasercl: The default type of `self` is `WorkerGlobalScope`, which as I recall has the `postMessage` method but doesn't include the overload with the `transfer` argument (the second one). We have to cast via `unknown` because `Worker` and `WorkerGlobalScope` are apparently too dissimilar to be cast to each other directly.

src/workers/types.ts
frasercl: Could probably use a more specific word. I wanted to capture that the information this represents (the way in which a request is being responded to) is orthogonal to a message's "type" (the specific action being requested). `WorkerResponseResult`?

toloudis: Result does sound good

src/loaders/IVolumeLoader.ts
```diff
@@ -34,6 +35,8 @@ export class VolumeDims {
  */
 export type PerChannelCallback = (volume: Volume, channelIndex: number) => void;
 
+export type RawChannelDataCallback = (channelIndex: number, data: Uint8Array, atlasDims?: [number, number]) => void;
```
toloudis, line 38: is atlasDims the number of rows and columns to partition the Uint8Array into?

src/loaders/IVolumeLoader.ts
```diff
@@ -65,3 +64,60 @@ export interface IVolumeLoader {
...
+   * Returns a promise that resolves to reflect any modifications to `imageInfo` and/or `loadSpec` that need to be made
+   * based on this load. Actual loaded channel data is passed to `onData` as it is loaded. Depending on the format,
+   * the returned array may be in simple 3d dimension order or reflect a 2d atlas. If the latter, the dimensions of the
+   * atlas are passed as the third argument to `onData`.
+   */
+  abstract loadRawChannelData(
+    imageInfo: ImageInfo,
+    loadSpec: LoadSpec,
+    onData: RawChannelDataCallback
+  ): Promise<[ImageInfo | undefined, LoadSpec | undefined]>;
```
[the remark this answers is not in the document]
toloudis: Is this function only called internally?  Could it be private?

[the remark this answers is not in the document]
toloudis: > would it make any sense for the promise type to include the arguments to RawChannelDataCallback? Or will the ImageInfo,Loadspec resolve at a different time than the onData callback is called?

I think I now understand that this function is also potentially loading multiple channels and needs to fire the callback once per channel received.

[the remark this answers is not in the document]
frasercl: At least in the case of `OMEZarrLoader`, this function isn't `async` and returns a promise that resolves immediately.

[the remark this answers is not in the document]
frasercl: > I think I now understand that this function is also potentially loading multiple channels and needs to fire the callback once per channel received.

Right - this is basically supposed to be a stripped-down equivalent to `loadVolumeData`, so the callback argument works the same as it would there.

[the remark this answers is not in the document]
frasercl: > Is this function only called internally? Could it be private? (I forget if ts has a notion of "protected")

The intent of this function is to be a version of `loadVolumeData` that doesn't need access to the `Volume` it's working on, so it's called externally when a loader is running on a worker. It just so happened that all volume loaders modified `Volume` in a similar enough way to also allow a common implementation of `loadVolumeData` based around it.

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -233,67 +224,50 @@ class JsonImageInfoLoader implements IVolumeLoader {
...
+      for (let ch = 0; ch < Math.min(image.channels.length, 4); ++ch) {
+        const chindex = image.channels[ch];
+        cache?.insert(`${image.name}/${chindex}`, channelsBits[ch]);
+        // NOTE: the atlas dimensions passed in here are currently unused by `JSONImageInfoLoader`
+        onData(chindex, channelsBits[ch], [bitmap.width, bitmap.height]);
+      }
+    });
 
-    return requests;
+    await Promise.all(loadPromises);
```
toloudis, line 270: is this forcing us to block on all channels?  I guess the onData calls will still happen asynchronously but still I wonder if we need to await here?

src/loaders/IVolumeLoader.ts
```diff
@@ -65,3 +64,60 @@ export interface IVolumeLoader {
...
+  abstract createImageInfo(loadSpec: LoadSpec): Promise<[ImageInfo, LoadSpec]>;
+
+  /**
+   * Begins loading per-channel data for the volume specified by `imageInfo` and `loadSpec`.
+   *
+   * Returns a promise that resolves to reflect any modifications to `imageInfo` and/or `loadSpec` that need to be made
+   * based on this load. Actual loaded channel data is passed to `onData` as it is loaded. Depending on the format,
+   * the returned array may be in simple 3d dimension order or reflect a 2d atlas. If the latter, the dimensions of the
+   * atlas are passed as the third argument to `onData`.
+   */
```
ShrimpCryptid, line 89: Really appreciate these comments 👌👌

src/loaders/IVolumeLoader.ts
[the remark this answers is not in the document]
toloudis: Makes me wonder if this could just be the new IVolumeLoader interface and not need the layer of ThreadableVolumeLoader

src/loaders/IVolumeLoader.ts
```diff
@@ -65,3 +64,60 @@ export interface IVolumeLoader {
...
+  abstract loadDims(loadSpec: LoadSpec): Promise<VolumeDims[]>;
+
+  /**
+   * Creates an `ImageInfo` object from a `LoadSpec`, which may be passed to the `Volume` constructor to create an
+   * empty volume that can accept data loaded with the given `LoadSpec`.
+   *
+   * Also returns a new `LoadSpec` that may have been modified from the input `LoadSpec` to reflect the constraints or
+   * abilities of the loader. This new `LoadSpec` should be used when constructing the `Volume`, _not_ the original.
+   */
+  abstract createImageInfo(loadSpec: LoadSpec): Promise<[ImageInfo, LoadSpec]>;
```
toloudis, line 80: What do you think of this?  Because I saw it used in a couple of places.

```suggestion
  abstract createImageInfo(loadSpec: LoadSpec): Promise<{imageInfo:ImageInfo, loadSpec:LoadSpec}>;
```

src/loaders/OpenCellLoader.ts
toloudis, line 1: Some day this OpenCellLoader should be removed from here and put in the test app only

src/workers/VolumeLoadWorker.ts
toloudis: I agree, this whole file is a very elegant way to set up a web worker to handle different kinds of messages!  It looks like tasks that need to transfer data do so in their own postMessage calls, and then those are always followed by a SUCCESS postMessage, right?

src/workers/VolumeLoadWorker.ts
ShrimpCryptid: weirddd

ShrimpCryptid:
Overall looks really good, I like how clean the change to the ThreadableLoader was! I'm going to wait until all the comments are resolved before approving. 👌

src/workers/LoadWorkerHandle.ts
```diff
@@ -0,0 +1,207 @@
...
+ * A handle that holds the worker and lets us interact with it through async calls and events rather than messages.
+ * This is separate from `LoadWorker` so that `sendMessage` and `onChannelData` can be shared with `WorkerLoader`s
+ * without leaking the API outside this file.
+ */
+class SharedLoadWorkerHandle {
+  private worker: Worker;
+  private pendingRequests: (StoredPromise<WorkerMsgType> | undefined)[] = [];
+  private workerOpen = true;
+
+  public onChannelData: ((e: ChannelLoadEvent) => void) | undefined = undefined;
```
toloudis, line 32: where is this used? Not sure why it is stored here.

public/index.ts
toloudis: Coming back to this, it does seem like VolumeLoaderFactory is a good name, since the main point of the class is to call `createLoader`.  We should document it as one of the key top level exports of the volume-viewer library, like Volume, View3D etc.

src/loaders/IVolumeLoader.ts
frasercl: `atlasDims` is the width/height of the atlas in pixels. It's basically there to account for the fact that `OMEZarrLoader` gets data into a `Volume` via `setChannelDataFromVolume`, but `JsonImageInfoLoader` gets data in with `setChannelDataFromAtlas`. The latter has arguments for `atlasWidth` and `atlasHeight`, so if `atlasDims` is present in the callback then `ThreadableVolumeLoader` calls it with those arguments, and if it isn't it calls `setChannelDataFromVolume`.

public/index.ts
frasercl: Peyton and I discussed this a bit and my favorite idea was `VolumeLoaderContext`. I don't 100% agree that "loader" is too behind-the-scenes a term to be included in the public API, but I definitely think this PR could use some work to disambiguate all the items named with the words "loader," "worker," "volume" in some combination.

src/workers/VolumeLoadWorker.ts
frasercl: In the case of the one action with events, the SUCCESS actually comes before all the events. Though that's just a consequence of how the actual function it's calling (`loadRawChannelData`) works: it returns immediately with updates to `ImageInfo` and/or `LoadSpec`, then runs callbacks when data actually arrives. So I suppose you can think of the `postMessage` here as capturing when a function returns, and ones in tasks as capturing callbacks being fired.

src/loaders/JsonImageInfoLoader.ts
frasercl: We don't need to await here, in fact I don't think this function is ever await-ed by its callers so this has no impact on control flow. It just seemed odd to produce an array of promises and do nothing with them (though I guess the `map` above could be a `forEach` instead?)

src/workers/LoadWorkerHandle.ts
frasercl: The idea of this class is to track outgoing messages to the worker, and to convert the stream of incoming messages from the worker into resolved `Promise`s and triggered events. This is meant to work a bit like attaching an event listener: `WorkerLoader`s can set this property to handle data loads (recall that only one `WorkerLoader` can be "active" at a time, so we shouldn't get collisions, though we do still check that messages received are meant for the current loader). It's called by this class anytime it receives a message from the worker with `responseKind: WorkerResponseResult.EVENT`.

src/loaders/IVolumeLoader.ts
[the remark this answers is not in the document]
frasercl: I'm 50/50 on doing that. On the one hand, `IVolumeLoader` still represents a smaller set of requirements and there's a genuine distinction between when a property or method should be typed with one vs. the other. On the other hand, I have a hard time imagining an `IVolumeLoader` that couldn't be implemented via `ThreadableVolumeLoader` - i.e., one that would want to make changes other than those permitted by `ThreadableVolumeLoader` (modify `ImageInfo` more than once, say, or call some method on a `Volume`). 

Maybe the cleanest way from a types perspective would be to convert the abstract methods of `ThreadableVolumeLoader` to the new `IVolumeLoader` interface and move the `Volume`-changing code currently implemented within `ThreadableVolumeLoader` somewhere else? But it's not super obvious to me where it might go instead.

public/index.ts
```diff
@@ -86,7 +86,7 @@ const TEST_DATA: Record<string, TestDataSpec> = {
 
 let view3D: View3d;
 
-const loadWorker = new LoadWorker(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
+const loadWorker = new VolumeLoaderContext(CACHE_MAX_SIZE, CONCURRENCY_LIMIT, PREFETCH_CONCURRENCY_LIMIT);
```
ShrimpCryptid: Rename this variable too?

ShrimpCryptid (approved):
New changes look sensible, thank you for addressing the readability concerns! Appreciate the extra comments too.

src/loaders/IVolumeLoader.ts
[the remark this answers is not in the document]
toloudis: Lets leave this for later discussion... I thought I was noticing that everything is now a ThreadableVolumeLoader and there are no pure IVolumeLoaders.

toloudis (approved):
great stuff!

PR #182, opened 2024-02-12
Prefetch priority

### LINKED ISSUE #179: generalize prefetching to allow preferred axis and ranges
If you are playing through T in XY mode, don't fetch +/- Z.  And vice versa.

frasercl:
Resolve #179: allows certain prefetch directions to be "prioritized," so e.g. if the user is playing through time, we don't waste fetches requesting adjacent z-slices at each timestep we visit.

- `ChunkPrefetchIterator`, which facilitates picking the chunks we prefetch, gets a new priority feature and a new unit test to match.
- `OMEZarrLoader` gets a new method, `setPrefetchPriorityDirections`, which sets priority. It also gets some new documentation, while I was there.
- `WorkerLoader` gets an identical method, and glue code to pass priority info into a workerized loader if it's an `OMEZarrLoader`.
- The whole package gets a new export: `PrefetchDirection`, the `const enum` previously used only internally to `ChunkPrefetchIterator`, which is now required to express prefetch direction priority.

### DIFF
--- a/src/workers/VolumeLoadWorker.ts
+++ b/src/workers/VolumeLoadWorker.ts
@@ -1,7 +1,6 @@
 import VolumeCache from "../VolumeCache";
 import { VolumeFileFormat, createVolumeLoader, pathToFileType } from "../loaders";
 import { ThreadableVolumeLoader } from "../loaders/IVolumeLoader";
-import { OMEZarrLoader } from "../loaders/OmeZarrLoader";
 import RequestQueue from "../utils/RequestQueue";
 import SubscribableRequestQueue from "../utils/SubscribableRequestQueue";
 import {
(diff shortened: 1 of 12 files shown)


src/workers/VolumeLoadWorker.ts
```diff
@@ -77,6 +78,12 @@ const messageHandlers: { [T in WorkerMsgType]: MessageHandler<T> } = {
       }
     );
   },
+
+  [WorkerMsgType.SET_PREFETCH_PRIORITY_DIRECTIONS]: (directions) => {
+    // Silently does nothing if the loader isn't an `OMEZarrLoader`
+    (loader as OMEZarrLoader | undefined)?.setPrefetchPriorityDirections?.(directions);
```
toloudis, line 84: Would it be possible to have IVolumeLoader provide a default implementation that does nothing, and then have only OMEZarrLoader override the implementation so you don't have to do a type check/coercion here?

ShrimpCryptid (approved):
LOVE the new documentation and commenting. This looks great and was easy to follow.

toloudis (approved):
👍

PR #183, opened 2024-02-15
Fix flashing during Z slice playback

frasercl:
### Problem

Currently, when playing through Z slices in website-3d-cell-viewer, if the next slice is across a chunk boundary (which it always is with our new chunking), the current slice disappears before the next one loads, creating a flashing effect. This happens because the viewer immediately moves to the next unloaded z slice, rather than lingering on the current one until the next is available.

### Solution

I spun my wheels on this for a sec, thinking I'd have to mess with how/when a `Volume`'s `ImageInfo` is updated on new loads. Turns out the trouble was just with the slice setting in `Atlas2DSlice`. Now, if an update to `zSlice` would move the viewed slice into an unloaded region, that update is instead deferred until a volume dimension update indicates that the region is loaded.

### DIFF
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -67,11 +67,15 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
     this.updateSettings(settings, SettingsFlags.ALL);
   }
 
-  private updateSlice(): void {
+  /**
+   * Syncs `this.settings.zSlice` with the corresponding shader uniform, or defers syncing until the slice is loaded.
+   * @returns a boolean indicating whether the slice is out of bounds of the volume entirely.
+   */
+  private updateSlice(): boolean {
     const slice = Math.floor(this.settings.zSlice);
     const sizez = this.volume.imageInfo.volumeSize.z;
     if (slice < 0 || slice >= sizez) {
-      return;
+      return false;
     }
 
     const regionMinZ = this.volume.imageInfo.subregionOffset.z;
@@ -83,6 +87,8 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
       this.setUniform("Z_SLICE", slice);
       this.sliceUpdateWaiting = false;
     }
+
+    return true;
   }
 
   public updateVolumeDimensions(): void {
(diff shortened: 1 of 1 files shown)


src/Atlas2DSlice.ts
```diff
@@ -147,15 +169,11 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
       this.setUniform("AABB_CLIP_MAX", bounds.bmax);
-      const slice = Math.floor(this.settings.zSlice);
-      const sizez = this.volume.imageInfo.volumeSize.z;
-      if (slice >= 0 && slice <= sizez - 1) {
-        this.setUniform("Z_SLICE", slice);
-        const sliceRatio = slice / sizez;
-        this.volume.updateRequiredData({
-          subregion: new Box3(new Vector3(0, 0, sliceRatio), new Vector3(1, 1, sliceRatio)),
-        });
-      }
```
ShrimpCryptid, line 158: The bounds checking that was originally done here is now done in `updateSlice`, but it doesn't stop the call to `this.volume.updateRequiredData`... Should `updateSlice` have a return value to indicate if the slice is in range?

src/Atlas2DSlice.ts
```diff
@@ -147,15 +169,11 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
+      this.updateSlice();
+      const sliceRatio = Math.floor(this.settings.zSlice) / this.volume.imageInfo.volumeSize.z;
+      this.volume.updateRequiredData({
+        subregion: new Box3(new Vector3(0, 0, sliceRatio), new Vector3(1, 1, sliceRatio)),
+      });
```
toloudis, line 176: +1 for Peyton's comment. The out of bounds check probably should skip this code too, as it was before.  I imagine that the upper levels of the code really prevent it from being out of range but it's still a good sanity check here. Does that sound right?

src/Atlas2DSlice.ts
frasercl: Yep - whoops! I remember making a mental note that I'd have to preserve that behavior when factoring out updates into their own function, but I missed it when I circled back and actually did it.

src/Atlas2DSlice.ts
frasercl: Yeah, I'd bet this can never occur in practice, but it's still a good idea to preserve the sanity check.

toloudis (approved):

ShrimpCryptid (approved):

PR #185, opened 2024-02-24
sync channels during playback

toloudis:
Adds capability to wait for all channels to arrive before updating volume data.  This is for time series or Z slice playback when multiple channels are enabled.  When it was always async, it was hard to watch playback due to channels arriving on alternate redraws.    Front end implementation should turn sync on during playback and turn it off when not in playback.  (We can also consider always having it on)

Estimated time to review: ~15 min

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -920,6 +920,7 @@ function onVolumeCreated(volume: Volume) {
 
 function playTimeSeries(onNewFrameCallback: () => void) {
   window.clearTimeout(myState.timerId);
+  myState.loader.syncMultichannelLoading(true);
   myState.isPlaying = true;
 
   const loadNextFrame = () => {
@@ -1180,6 +1181,7 @@ function main() {
   pauseBtn?.addEventListener("click", () => {
     window.clearTimeout(myState.timerId);
     myState.isPlaying = false;
+    myState.loader.syncMultichannelLoading(false);
   });
 
   const forwardBtn = document.getElementById("forwardBtn");
--- a/public/index.ts
+++ b/public/index.ts
@@ -1035,7 +1035,7 @@ async function createLoader(data: TestDataSpec): Promise<IVolumeLoader> {
 async function loadVolume(loadSpec: LoadSpec, loader: IVolumeLoader): Promise<void> {
   const volume = await loader.createVolume(loadSpec, onChannelDataArrived);
   onVolumeCreated(volume);
-  loader.loadVolumeData(volume, false);
+  loader.loadVolumeData(volume);
 
   // Set default zSlice
   goToZSlice(Math.floor(volume.imageInfo.subregionSize.z / 2));
--- a/public/index.ts
+++ b/public/index.ts
@@ -40,6 +40,10 @@ const TEST_DATA: Record<string, TestDataSpec> = {
     type: VolumeFileFormat.TIFF,
     url: "https://animatedcell-test-data.s3.us-west-2.amazonaws.com/AICS-12_881.ome.tif",
   },
+  zarrEMT: {
+    url: "https://dev-aics-dtp-001.int.allencell.org/dan-data/3500005818_20230811__20x_Timelapse-02(P27-E7).ome.zarr",
+    type: VolumeFileFormat.ZARR,
+  },
   zarrIDR1: {
     type: VolumeFileFormat.ZARR,
     url: "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0076A/10501752.zarr",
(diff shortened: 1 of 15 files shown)


src/loaders/IVolumeLoader.ts
```diff
@@ -41,7 +41,8 @@ export type LoadedVolumeInfo = {
  */
 export type PerChannelCallback = (volume: Volume, channelIndex: number) => void;
 
-export type RawChannelDataCallback = (channelIndex: number, data: Uint8Array, atlasDims?: [number, number]) => void;
+// allow lists of channel indices and data arrays to be passed to the callback
+export type RawChannelDataCallback = (channelIndex: number[], data: Uint8Array[], atlasDims?: [number, number]) => void;
```
toloudis, line 45: key change: the channel data callback can now handle groups of channels

src/loaders/IVolumeLoader.ts
```diff
@@ -119,13 +134,17 @@ export abstract class ThreadableVolumeLoader implements IVolumeLoader {
     loadSpecOverride?: LoadSpec,
     onChannelLoaded?: PerChannelCallback
   ): Promise<void> {
-    const onChannelData: RawChannelDataCallback = (channelIndex, data, atlasDims) => {
-      if (atlasDims) {
-        volume.setChannelDataFromAtlas(channelIndex, data, atlasDims[0], atlasDims[1]);
-      } else {
-        volume.setChannelDataFromVolume(channelIndex, data);
+    const onChannelData: RawChannelDataCallback = (channelIndices, dataArrays, atlasDims) => {
```
toloudis, line 137: modified to loop over channels arriving and handle all of them

src/loaders/OmeZarrLoader.ts
```diff
@@ -496,7 +505,12 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
       try {
         const result = await zarrGet(level, sliceSpec, { opts: { subscriber, reportKey } });
         const u8 = convertChannel(result.data);
-        onData(ch, u8);
+        if (this.syncChannels) {
+          resultChannelData.push(u8);
+          resultChannelIndices.push(ch);
+        } else {
+          onData([ch], [u8]);
+        }
```
toloudis, line 513: key change above: if syncChannels, then don't call onData but instead gather them into an array to call onData later after we are sure they have all arrived.

public/index.ts
```diff
@@ -1176,6 +1181,7 @@ function main() {
   pauseBtn?.addEventListener("click", () => {
     window.clearTimeout(myState.timerId);
     myState.isPlaying = false;
+    myState.loader.syncMultichannelLoading(false);
```
ShrimpCryptid, line 1184: What happens if `syncMultiChannelLoading` is turned off midway through a load?

src/loaders/OmeZarrLoader.ts
```diff
@@ -515,6 +529,9 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
     this.beginPrefetch(keys, level);
 
     Promise.all(channelPromises).then(() => {
+      if (this.syncChannels) {
+        onData(resultChannelIndices, resultChannelData);
+      }
```
ShrimpCryptid, line 534: Left a similar comment above. If `syncChannels` is turned off, is the fetched data discarded? (Would a mismatch between the UI and loaded data occur, since pausing would stop at the currently rendered frame?)

src/workers/LoadWorkerHandle.ts
```diff
@@ -212,10 +212,14 @@ class WorkerLoader extends ThreadableVolumeLoader {
    * Change which directions to prioritize when prefetching. All chunks will be prefetched in these directions before
    * any chunks are prefetched in any other directions. Has no effect if this loader doesn't support prefetching.
    */
-  setPrefetchPriorityDirections(directions: PrefetchDirection[]): Promise<void> {
+  setPrefetchPriority(directions: PrefetchDirection[]): Promise<void> {
```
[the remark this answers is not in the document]
frasercl: Perhaps it was a mistake to provide default implementations in `ThreadableVolumeLoader`?

frasercl (approved):
This is a refreshingly small change for what I thought might be a much more challenging problem!

ShrimpCryptid (approved):

src/loaders/OmeZarrLoader.ts
```diff
@@ -146,6 +146,8 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
   //   This may cause errors or incorrect results otherwise!
   private maxExtent?: Box3;
 
+  private syncChannels = false;
```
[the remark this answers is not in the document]
toloudis: Yes it seems reasonable to be able to initialize it to true!  I'll add it.

Issue #78, opened 2022-11-29
load from multiple data sources


toloudis:
**Background**

Images often have other derivative images computed and stored in a different place. 
In order to combine them for CFE, we rebuild combined images which results in redundant storage.  For other applications the burden is entirely on the users to do this.
How should users specify multiple data sources from which to load image data?

Today you can specify a single file location in the url of the standalone viewer.   And in CFE, the application provides that url.

User might know all the data sources initially, or they might add channels to the image after the first batch is loaded.  I'm not sure if both of these modes are needed.
 
Sharmi Seshamani of AIND is our example user who specified this request initially.

**User needs statement(s)**

As a user, I want to be load a raw microscope image and its segmentation image from different storage locations, and have them show up as a combined multichannel image.

**Size**
m

**Requirements**
Discovery, Final design deliverables

**Deliverables**
Meeting with decision making.

**Stakeholder(s)**
@frasercl @toloudis

**Timeline**

**Notes**

We will need to deal with error conditions such as dimension mismatches etc.

toloudis:
there should be a GUI path and a scripted url/json path to get data in to the viewer.

as a start, url and a selection of channels, and if time series, a time index.

user says "add data"
user picks url
we fetch metadata and show user what channels are available.
user picks channels
we load them.
if xyz dims do not match, then what? fail? yes that's an error until we want to do something more clever.

toloudis:
get urls from Sharmi maybe, to have a practical test for this feature

toloudis:
also see https://github.com/allen-cell-animated/volume-viewer/issues/87

toloudis:
Requirement change:
For a data release, we would like to implement this only at the url-query-parameter level.  We need to tell the viewer to get raw data from one zarr and segmentation data from a different one.
Something like:
viewer.html?url=first_channels_zarr&url2=more_channels_zarr 
?  Maybe something better than url2 but you get the idea.

Internally we have discussed being able to pass both urls to the volume loader.

We will use the first url as the dimensions that all following urls have to match.
Requirement:
our segmentations will have level1 match the raw level0 in resolution.  We know our segmentations have level0 at 100x and raw is 20x.  We are planning to generate zarrs which downsample the segmentation from level0 to level1 by that 100x-to-20x factor so that it matches the raw's level0.  We need the viewer to be able to try to find the matching resolution. Probably ok to error out if no match is found.

PR #186, opened 2024-02-27
Update unit test packages

frasercl:
Quick review (<5min). (Do not be fooled by the change numbers. It's all in package-lock.json!)

I encountered a problem importing some zarrita functions while writing unit tests for #78. Updating the test packages, while a good thing to do for maintenance, seemed to break everything even more. The solution to both problems was to convert this project to explicitly use ES modules (which provide the `import Foo from "bar";` syntax) rather than CommonJS (where imports look like `const Foo = require("bar");`). Shout out to @toloudis for patiently wading with me through all these config and command line option changes.

While I was messing with config files, I also added format on save.


ShrimpCryptid (approved):
LGTM! Glad that the solution was simple and explainable :')

ascibisz (approved):
Looks good to me!

PR #188, opened 2024-03-05
Multi-source zarr

frasercl:
Rough review time: long (60min)

Implements most of #78: enable `OMEZarrLoader` to load from multiple zarrs at once, and enable it to resolve potential mismatches in scale levels.

### How this works to outside users

- `OMEZarrLoader.createLoader`'s `url` param is now `urls`, and can take either a `string` or a `string[]`.
  - If `urls` is an array with more than one element, the loader loads the dimensions of every scale level in every zarr, then tries to match up each zarr's scale levels by size.
  - If the above step succeeds, the resulting volume behaves as if it were loaded from a single OME-Zarr with all the channels of all zarr sources in the order they were specified in `urls`, and all scale levels that "matched up" in all sources.
- `scene: number` is updated to `scenes: number | number[]` to match, though it can be a single number when `urls` is an array to get the same scene index from all sources.

### What I changed to make that work

- Move every property of `OMEZarrLoader` that could differ based on zarr sources into its own type, `ZarrSource`, which we can put in an array.
- Rework `OMEZarrLoader` to generate an array of `ZarrSource`s on construction and use them in its regular methods (a large portion of the changes in this PR are just dealing with the fact that all the properties that the loader has to read are inside an array of `ZarrSource`s now)
- Add a big new function, `matchSourceScaleLevels`, to zarr utils that does the "matching up" described above by removing unmatched scale levels from an array of `ZarrSource`s.
- Add unit tests for `matchSourceScaleLevels`.


src/loaders/zarr_utils/WrappedStore.ts
frasercl, line 1: All the changes in this file are just to make it easier to make mock `ZarrSource`s for testing (zarrita won't let you create your own zarr array objects unless the store they connect to implements its `Readable` or `AsyncReadable` interfaces).

src/loaders/zarr_utils/types.ts
```diff
@@ -93,3 +93,20 @@ export type OMEZarrMetadata = {
...
+   * dimension order as a mapping from dimensions to their indices in dimension-ordered arrays for this source.
+   */
+  axesTCZYX: TCZYX<number>;
+  /** OME-specified metadata record with most useful info on the current image, e.g. sizes, axis order, etc. */
+  multiscaleMetadata: OMEMultiscale;
+  /** OME-specified "transitional" metadata record which we mostly ignore, but which gives channel & volume names. */
+  omeroMetadata: OmeroTransitionalMetadata;
+  /** Which channels in the volume come out of this source - i.e. source channel 0 is volume channel `channelOffset` */
+  channelOffset: number;
+};
```
frasercl, line 112: This is the essential type added by this PR. Most of the fields correspond to fields removed from `OMEZarrLoader` because they describe just one zarr and we need to handle multiple. `channelOffset` is the only new one.

src/loaders/OmeZarrLoader.ts
```diff
@@ -310,22 +395,24 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+      const numDims = getDimensionCount(this.sources[sourceIdx].axesTCZYX);
       const coordsInDimensionOrder = key
         .trim()
         .split("/")
         .slice(-numDims)
         .filter((s) => s !== "")
         .map((s) => parseInt(s, 10));
-      return this.orderByTCZYX(coordsInDimensionOrder, 0);
+      const sourceCoords = this.orderByTCZYX(coordsInDimensionOrder, 0, sourceIdx);
+      sourceCoords[1] += this.sources[sourceIdx].channelOffset;
```
toloudis, line 409: a comment would be nice to reassure me that we are incrementing the channel index for a good downstream reason because it no longer is a valid "chunk" coordinate (not sure how it gets used)

src/loaders/OmeZarrLoader.ts
```diff
@@ -343,7 +430,10 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
       if (prefetchCount >= this.fetchOptions.maxPrefetchChunks) {
         break;
       }
-      this.prefetchChunk(scaleLevel.path, chunk, subscriber);
+      const [sourceIdx, sourceCh] = this.matchChannelToSource(chunk[1]);
```
toloudis, line 433: 👍

src/loaders/OmeZarrLoader.ts
```diff
@@ -360,23 +450,23 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
   ): Promise<{ imageInfo: ImageInfo }> {
     const maxExtent = this.maxExtent ?? new Box3(new Vector3(0, 0, 0), new Vector3(1, 1, 1));
-    const [z, y, x] = this.axesTCZYX.slice(2);
+    const [z, y, x] = this.sources[0].axesTCZYX.slice(2);
     const subregion = composeSubregion(loadSpec.subregion, maxExtent);
 
     const levelIdx = pickLevelToLoad({ ...loadSpec, subregion }, this.getLevelShapesZYX());
-    const level = this.scaleLevels[levelIdx];
-    const levelShape = level.shape;
+    const arraysAtLevel = this.sources.map((src) => src.scaleLevels[levelIdx]);
```
toloudis, line 457: maybe this is just `level = this.sources[0].scaleLevels[levelIdx]; levelShape=level.shape` without having to map a new array ?

src/loaders/zarr_utils/WrappedStore.ts
```diff
@@ -17,8 +17,14 @@ type WrappedStoreOpts<Opts> = {
  * `Readable` is zarrita's minimal abstraction for any source of data.
  * `WrappedStore` wraps another `Readable` and adds (optional) connections to `VolumeCache` and `RequestQueue`.
  */
-class WrappedStore<Opts, S extends Readable<Opts> = Readable<Opts>> implements AsyncReadable<WrappedStoreOpts<Opts>> {
+class WrappedStore<Opts, S extends Readable<Opts> = Readable<Opts>>
```
toloudis, line 20: Nit: Does it make sense to change this to WrappedReadOnlyStore now that you have a explicit set function that does nothing?  Or was it the same before this change?

src/loaders/zarr_utils/types.ts
```diff
@@ -93,3 +93,20 @@ export type OMEZarrMetadata = {
 };
 
 export type NumericZarrArray = zarr.Array<zarr.NumberDataType, WrappedStore<RequestInit>>;
+
+/** A record with everything we need to access and use a single remote source of multiscale OME-Zarr data. */
+export type ZarrSource = {
+  /** Representations of each scale level in this zarr. We pick one and pass it to `zarrGet` to load data. */
```
toloudis, line 99: where is zarrGet defined? Is that a zarrita native function?

src/loaders/zarr_utils/utils.ts
```diff
@@ -101,3 +109,130 @@ export function getScale(dataset: OMEDataset | OMEMultiscale, orderTCZYX: TCZYX<
...
+    return 1;
+  } else {
+    return undefined;
+  }
+}
+
+function scaleTransformsAreEqual(aSrc: ZarrSource, aLevel: number, bSrc: ZarrSource, bLevel: number): boolean {
+  const aScale = getScale(aSrc.multiscaleMetadata.datasets[aLevel], aSrc.axesTCZYX);
+  const bScale = getScale(bSrc.multiscaleMetadata.datasets[bLevel], bSrc.axesTCZYX);
+  return aScale[2] === bScale[2] && aScale[3] === bScale[3] && aScale[4] === bScale[4];
```
toloudis, line 145: Nit: would this comparison benefit from a small epsilon?  When the same-shape multiresolution levels are offset then the scale values could be unequal.

src/loaders/zarr_utils/utils.ts
```diff
@@ -101,3 +109,130 @@ export function getScale(dataset: OMEDataset | OMEMultiscale, orderTCZYX: TCZYX<
...
+}
+
+/**
+ * Ensures that all scale levels in `sources` are matched up by size. More precisely: enforces that, for any scale
+ * level `i`, the size of zarr array `s[i]` is equal for every source `s`. We accomplish this by removing any arrays
+ * (and their associated OME dataset metadata) which don't match up in all sources.
+ *
+ * Assumes all sources have scale levels ordered by size from largest to smallest. (This should always be true for
+ * compliant OME-Zarr data.)
```
toloudis, line 154: Just wondering, Is that confirmed by the v0.4 ome-ngff spec?

toloudis: also, should the comment for this function say that the first source is the target for which all others are trying to match?

toloudis: probably also should say that this function modifies the passed in `sources` array, or even consider having it return the modified array.

toloudis (commented):
Looking great, most of my comments are questions for understanding/readability.

src/loaders/OmeZarrLoader.ts
```diff
@@ -227,17 +285,23 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+    const [t, , z, y, x] = source0.axesTCZYX;
     const hasT = t > -1;
-    const hasC = c > -1;
     const hasZ = z > -1;
 
-    const shape0 = this.scaleLevels[0].shape;
+    const shape0 = source0.scaleLevels[0].shape;
     const levelToLoad = pickLevelToLoad(loadSpec, this.getLevelShapesZYX());
-    const shapeLv = this.scaleLevels[levelToLoad].shape;
+    const shapeLv = source0.scaleLevels[levelToLoad].shape;
```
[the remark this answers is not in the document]
toloudis: I saw later that your big matching function only gives the intersection of matching levels across all sources, which is kind of awesome.

src/loaders/zarr_utils/types.ts
frasercl: `zarrGet` is the alias I import zarrita's `get` function under in `OMEZarrLoader` - I believe they prefer if you import it as `zarr.get`, but I think I was worried about tree shaking?

src/loaders/zarr_utils/utils.ts
frasercl: I did check the spec on this: https://ngff.openmicroscopy.org/latest/#multiscale-md

> Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.

frasercl: The function doesn't really give one source any special treatment, though (per the comment in that file) I only ever use the first `multiscaleMetadata` after this function does its work.

frasercl: I'm all about immutable data and would have implemented this function to return an entirely new `sources` array if that were feasible, but `ZarrSource` is a bit too complex an object to want to clone. It's definitely worth adding a comment though to emphasize "hey, this modifies its input!" (It could also return only the fields it modifies, i.e. new `NumericZarrArray[]` and new `OMEDataset[]`?)

src/loaders/zarr_utils/utils.ts
frasercl: Good call! It probably would.

src/loaders/OmeZarrLoader.ts
```diff
@@ -144,55 +154,103 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+      channelCount += s.omeroMetadata.channels.length;
+    }
+    // Ensure the sizes of all sources' scale levels are matched up. See this function's docs for more.
+    matchSourceScaleLevels(sources);
+    // TODO: if `matchSourceScaleLevels` returned successfully, every one of these sources' `multiscaleMetadata` is the
+    // same in every field we care about, so we only ever use the first source's `multiscaleMetadata` after this point.
+    // Should we only store one `OMEMultiscale` record total, rather than one per source?
     const priorityDirs = fetchOptions?.priorityDirections ? fetchOptions.priorityDirections.slice() : undefined;
-    return new OMEZarrLoader(store, scaleLevels, multiscale, omero, axisTCZYX, queue, fetchOptions, priorityDirs);
+    return new OMEZarrLoader(sources, queue, fetchOptions, priorityDirs);
```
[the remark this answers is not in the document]
frasercl: The "private constructor" pattern I have going on here is strange as well, but I blame that mostly on JavaScript. It's a consequence of having both classes with privileged constructors and `async`, and not anticipating how they might interact (which is pretty reasonable). I feel okay about this both because I kind of just treat the constructor as the final step in this static method, and because all the *dimensions* (and metadata) loading happens here, and all the *data* loading happens in the rest of the class.

src/loaders/zarr_utils/utils.ts
toloudis: Cool, for me it is sufficient to just say it clearly in the comment that the input arg comes out modified

src/loaders/OmeZarrLoader.ts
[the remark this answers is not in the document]
frasercl: I've been searching for the right word to describe the thing that `matchSourceScaleLevels` identifies ("matched sets?" "slices of equal scale levels?") when "intersection" was right there 🙃

src/loaders/zarr_utils/types.ts
toloudis: Was only wondering because of the comment it's hard to tell whether that's our code or third party code

src/loaders/OmeZarrLoader.ts
```diff
@@ -144,55 +154,103 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+   * Converts a volume channel index to the index of its zarr source and its channel index within that zarr.
+   * e.g., if the loader has 2 sources, the first with 3 channels and the second with 2, then `matchChannelToSource(4)`
+   * returns `[1, 1]` (the second channel of the second source).
+   */
+  private matchChannelToSource(channelIdx: number): [number, number] {
+    const lastSrcIdx = this.sources.length - 1;
+    const lastSrc = this.sources[lastSrcIdx];
+    const lastSrcNumChannels = lastSrc.scaleLevels[0].shape[lastSrc.axesTCZYX[1]];
+
+    if (channelIdx > lastSrc.channelOffset + lastSrcNumChannels) {
```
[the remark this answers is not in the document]
frasercl: Without this test, the `findIndex` test below wouldn't distinguish between "a channel index *within* the last source" and "a channel index *outside* the last source." E.g. if you have a single source with 2 channels, `findChannelIndex(3)` will hit this error, but would incorrectly return `[0, 2]` (the third channel of source 0) without it.

src/loaders/OmeZarrLoader.ts
frasercl: `arraysAtLevel` is passed into `beginPrefetch` way down lower in this function, though it could also just receive the scale level index

src/loaders/zarr_utils/WrappedStore.ts
frasercl: I believe calling `set` before would have resulted in an error, so this store was read-only to begin with

src/loaders/OmeZarrLoader.ts
frasercl: The thing I described above is nicer, so I switched to that

toloudis (approved):
After the last changes, I'd consider adding a test assertion that the set of selected shapes is in descending order.  I wasn't sure if the test was capturing that.

Otherwise this looks good!  Some day down the road, it will be interesting to see if we can have a later feature in the app to "add" a second zarr after the first one is loaded.

src/loaders/zarr_utils/utils.ts
```diff
@@ -101,3 +109,135 @@ export function getScale(dataset: OMEDataset | OMEMultiscale, orderTCZYX: TCZYX<
...
+        // ...they have different numbers of timesteps
+        const largestT = smallestSrc.axesTCZYX[0] > -1 ? smallestArr.shape[smallestSrc.axesTCZYX[0]] : 1;
+        const currentT = currentSrc.axesTCZYX[0] > -1 ? currentArr.shape[currentSrc.axesTCZYX[0]] : 1;
+        if (largestT !== currentT) {
+          throw new Error("Incompatible zarr arrays: different numbers of timesteps");
+        }
+        // ...they have different chunk sizes (TODO update prefetching so this restriction can be removed)
+        if (!smallestArr.chunks.every((val, idx) => val === currentArr.chunks[idx])) {
+          throw new Error("Incompatible zarr arrays: chunk shapes are mismatched");
+        }
```
frasercl, line 202: #193 addresses this TODO. I'll move it out of draft when this PR merges.

src/loaders/OmeZarrLoader.ts
```diff
@@ -415,8 +508,12 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
       // Build slice spec
       const { min, max } = regionPx;
-      const unorderedSpec = [loadSpec.time, ch, slice(min.z, max.z), slice(min.y, max.y), slice(min.x, max.x)];
-      const sliceSpec = this.orderByDimension(unorderedSpec as TCZYX<number | Slice>);
+      const { sourceIndex: sourceIdx, channelIndexInSource: sourceCh } = this.matchChannelToSource(ch);
+      const unorderedSpec = [loadSpec.time, sourceCh, slice(min.z, max.z), slice(min.y, max.y), slice(min.x, max.x)];
+
+      const level = this.sources[sourceIdx].scaleLevels[levelIdx];
+      const sliceSpec = this.orderByDimension(unorderedSpec as TCZYX<number | Slice>, sourceIdx);
+      const reportKey = (key: string, sub: SubscriberId) => reportKeyBase(sourceIdx, key, sub);
```
meganrm, line 516: this is a pretty generic comment, but looking at this on GitHub it seems like this is still part of `loadRawChannelData`? In that case, this is a very long function that just got longer. could this "build slice spec" be it's own function ?

src/loaders/OmeZarrLoader.ts
```diff
@@ -302,12 +383,17 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
-    const separator = basePath.endsWith("/") ? "" : "/";
-    const key = basePath + separator + this.orderByDimension(coords).join("/");
+  private async prefetchChunk(
+    scaleLevel: NumericZarrArray,
+    coords: TCZYX<number>,
+    subscriber: SubscriberId
+  ): Promise<void> {
+    const { store, path } = scaleLevel;
+    const separator = path.endsWith("/") ? "" : "/";
+    const key = path + separator + this.orderByDimension(coords).join("/");
```
meganrm, line 393: in my experience, this sort of string manipulation is good to factor out and have unit tests for because it's an easy place to have unintentional results

meganrm (approved):
the tests look good, my comments are minor suggestions

src/loaders/OmeZarrLoader.ts
frasercl: I took a second look at `loadRawChannelData`, and the easiest part to factor out seemed to be the `ImageInfo`-updating code at the beginning. So I moved that section to its own function. The remaining code in `loadRawChannelData` is still pretty high cognitive load (lots of nested scopes, plus this `async` closure that gets called a bunch of times) but it's all tougher to separate. Might come back later and try though.

PR #190, opened 2024-03-13
Fully qualified imports

frasercl:
Review time: pretty quick (15min)

#186 added `"type": "module"` to package.json. Unknown to us, doing this caused this package to come under stricter module resolution rules. Specifically, paths to modules must be fully specified: no leaving off the .js extension in a file path, and no expecting the resolver to implicitly find an index.js if the path is a directory. So while the test app still works just fine on main, the package breaks when used by any downstream packages (website-3d-cell-viewer).

The rules and (conflicting) standards around how the JS module system should work are... complicated... but TypeScript's [current guidance](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-specifiers-are-not-transformed) seems to be that modules should be referenced by the fully qualified path they will have *after they are built*, and TypeScript will handle resolving back from js-land to ts-land when type checking:

> There is no compiler option that enables transforming, substituting, or rewriting module specifiers. Consequently, module specifiers must be written in a way that works for the code’s target runtime or bundler, and it’s TypeScript’s job to understand those *output*-relative specifiers.

So this PR fixes the issue by fully specifying all imports across this project, as TypeScript tells us to do. It also adds a bunch of `type` keywords to imports to help TypeScript know to elide them.

### Steps to verify:

- Link your local copy of this package to website-3d-cell-viewer: run `npm link` in volume-viewer and `npm link @aics/volume-viewer` in website-3d-cell-viewer.

Then, for whichever branch you want to test:
- `npm run build` volume-viewer
- `npm start` website-3d-cell-viewer

Expected behavior on `main`: lots of errors.

Expected behavior on this branch (`fix/fully-qualified-imports`): works just fine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,7 +51,6 @@
         "raw-loader": "^4.0.2",
         "react": "^16.14.0",
         "react-dom": "^16.14.0",
-        "regenerator-runtime": "^0.13.9",
         "rimraf": "^3.0.2",
         "ts-node": "^10.9.2",
         "typescript": "^4.3.5",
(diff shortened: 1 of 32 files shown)


toloudis:
`npm start` doesn't work for me on this volume-viewer branch, yet

PR #193, opened 2024-03-19
Handle per-source chunk dimensions when prefetching zarr

frasercl:
Review time: moderate (~30min)

#188 enabled loading into the same volume from multiple zarr sources (resolving most of #78), and included code to ensure those sources have scale levels with matching dimensions. However, its implementation has an important limitation: matching scale levels also must have matching chunk sizes. This limitation wasn't imposed due to any difficulties with fetching zarrs with different chunking strategies - zarrita will happily accept a slice into an array with any chunking strategy and work out which chunks to fetch on its own. But our *pre*fetching implementation assumed that it could safely apply the same chunk dimensions to every channel, which is not safe when different channels may come out of different source zarrs with different chunking strategies. So this PR allows us to deal with per-channel chunk dims when prefetching, removing the need for matching chunking strategies.

### Changes

- **ChunkPrefetchIterator.ts**: `ChunkPrefetchIterator` now accepts chunk sizes per source. The type of the corresponding field in the `ChunkPrefetchIterator` constructor was formerly `TZYX` (`[number, number, number, number]`), representing the number of chunks along those four axes. The type is now `TCZYX<number>[]`, representing the number of chunks along these four axes *per source*, and including C (the number of channels in the given source) to match up sources with channels.
- **OmeZarrLoader.ts**: now provides per-source dimensions.
- **ChunkPrefetchIterator.test.ts**: updated to use the new dimensions type described above, and added a test to verify that per-source bounds are handled as specified.
- **zarr_utils/utils.ts**: now that the above changes have been made, remove the restriction on matching chunking strategies.

### DIFF
--- a/src/loaders/zarr_utils/ChunkPrefetchIterator.ts
+++ b/src/loaders/zarr_utils/ChunkPrefetchIterator.ts
@@ -6,7 +6,16 @@ type PrefetchDirectionState = {
   direction: PrefetchDirection;
   chunks: TCZYX<number>[];
   start: number;
-  end: number;
+  /** May be either a number for all channels or an array of ends per-channels */
+  end: number | number[];
+};
+
+const allEqual = <T>(arr: T[]): boolean => arr.every((v) => v === arr[0]);
+
+const pushN = <T>(arr: T[], val: T, n: number): void => {
+  for (let i = 0; i < n; i++) {
+    arr.push(val);
+  }
 };
 
 const directionToIndex = (dir: PrefetchDirection): number => {
@@ -37,7 +46,7 @@ export default class ChunkPrefetchIterator {
   constructor(
     chunks: TCZYX<number>[],
     tzyxMaxPrefetchOffset: TZYX,
-    tzyxNumChunks: TZYX,
+    tczyxChunksPerDimension: TCZYX<number>[],
     priorityDirections?: PrefetchDirection[]
   ) {
     // Get min and max chunk coordinates for T/Z/Y/X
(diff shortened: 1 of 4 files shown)


src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -37,7 +46,7 @@ export default class ChunkPrefetchIterator {
   constructor(
     chunks: TCZYX<number>[],
     tzyxMaxPrefetchOffset: TZYX,
-    tzyxNumChunks: TZYX,
+    tczyxChunksPerDimension: TCZYX<number>[],
```
toloudis, line 49: Not sure I understand what this arg contains. Would `tczyxChunksPerSource` be more appropriate here?  tczyx are dimensions...

toloudis (approved):
one variable name was mysterious to me

src/loaders/zarr_utils/ChunkPrefetchIterator.ts
```diff
@@ -61,11 +70,26 @@ export default class ChunkPrefetchIterator {
 
     for (const [direction, start] of extrema.flat().entries()) {
       const dimension = direction >> 1;
-      let end: number;
+      const tczyxIndex = dimension + Number(dimension !== 0);
```
ascibisz, line 73: I don't quite understand this line. What's the `Number(dimension !== 0)` for? I trust that it's important, I just can't figure out why....

frasercl: Yeah - it's a silly, hard-to-parse trick to add 0 when `dimension` is 0 and 1 otherwise, to skip index 1. The contents of the function you read are inlined here just because I wanted that intermediate `dimension` value.

ascibisz (approved):

Issue #189, opened 2024-03-06
xy mode image goes to white


toloudis:
## Description

It seems that the transfer function is not resetting when switching from 3D to XY mode if loading is interrupted?
If you switch to XY and then uncheck a channel before all data arrived, the transfer function will be reused from the 3D view which is inconsistent and causes the image to be unreadable

## Expected Behavior

Data should appear with a sensible transfer function that doesn't need user to adjust

## Reproduction

1. Load  https://allen-cell-animated.github.io/website-3d-cell-viewer/?url=https://animatedcell-test-data.s3.us-west-2.amazonaws.com/20200323_F01_001/P13-C4.zarr
2. switch to XY mode
3. uncheck the second channel (before XY mode is completely loaded)
4. view turns solid white

## Environment

tested on chrome on macos or windows

toloudis:
Can't repro at the moment.  We can reopen or create new ticket if this happens again.

PR #194, opened 2024-03-20
Feature/lut refactor

toloudis:
Decoupling Lut from Histogram for future work.

There are some things about the way the lookup tables (Luts) and Histograms are handled that make it hard to update certain things. 
See #189  or https://github.com/allen-cell-animated/website-3d-cell-viewer/issues/192 for examples.

This PR is a first step and should introduce no functionality changes.

Key changes:

**Lut.ts** is a new module that is responsible only for a lookup table as array, plus control points to describe its piecewise linear function.
**Histogram.ts** no longer has any responsibilities for creating luts.  
**Channel.ts** no longer has any responsibilities for creating luts.
All the rest of the changes stem from these.

Again, no functionality is supposed to be changed as a result of this PR. 
It does have an API change for which there will need to be some changes in website-3d-cell-viewer.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -5,6 +5,7 @@ import {
   ImageInfo,
   IVolumeLoader,
   LoadSpec,
+  Lut,
   JsonImageInfoLoader,
   View3d,
   Volume,
@@ -632,7 +633,8 @@ function showChannelUI(volume: Volume) {
       // this doesn't give good results currently but is an example of a per-channel button callback
       autoIJ: (function (j) {
         return function () {
-          const lut = volume.getHistogram(j).lutGenerator_auto2();
+          const [hmin, hmax] = volume.getHistogram(j).findAutoIJBins();
+          const lut = new Lut().createFromMinMax(hmin, hmax);
           // TODO: get a proper transfer function editor
           // const lut = { lut: makeColorGradient([
           //     {offset:0, color:"black"},
@@ -648,7 +650,8 @@ function showChannelUI(volume: Volume) {
       // this doesn't give good results currently but is an example of a per-channel button callback
       auto0: (function (j) {
         return function () {
-          const lut = volume.getHistogram(j).lutGenerator_auto();
+          const [b, e] = volume.getHistogram(j).findAutoMinMax();
+          const lut = new Lut().createFromMinMax(b, e);
           volume.setLut(j, lut.lut);
           view3D.updateLuts(volume);
         };
@@ -656,7 +659,8 @@ function showChannelUI(volume: Volume) {
       // this doesn't give good results currently but is an example of a per-channel button callback
       bestFit: (function (j) {
         return function () {
-          const lut = volume.getHistogram(j).lutGenerator_bestFit();
+          const [hmin, hmax] = volume.getHistogram(j).findBestFitBins();
+          const lut = new Lut().createFromMinMax(hmin, hmax);
           volume.setLut(j, lut.lut);
           view3D.updateLuts(volume);
         };
@@ -664,7 +668,9 @@ function showChannelUI(volume: Volume) {
       // eslint-disable-next-line @typescript-eslint/naming-convention
       pct50_98: (function (j) {
         return function () {
-          const lut = volume.getHistogram(j).lutGenerator_percentiles(0.5, 0.998);
+          const hmin = volume.getHistogram(j).findBinOfPercentile(0.5);
+          const hmax = volume.getHistogram(j).findBinOfPercentile(0.983);
+          const lut = new Lut().createFromMinMax(hmin, hmax);
           volume.setLut(j, lut.lut);
           view3D.updateLuts(volume);
         };
@@ -672,7 +678,7 @@ function showChannelUI(volume: Volume) {
       colorizeEnabled: false,
       colorize: (function (j) {
         return function () {
-          const lut = volume.getHistogram(j).lutGenerator_labelColors();
+          const lut = new Lut().createLabelColors(volume.getHistogram(j));
           volume.setColorPalette(j, lut.lut);
           myState.channelGui[j].colorizeEnabled = !myState.channelGui[j].colorizeEnabled;
           if (myState.channelGui[j].colorizeEnabled) {
--- a/public/index.ts
+++ b/public/index.ts
@@ -635,14 +635,6 @@ function showChannelUI(volume: Volume) {
         return function () {
           const [hmin, hmax] = volume.getHistogram(j).findAutoIJBins();
           const lut = new Lut().createFromMinMax(hmin, hmax);
-          // TODO: get a proper transfer function editor
-          // const lut = { lut: makeColorGradient([
-          //     {offset:0, color:"black"},
-          //     {offset:0.2, color:"black"},
-          //     {offset:0.25, color:"red"},
-          //     {offset:0.5, color:"orange"},
-          //     {offset:1.0, color:"yellow"}])
-          // };
           volume.setLut(j, lut.lut);
           view3D.updateLuts(volume);
         };
@@ -783,7 +775,10 @@ function showChannelUI(volume: Volume) {
       .onChange(
         (function (j) {
           return function (value) {
-            volume.getChannel(j).lutGenerator_windowLevel(value, myState.channelGui[j].level);
+            const hwindow = value;
+            const hlevel = myState.channelGui[j].level;
+            const lut = new Lut().createFromWindowLevel(hwindow, hlevel);
+            volume.setLut(j, lut.lut);
             view3D.updateLuts(volume);
           };
         })(i)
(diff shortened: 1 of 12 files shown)


public/index.ts
```diff
@@ -632,7 +633,8 @@ function showChannelUI(volume: Volume) {
       // this doesn't give good results currently but is an example of a per-channel button callback
       autoIJ: (function (j) {
         return function () {
-          const lut = volume.getHistogram(j).lutGenerator_auto2();
+          const [hmin, hmax] = volume.getHistogram(j).findAutoIJBins();
+          const lut = new Lut().createFromMinMax(hmin, hmax);
```
toloudis, line 637: @frasercl This gets slightly more verbose because it was decoupled.  Many of the Luts can be created in these two steps: Get info from histogram, and then pass it to the minmax constructor.  But a couple of them actually do more complicated stuff and for those I just pass the whole Histogram in to the Lut construction function. (createLabelColors and createFromEqHistogram).  This change necessitates changes up at the w3cv level too.  Arguably one could just pass the whole histogram to all the Lut creation funcs instead.

src/Lut.ts
```diff
@@ -0,0 +1,500 @@
...
+   *  |            /
+   *  |           /
+   *  |          /
+   * 0+=========+---------------+-----
+   *  0         b    e         255
+   * @return {Lut}
+   * @param {number} b
+   * @param {number} e
+   */
+  createFromMinMax(b: number, e: number): Lut {
```
toloudis, line 87: most of the create calls modify and return `this` so that you can call `lut = new Lut().createFunc(params);` in one line.

src/Lut.ts
```diff
@@ -0,0 +1,500 @@
...
+function lerp(xmin, xmax, a) {
+  return a * (xmax - xmin) + xmin;
+}
+
+// We have an intensity value that is in the range of valueMin to valueMax.
+// This domain is assumed to have been remapped from oldMin to oldMax.
+// We now wish to find the intensity value that corresponds to the same relative position in the new domain of newMin to newMax.
+// For our Luts valueMin will always be 0, and valueMax will always be 255.
+// oldMin and oldMax will be the domain of the original raw data intensities.
+// newMin and newMax will be the domain of the new raw data intensities.
```
interim17, line 17: Very helpful explanation!

src/Lut.ts
```diff
@@ -0,0 +1,500 @@
...
+   * If e === b, then we use a step function with f(b) = 0 and f(b + 1) = 1
+   *  |
+   * 1|               +---------+-----
+   *  |              /
+   *  |             /
+   *  |            /
+   *  |           /
+   *  |          /
+   * 0+=========+---------------+-----
+   *  0         b    e         255
```
interim17, line 82:
```
  ██████╗ ██████╗  ██████╗ ██╗          ██████╗ ██████╗  █████╗ ██████╗ ██╗  ██╗
 ██╔════╝██╔═══██╗██╔═══██╗██║         ██╔════╝ ██╔══██╗██╔══██╗██╔══██╗██║  ██║
 ██║     ██║   ██║██║   ██║██║         ██║  ███╗██████╔╝███████║██████╔╝███████║
 ██║     ██║   ██║██║   ██║██║         ██║   ██║██╔══██╗██╔══██║██╔═══╝ ██╔══██║
 ╚██████╗╚██████╔╝╚██████╔╝███████╗    ╚██████╔╝██║  ██║██║  ██║██║     ██║  ██║
  ╚═════╝ ╚═════╝  ╚═════╝ ╚══════╝     ╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═╝╚═╝     ╚═╝  ╚═╝

File: src/Lut.ts [hunk 5]
```diff
@@ -0,0 +1,500 @@
...
+
+    // Generate 2 to 4 control points for a minMax LUT, from left to right
+    const controlPoints: ControlPoint[] = [];
+
+    // Add starting point at x = 0
+    let startVal = 0;
+    if (b < 0) {
+      startVal = -b / (e - b);
+    }
+    controlPoints.push({ x: 0, opacity: startVal, color: [255, 255, 255] });
```
interim17, line 139: For my understanding: if `b` is negative -10 and e is 20, the starting value is the ratio 10/30, so the opacity of the first control point is 0.33. So assuming that `e` is inside the range normalized to 255, this lut will range from 30-100% opacity?

If an input value is less than -10 will it get 0.33 for opacity? Or will it get 0?

interim17 (approved):
Going to keep reading on this but it definitely takes a bit of chewing for me. I don't see anything that seems off, or would prevent approval after reading through it once, and the branch seems to be running fine. Will be curious to read comments from @frasercl or others who understand this better than I do.

src/Lut.ts
toloudis: You have it right.  If b is `-10` and e is `20`, then the lookup table will start at 0.33 at x=0, rise up linearly to 1.0 at x=20, and then stay flat at 1.0 all the way to x=255.

Input values to the lookup table are clamped to 0..255.  So it will never sample outside of those bounds, so you'll always get numbers from 0.33 up to 1.0.

src/Lut.ts
toloudis: Haha sorry to say, I didn't just add it in this go-round.  This whole PR is mostly shuffling code out of Histogram and into Lut.  It's been sitting there all along!  But yeah I love that graph too... 📈

src/Lut.ts
```diff
@@ -0,0 +1,500 @@
...
+
+/**
+ * @typedef {Object} Lut Used for rendering.
+ * @property {Array.<number>} lut LUT_ARRAY_LENGTH element lookup table as array
+ * (maps scalar intensity to a rgb color plus alpha, with each value from 0-255)
+ * @property {Array.<ControlPoint>} controlPoints
+ */
+export class Lut {
+  public lut: Uint8Array;
+  public controlPoints: ControlPoint[];
```
frasercl, line 63: Having two representations of basically the same information in here makes me a bit nervous, but it seems to be responsibly enough managed in the body of the class and does have enough benefit to make sense.

frasercl (approved):
My comments are mostly about other comments - this looks good otherwise! Definitely like this structure better, and having lut-creation functions separate from certain methods to preserve on `Histogram` splits the work of reading and understanding this part of the codebase into two much more manageable chunks.

react-example/VolumeViewer.jsx
```diff
@@ -31,8 +31,11 @@ export class VolumeViewer extends React.Component {
                     name: `${AICS_CELL_URL}/${img.name}`,
                 }));
                 VolumeLoader.loadVolumeAtlasData(aimg, jsondata.images, (url, channelIndex) => {
-                    aimg.channels[channelIndex].lutGenerator_percentiles(0.5, 0.998);
-
+                    const hmin = aimg.getHistogram(channelIndex).findBinOfPercentile(0.5);
+                    const hmax = aimg.getHistogram(channelIndex).findBinOfPercentile(0.983);
+                    const lut = new Lut().createFromMinMax(hmin, hmax);
```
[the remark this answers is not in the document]
toloudis: Yeah, I don't know if anyone ever tests or uses the react-example.

src/Lut.ts
```diff
@@ -0,0 +1,500 @@
...
+  }
+
+  /**
+   * Generate a lookup table with a different color per intensity value
+   * @return {Lut}
+   */
+  createLabelColors(histogram: Histogram): Lut {
+    const lut = new Uint8Array(LUT_ARRAY_LENGTH).fill(0);
+    // TODO specify type for control point
+    const controlPoints: ControlPoint[] = [];
```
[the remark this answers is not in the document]
toloudis: hm I think that TODO is done

src/Lut.ts
```diff
@@ -0,0 +1,500 @@
...
+      // just reset to whole range in this case...?
+      return this.createFullRange();
+    }
+  }
+
+  /**
+   * Generate a lookup table with a different color per intensity value
+   * @return {Lut}
+   */
+  createLabelColors(histogram: Histogram): Lut {
```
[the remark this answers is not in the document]
toloudis: Ok yeah I can clarify that

src/Lut.ts
toloudis: I would think of the `lut` as a cached version of the data array that the `controlPoints` generate.  It's easy to go from controlPoints to lut, but hard to go the other way.  However, control points to lut is not a trivial operation so I keep the lut array around.  We can think about the implications of removing that...  Agave stores even higher level representations like pctMin/pctMax, window/level, isovalue, in addition to control points, based on how the user configured the lut.

src/Histogram.ts
```diff
@@ -378,14 +118,11 @@ export default class Histogram {
...
-    return this.lutGenerator_minMax(hmin, hmax);
+    return [hmin, hmax];
   }
 
-  /**
-   * Generate a lookup table attempting to replicate ImageJ's "Auto" button
-   * @return {Lut}
-   */
-  lutGenerator_auto2(): Lut {
+  // Generate a lookup table attempting to replicate ImageJ's "Auto" button
```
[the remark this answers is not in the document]
toloudis: Oops yes I will fix those.

PR #195, opened 2024-03-20
Multiresolution loading options in tweakpane

### LINKED ISSUE #184: tweakpane option to use automatic resolution level or explicit
- checkbox for "auto select resolution level" vs "explicit choice of resolution level"
- numeric input for explicit resolution level choice

frasercl:
Review time: pretty quick (15min)

Resolves #184: exposes settings for picking which multiscale level to load via tweakpane.

The hidden tweakpane now contains the following options, under a folder labeled `Multiscale loading`:

- `multiscaleLevel` sets the *minimum* scale level that may be loaded. (Note that higher-indexed scale levels are more downsampled, and therefore *lower* resolution.) Loaders may still pick a higher (more downsampled) multiscale index based on the other settings below. Therefore, leaving this property at its default of 0 effectively disables enforcing a resolution level.
- `maxAtlasEdge` modifies the maximum atlas dimension used by the loader internally when deciding which scale level is appropriate to load. When not constrained by `multiscaleLevel`, the loader picks the highest-resolution scale level that can be packed into a texture atlas with sides no longer than `maxAtlasEdge`.
- `scaleLevelBias` allows nudging the scale level index determined based on `maxAtlasEdge`. E.g. if the loader determines (as described above) that level 1 is the least downsampled scale level it can load, but `scaleLevelBias` is 1, the loader will instead load level 2.

Internally, tweakpane modifies properties with the same names on `Volume.loadSpecRequired`, which represents the data requested for display based on current settings. It does this via `Volume.updateRequiredData`, which makes changes to `loadSpecRequired`, then checks if the changes indicate that new data needs to be loaded. Only `multiscaleLevel` existed in the `LoadSpec` type previously. This PR adds the other two, hooks them up to the `OMEZarrLoader` internals, and properly handles them in `updateRequiredData`'s reload check.

While I was testing, I found a bug that was causing the loader to unconditionally pick the highest available scale level in Z-slice mode, regardless of the value of `maxAtlasEdge`. So I fixed that too. This accounts for the changes to Atlas2DSlice.ts and VolumeDrawable.ts.

### Steps to test:

- `npm start` and open the viewer.
- Ctrl + Alt + 1 to reveal the hidden tweakpane.
- The options listed above should be present and work as described when a zarr volume is open.

### DIFF
--- a/src/Atlas2DSlice.ts
+++ b/src/Atlas2DSlice.ts
@@ -178,9 +178,10 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
 
       const sliceInBounds = this.updateSlice();
       if (sliceInBounds) {
-        const sliceRatio = Math.floor(this.settings.zSlice) / this.volume.imageInfo.volumeSize.z;
+        const sliceLowerBound = Math.floor(this.settings.zSlice) / this.volume.imageInfo.volumeSize.z;
+        const sliceUpperBound = (Math.floor(this.settings.zSlice) + 1) / this.volume.imageInfo.volumeSize.z;
         this.volume.updateRequiredData({
-          subregion: new Box3(new Vector3(0, 0, sliceRatio), new Vector3(1, 1, sliceRatio)),
+          subregion: new Box3(new Vector3(0, 0, sliceLowerBound), new Vector3(1, 1, sliceUpperBound)),
         });
       }
     }
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -242,6 +242,7 @@ export default class Volume {
 
   /** Call on any state update that may require new data to be loaded (subregion, enabled channels, time, etc.) */
   async updateRequiredData(required: Partial<LoadSpec>, onChannelLoaded?: PerChannelCallback): Promise<void> {
+    console.log("here");
     this.loadSpecRequired = { ...this.loadSpecRequired, ...required };
     let noReload =
       this.loadSpec.time === this.loadSpecRequired.time &&
@@ -254,19 +255,24 @@ export default class Volume {
     if (
       noReload &&
       (!this.loadSpec.subregion.equals(this.loadSpecRequired.subregion) ||
-        this.loadSpecRequired.maxAtlasEdge !== this.loadSpec.maxAtlasEdge)
+        this.loadSpecRequired.maxAtlasEdge !== this.loadSpec.maxAtlasEdge ||
+        this.loadSpecRequired.multiscaleLevel !== this.loadSpec.multiscaleLevel)
     ) {
       const currentScale = this.imageInfo.multiscaleLevel;
-      // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
-      const minLevel = this.loadSpec.multiscaleLevel ?? 0;
       // Loaders should cache loaded dimensions so that this call blocks no more than once per valid `LoadSpec`.
       const dims = await this.loader?.loadDims(this.loadSpecRequired);
       if (dims) {
+        // `LoadSpec.scaleLevelBias`, if specified, nudges the scale level we pick up or down.
+        const levelBias = this.loadSpecRequired.scaleLevelBias ?? 0;
+        // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
+        const minLevel = this.loadSpecRequired.multiscaleLevel ?? 0;
+
         const dimsZYX = dims.map(({ shape }): [number, number, number] => [shape[2], shape[3], shape[4]]);
         const optimalLevel = estimateLevelForAtlas(dimsZYX, this.loadSpecRequired.maxAtlasEdge);
-        const levelBias = this.loadSpecRequired.scaleLevelBias ?? 0;
-        const loadableLevel = Math.max(0, Math.min(dims.length - 1, optimalLevel + levelBias));
-        noReload = currentScale <= Math.max(loadableLevel, minLevel);
+
+        const levelToLoad = Math.max(optimalLevel + levelBias, minLevel);
+        const clampedLevelToLoad = Math.max(0, Math.min(dims.length - 1, levelToLoad));
+        noReload = currentScale === clampedLevelToLoad;
       }
     }
 
@@ -277,8 +283,6 @@ export default class Volume {
       this.loadSpec = {
         ...this.loadSpecRequired,
         subregion: this.loadSpecRequired.subregion.clone(),
-        // preserve multiscale option from original `LoadSpec`, if any
-        multiscaleLevel: this.loadSpec.multiscaleLevel,
       };
       this.loader?.loadVolumeData(this, undefined, onChannelLoaded);
     }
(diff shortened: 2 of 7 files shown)


src/Atlas2DSlice.ts
```diff
@@ -178,9 +178,10 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
 
       const sliceInBounds = this.updateSlice();
       if (sliceInBounds) {
-        const sliceRatio = Math.floor(this.settings.zSlice) / this.volume.imageInfo.volumeSize.z;
+        const sliceLowerBound = Math.floor(this.settings.zSlice) / this.volume.imageInfo.volumeSize.z;
+        const sliceUpperBound = (Math.floor(this.settings.zSlice) + 1) / this.volume.imageInfo.volumeSize.z;
         this.volume.updateRequiredData({
-          subregion: new Box3(new Vector3(0, 0, sliceRatio), new Vector3(1, 1, sliceRatio)),
+          subregion: new Box3(new Vector3(0, 0, sliceLowerBound), new Vector3(1, 1, sliceUpperBound)),
```
toloudis, line 184: interesting, I thought we had special handling for when you used the same z value

src/Volume.ts
```diff
@@ -244,19 +246,32 @@ export default class Volume {
...
+        const levelBias = this.loadSpecRequired.scaleLevelBias ?? 0;
+        // `LoadSpec.multiscaleLevel`, if specified, forces a cap on the scale level we can load.
+        const minLevel = this.loadSpecRequired.multiscaleLevel ?? 0;
+
+        const dimsZYX = dims.map(({ shape }): [number, number, number] => [shape[2], shape[3], shape[4]]);
+        const optimalLevel = estimateLevelForAtlas(dimsZYX, this.loadSpecRequired.maxAtlasEdge);
+
+        const levelToLoad = Math.max(optimalLevel + levelBias, minLevel);
+        const clampedLevelToLoad = Math.max(0, Math.min(dims.length - 1, levelToLoad));
+        noReload = currentScale === clampedLevelToLoad;
```
toloudis, line 274: now that you calculated clampedLevelToLoad, I am not seeing where it's being used to actually load anything.

toloudis: I think I see the same computation in zarr_utils... maybe they could be factored into a simple utils function.

src/loaders/IVolumeLoader.ts
```diff
@@ -6,6 +6,14 @@ import { PrefetchDirection } from "./zarr_utils/types.js";
 
 export class LoadSpec {
   time = 0;
+  /** The max size of a volume atlas that may be produced by a load. Used to pick the appropriate multiscale level. */
+  maxAtlasEdge?: number;
```
toloudis, line 10: Does this affect the actual texture size?

src/Atlas2DSlice.ts
frasercl: There is within the loader, but this change addresses a bug in how scale levels are picked. If min and max Z are the same when picking scale levels, we ended up calculating the required tile size from X and Y, then searching for a scale level that can fit 0 of those tiles. Perhaps the better way to handle this would be enforcing a maximum of 1 when doing that calculation?

src/Volume.ts
```diff
@@ -244,19 +246,32 @@ export default class Volume {
...
 
     // An update to `subregion` should trigger a reload when the new subregion is not contained in the old one
     // OR when the new subregion is smaller than the old one by enough that we can load a higher scale level.
-    if (noReload && !this.loadSpec.subregion.equals(this.loadSpecRequired.subregion)) {
+    if (
+      noReload &&
+      (!this.loadSpec.subregion.equals(this.loadSpecRequired.subregion) ||
+        this.loadSpecRequired.maxAtlasEdge !== this.loadSpec.maxAtlasEdge ||
+        this.loadSpecRequired.multiscaleLevel !== this.loadSpec.multiscaleLevel)
+    ) {
```
meganrm, line 259: when there are still many conditionals I prefer to pull those out into a function whose name tells you what you're ultimately checking for so this if statement is easier to read

src/Volume.ts
```diff
@@ -244,19 +246,32 @@ export default class Volume {
     let noReload =
       this.loadSpec.time === this.loadSpecRequired.time &&
       this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion) &&
-      this.loadSpecRequired.channels.every((channel) => this.loadSpec.channels.includes(channel));
+      this.loadSpecRequired.channels.every((channel) => this.loadSpec.channels.includes(channel)) &&
+      this.loadSpecRequired.scaleLevelBias === this.loadSpec.scaleLevelBias;
```
meganrm, line 250: Kinda same comment as below: maybe it's the GitHub wrapping here, but I'm having trouble following what this is doing

src/Volume.ts
```diff
@@ -244,19 +246,32 @@ export default class Volume {
     let noReload =
```
meganrm, line 246: this is just a readability comment: after reading through this, I think I would divide this into a function that handles all the required checks, including the call to get the `dims` and then an update function that just does lines 264- 274

src/Volume.ts
frasercl: reorganized where this kind of scale level resolution happens pretty significantly, to hopefully be more general

src/Volume.ts
frasercl: Based on your comments and Dan's above, I broke out a bunch of functionality from this function into new ones. The new `updateRequiredData` should be significantly slimmer and more comprehensible.

src/Atlas2DSlice.ts
frasercl: New changes do the min-checking I describe above, though I'm on the fence about whether to revert this change.

toloudis: I figure both are good

toloudis: The one odd thing here is that it's hard to tell whether the caller intends zSlice or zSlice+1, just by looking at the bounds. You almost want to do zSlice-0.5, zSlice+0.5 .  I don't know if that bias is commented anywhere.

src/Volume.ts
```diff
@@ -238,42 +240,63 @@ export default class Volume {
...
+    return (
+      this.loadSpec.time !== this.loadSpecRequired.time || // time point changed
+      !this.loadSpec.subregion.containsBox(this.loadSpecRequired.subregion) || // new subregion not contained in old
+      this.loadSpecRequired.channels.some((channel) => !this.loadSpec.channels.includes(channel)) // new channel(s)
+    );
+  }
+
+  /**
+   * Returns `true` iff differences between `loadSpec` and `loadSpecRequired` indicate a new load *may* get a
+   * different scale level than is currently loaded.
```
toloudis, line 254: Can you say why this is a "maybe" and not "definitely will load new scale level?"

meganrm (approved):
As someone less familiar with all the functionality, this is way easier to read after those more recent commits

toloudis (approved):
One small comment about "may load" vs "will load" but otherwise this is looking good.  I also like the encapsulation into those smaller functions

src/Volume.ts
frasercl: Updated JSDoc comment here. It says:

> This checks for changes in properties that *can*, but do not *always*, change the scale level the loader picks. For example, a smaller subregion *may* mean a higher scale level will fit within memory constraints, or it may not. A higher `scaleLevelBias` *may* nudge the volume into a higher scale level, or we may already be at the max imposed by `multiscaleLevel`.

PR #198, opened 2024-03-28
Channel object now uses the Lut object

toloudis:
Estimated time to review: <15 min.

Following up on a previous refactor PR, we now use the results.  
Channels were previously storing a lookup table as a simple Uint8Array.  Now we have an encapsulated class called Lut. 
Modified Channel to use it, fixed the remapping code to include both ControlPoints and array, and tested the results of the remapping in the test app.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -636,7 +636,7 @@ function showChannelUI(volume: Volume) {
         return function () {
           const [hmin, hmax] = volume.getHistogram(j).findAutoIJBins();
           const lut = new Lut().createFromMinMax(hmin, hmax);
-          volume.setLut(j, lut.lut);
+          volume.setLut(j, lut);
           view3D.updateLuts(volume);
         };
       })(i),
@@ -645,7 +645,7 @@ function showChannelUI(volume: Volume) {
         return function () {
           const [b, e] = volume.getHistogram(j).findAutoMinMax();
           const lut = new Lut().createFromMinMax(b, e);
-          volume.setLut(j, lut.lut);
+          volume.setLut(j, lut);
           view3D.updateLuts(volume);
         };
       })(i),
@@ -654,7 +654,7 @@ function showChannelUI(volume: Volume) {
         return function () {
           const [hmin, hmax] = volume.getHistogram(j).findBestFitBins();
           const lut = new Lut().createFromMinMax(hmin, hmax);
-          volume.setLut(j, lut.lut);
+          volume.setLut(j, lut);
           view3D.updateLuts(volume);
         };
       })(i),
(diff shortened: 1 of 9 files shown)


public/index.ts
```diff
@@ -880,10 +880,7 @@ function loadImageData(jsonData: ImageInfo, volumeData: Uint8Array[]) {
 function onChannelDataArrived(v: Volume, channelIndex: number) {
   const currentVol = v; // myState.volume;
 
-  const hmin = currentVol.getHistogram(channelIndex).findBinOfPercentile(0.5);
-  const hmax = currentVol.getHistogram(channelIndex).findBinOfPercentile(0.983);
-  const lut = new Lut().createFromMinMax(hmin, hmax);
-  currentVol.setLut(channelIndex, lut.lut);
```
toloudis, line 886: removing this code allows the automatic Lut remapping to happen in the test app.  (Happens before this callback is fired)

src/Channel.ts
```diff
@@ -96,10 +95,17 @@ export default class Channel {
...
-    const newLut = remapLut(this.lut, this.rawMin, this.rawMax, min, max);
+    // If either of the min/max ranges are both zero, then we have undefined behavior and should
+    // not remap the lut.  This situation can happen at first load, for example,
+    // when one channel has arrived but others haven't.
+    if (!(this.rawMin === 0 && this.rawMax === 0) && !(min === 0 && max === 0)) {
+      const newLut = remapLut(this.lut.lut, this.rawMin, this.rawMax, min, max);
+      const newControlPoints = remapControlPoints(this.lut.controlPoints, this.rawMin, this.rawMax, min, max);
+      this.lut.lut = newLut;
+      this.lut.controlPoints = newControlPoints;
+    }
```
frasercl, line 106: could you use `Lut.remapDomains` here?

frasercl (approved):

src/Channel.ts
toloudis: Yup, thanks. I'm making that change.

src/Channel.ts
```diff
@@ -96,10 +96,14 @@ export default class Channel {
 
   public setRawDataRange(min: number, max: number): void {
     // remap the lut which was based on rawMin and rawMax to new min and max
-    const newLut = remapLut(this.lut, this.rawMin, this.rawMax, min, max);
+    // If either of the min/max ranges are both zero, then we have undefined behavior and should
+    // not remap the lut.  This situation can happen at first load, for example,
+    // when one channel has arrived but others haven't.
+    if (!(this.rawMin === 0 && this.rawMax === 0) && !(min === 0 && max === 0)) {
+      this.lut.remapDomains(this.rawMin, this.rawMax, min, max);
+    }
```
ShrimpCryptid, line 104: Should this check for `this.rawMin === this.rawMax` or `min === max` instead of zero values? I think from digging through the code this would cause a divide-by-zero in `Lut.remapDomain`.

toloudis: both being zero is a possible indicator that no data has arrived yet. but nonzero and the same means all pixels are only that intensity value. So probably should handle the div by zero situation by adding 1 or something.  I will post this as an issue for a subsequent PR.

ShrimpCryptid (approved):

PR #201, opened 2024-04-02
Fix/update-isosurfaces

frasercl:
Closes allen-cell-animated/website-3d-cell-viewer#189, resolving two related problems with isosurfaces:

1. The visibility of a channel's isosurface doesn't count towards its being "required" to load. I.e., if a channel is disabled in the volume but its isosurface is enabled, the channel is still marked as "not required" and its data will not be reloaded when some event requires it (time change, switch to single-slice mode, etc.).
2. Even while a channel is considered "required," isosurfaces aren't regenerated when new data is loaded, notably while playing through time.

While working on this change, I discovered that the viewer exposed two distinct code paths for changing isosurface settings. One was a set of basically stateless methods (`createIsosurface`, `updateIsosurface`, etc.) and the other was via the newer `setChannelOptions`. The example app in this repo uses the first method, while website-3d-cell-viewer uses the second. Since for a number of reasons we now prefer to use a single state-updating method in cases like this, I removed the first code path and updated the test app. So now I get to submit everyone's favorite kind of pull request: one with a net negative effect on the size of the codebase!

### DIFF
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -171,15 +171,9 @@ export default class VolumeDrawable {
           this.destroyIsosurface(channelIndex);
         } else if (!hasIso && options.isosurfaceEnabled) {
           // 127 is half of the intensity range 0..255
-          let isovalue = 127;
-          if (options.isovalue !== undefined) {
-            isovalue = options.isovalue;
-          }
+          const isovalue = options.isovalue ?? 127;
           // 1.0 is fully opaque
-          let isosurfaceOpacity = 1.0;
-          if (options.isosurfaceOpacity !== undefined) {
-            isosurfaceOpacity = options.isosurfaceOpacity;
-          }
+          const isosurfaceOpacity = options.isosurfaceOpacity ?? 1.0;
           this.createIsosurface(channelIndex, isovalue, isosurfaceOpacity, isosurfaceOpacity < 1.0);
         }
         this.updateChannelDataRequired(channelIndex);
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -470,11 +470,16 @@ export default class VolumeDrawable {
   }
 
   onChannelLoaded(batch: number[]): void {
-    this.meshVolume.onChannelData(batch);
-
     for (let j = 0; j < batch.length; ++j) {
       const idx = batch[j];
-      this.setChannelOptions(idx, this.channelOptions[idx]);
+      const channelOptions = this.channelOptions[idx];
+      // TODO: this is a relatively crude way to ensure that channel settings are synced up when volume data is loaded.
+      //    Can we instead audit which settings updated by `setChannelOptions` actually need to be reset on load?
+      this.setChannelOptions(idx, channelOptions);
+      if (channelOptions.isosurfaceEnabled) {
+        this.destroyIsosurface(idx);
+        this.createIsosurface(idx, channelOptions.isovalue, channelOptions.isosurfaceOpacity);
+      }
     }
 
     // let the outside world have a chance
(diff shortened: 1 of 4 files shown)


src/VolumeDrawable.ts
```diff
@@ -169,33 +169,47 @@ export default class VolumeDrawable {
...
-            isovalue = options.isovalue;
-          }
-          // 1.0 is fully opaque
-          let isosurfaceOpacity = 1.0;
-          if (options.isosurfaceOpacity !== undefined) {
-            isosurfaceOpacity = options.isosurfaceOpacity;
-          }
-          this.createIsosurface(channelIndex, isovalue, isosurfaceOpacity, isosurfaceOpacity < 1.0);
+        } else if (!hasIso && options.isosurfaceEnabled && this.volume.channels[channelIndex].loaded) {
+          const { isovalue, isosurfaceOpacity } = options;
```
ShrimpCryptid, line 173: Are `isovalue` and `isosurfaceOpacity` now guaranteed to be defined?

src/VolumeDrawable.ts
ShrimpCryptid, line 1: Nit: I know this PR is cleaner like this but it's still tripping me up that methods are called before they're declared... JavaScript moment

src/VolumeDrawable.ts
```diff
@@ -169,33 +169,47 @@ export default class VolumeDrawable {
...
+        // This channel is currently marked required, but both its volume and isosurface are off. Remove it!
+        this.volume.updateRequiredData({ channels: requiredChannels.filter((i) => i !== channelIndex) });
+      }
+    } else {
+      if (channelIsRequired) {
+        // This channel is not marked required, but either its volume or isosurface is on. Add it!
+        this.volume.updateRequiredData({ channels: [...requiredChannels, channelIndex] });
       }
     }
   }
```
ShrimpCryptid, line 215: I know this is a simple check but it's very readable! I did something similar in another PR and now I want to go back and restructure it to look like this.

src/VolumeDrawable.ts
```diff
@@ -450,17 +470,20 @@ export default class VolumeDrawable {
...
 
   onChannelLoaded(batch: number[]): void {
-    this.meshVolume.onChannelData(batch);
```
ShrimpCryptid, line 453: What did this line do before?

ShrimpCryptid (commented):
Left some question and comments! Overall looks good, just want to get more familiar with the codebase!

src/VolumeDrawable.ts
frasercl: See the changes to MeshVolume.ts, where I removed this function. It was doing something similar to the block of added code below - updating an isosurface on data load - but was written with an older version of the app in mind, where a volume receives a complete set of loaded data once and never changes again. I decided to move those updates up to this function rather than rework the existing method because `MeshVolume` only has access to the enabled/disabled state of an isosurface based on whether or not it has a three.js mesh object for it, which doesn't perfectly correspond to whether the isosurface is actually visible.

src/VolumeDrawable.ts
frasercl: They now have defaults applied when `createIsosurface` is called.

...I'm realizing now that this was a choice I made before removing the specialized methods on `View3d` for updating isosurfaces. When `createIsosurface` was also called by methods on `View3d`, it made sense to keep defaults consistent at the first place where those two code paths merged. Now, that method definitely can be `private`-ized and could maybe be removed entirely (it's just one line)

rugeli (approved):
It's initially took me a bit to understand how removed functions(like `onChannelData`) were managed. And then I found the answers as I scrolled through the conversation. Thanks to the detailed pr description and comments, they really helped my comprehension!

ShrimpCryptid (approved):
Thanks for answering my questions! Leaving it up to you if you want to make the change you mentioned in one of your replies.

src/VolumeDrawable.ts
frasercl: Reordered the method based on this comment because I could

PR #202, opened 2024-04-05
allow lut controlpoints to be out of our 0-255 range; simplify remapping

toloudis:
Context:
When playing through time series data, we have to adjust our lookup table to account for intensity ranges changing in the data.  However, the way we were doing it, there was too much clipping at the edges of our accepted 0-255 domain for the control points.  In the case of consecutive remappings, information out of the bounds was being lost by this clipping and things would quickly look terrible - often just going to complete black or complete white.

This PR stops trying to fix up edge cases when we have to remap lookup tables during time series playback.  

Instead the plan is to simply allow control points to be out of range, and build the lut array from the control points, correctly handling the out of range values.

See companion PR in website-3d-cell-viewer to properly take advantage of this. (https://github.com/allen-cell-animated/website-3d-cell-viewer/pull/210)

### DIFF
--- a/src/Lut.ts
+++ b/src/Lut.ts
@@ -259,10 +259,9 @@ export class Lut {
     let color1 = controlPointToRGBA(c1);
     let lastIndex = 1;
     let a = 0;
-    // if the first control point is after 0, act like there are 0s going all the way up to it.
-    // or lerp up to the first point?
-    for (let x = c0.x; x < 256; ++x) {
-      while (x > c1.x) {
+    for (let i = 0; i < 256; ++i) {
+      // find the two control points that i is between
+      while (i > c1.x) {
         // advance control points
         c0 = c1;
         color0 = color1;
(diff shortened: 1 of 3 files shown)


toloudis:
there is still an edge case with respect to remapping that can happen when two controlpoints are remapped to be too close to each other and fall into the same bin... to be handled in a future PR.

src/Lut.ts
```diff
@@ -475,84 +478,6 @@ export function remapControlPoints(
...
-      // otherwise, add a point at 255.
-      resultControlPoints.push({
-        x: 255,
-        opacity: resultControlPoints[resultControlPoints.length - 1].opacity,
-        color: resultControlPoints[resultControlPoints.length - 1].color,
-      });
-    }
-  }
-  return resultControlPoints;
+  return newControlPoints;
```
rugeli, line 482: Removing this big chunk of range-handling logic and returning `newControlPoints` instead looks very clean!

src/Lut.ts
```diff
@@ -464,7 +466,8 @@ export function remapControlPoints(
 
   // remap all cp x values.
   // interpolate all new colors and opacities
-  // then see if we need to clip?
+  // Do not clip values outside of 0-255. This is important to
```
rugeli, line 469: I've noticed that the removed code addressed control points outside the 0-255 range, and it appears we're moving away from clipping these values. Can you share more about this change?

ShrimpCryptid (approved):
LGTM!

src/Lut.ts
toloudis: The control points array is used to generate a densely sampled lookup table.  The lookup table is in the form of an array of result values.  The input values to the lookup table can not go outside of the 0-255 domain (strict array bounds).  The control points are just defining vertices of a piecewise linear function, and the lookup table is also representing the same function.   

If we clip the control points to the 0-255 domain when remapping them, then we lose information when the remapping would extend outside the bounds.  This is because they might be remapped AGAIN and return within bounds again.    It's ok for the control point values to be out of bounds as long as we build the lookup table array correctly.    I can explain this more in person if you'd like.

toloudis:
The edge case which I mentioned above was a time series that was having very anomalous remapping when time playback cycled back to zero.   I discovered that the "edge case" was due to bad data: a timelapse that reports having 121 times but only data for the first 120 of them.  The 121st time was returning data with min,max = 0,0 and a zero-filled array. I added one commit here that tries to ignore this scenario and doesn't reset the channel min and max to zeroes.    Our code was already gracefully continuing from this scenario but with bad lookup table.
There's probably a better fix down the road for handling this situation but this makes it work with the test data.

src/Lut.ts
rugeli: Thank you for taking the time to explain how the control points and lookup table interact regarding the bounds, it makes so much sense!

rugeli (approved):
Lgtm!

PR #203, opened 2024-04-08
Tweakpane fixes

### LINKED ISSUE #143: tweakpane leaks memory

### LINKED ISSUE #156: Don't rebuild tweakpane unless it is open
## Description

A significant portion of the work done on volume load is related to Tweakpane, which most users will never open. Now that we expect to do lots of volume loads, it's worth making that process fast.

## Expected Behavior

Tweakpane should not take up so much time on volume loads if it isn't being used.

## Reproduction

Switch to a time series volume, start recording in the devtools profiler, move the time slider, and check out how much load time is taken up by Tweakpane.

frasercl:
Review time: really little (5-10min)

Resolves #143 and resolves #156: use tweakpane more responsibly.

### How tweakpane is used currently:

- The user may press Ctrl+Alt+1 to open the tweakpane. Internally, the tweakpane element already existed (built on construction of `View3d`), and this action just reveals it with CSS.
- On every data load, the tweakpane is fully rebuilt.

### How tweakpane is used after this PR:

- The tweakpane does not exist before it is opened - it is (re)built whenever the user toggles it to visible, and destroyed when they toggle to invisible.
- On every data load, we call the pane's `refresh` method.
- We destroy the tweakpane with its own `dispose` method, rather than just unmounting it in the DOM.

I believe the original strategy of rebuilding the tweakpane on data loads was to anticipate when objects were replaced and rebind everything to new objects. But that seems to be pretty rare - it should currently just happen on switching to an entirely new volume (though other, deeper objects we could bind controls to in the future, e.g. `Channel` or `MeshVolume`, may be recreated more frequently and cause confusion). Hopefully a user who knows to open the tweakpane can deal with reopening it when a desync occurs. I'd still be grateful for a bit of extra testing here though. In particular, @toloudis I'm seeing different and better numbers in the heap profiler with this change versus the current state, but I'd be interested in a bit of verification since you did the tests that prompted #143 in the first place.

### DIFF
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -55,8 +55,7 @@ export class View3d {
   private reflectedLight: DirectionalLight;
   private fillLight: DirectionalLight;
 
-  private tweakpane: Pane;
-  private tweakpaneOpen: boolean;
+  private tweakpane: Pane | null;
 
   /**
    * @param {Object} options Optional options.
@@ -86,8 +85,7 @@ export class View3d {
     this.fillLight = new DirectionalLight();
     this.buildScene();
 
-    this.tweakpane = this.setupGui(this.canvas3d.containerdiv);
-    this.tweakpaneOpen = false;
+    this.tweakpane = null;
     window.addEventListener("keydown", this.handleKeydown);
   }
 
@@ -220,8 +218,8 @@ export class View3d {
   onVolumeData(volume: Volume, channels: number[]): void {
     this.image?.updateScale();
     this.image?.onChannelLoaded(channels);
-    if (volume.isLoaded()) {
-      this.tweakpane = this.setupGui(this.canvas3d.containerdiv);
+    if (volume.isLoaded() && this.tweakpane) {
+      this.tweakpane.refresh();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/View3d.ts
```diff
@@ -220,8 +218,8 @@ export class View3d {
   onVolumeData(volume: Volume, channels: number[]): void {
     this.image?.updateScale();
     this.image?.onChannelLoaded(channels);
-    if (volume.isLoaded()) {
-      this.tweakpane = this.setupGui(this.canvas3d.containerdiv);
+    if (volume.isLoaded() && this.tweakpane) {
+      this.tweakpane.refresh();
```
toloudis, line 222: The consideration here was that it's possible that new channels are arriving.  The current setupGui loops over all channels to build itself.  Not sure if refresh will pull the correct data.  But I think all the Channel objects are created prior to the first onVolumeData so this is probably good!

toloudis: I think I was confusing this with the separate dat.gui of the test viewer.  The internal setupGui doesn't loop over channels! (But someday it could).  So this is even safer than I was thinking.

frasercl: Yeah - I originally added some per-channel settings, but removed them in part because they relied on complicated workarounds. (See #109). Regardless, I'm pretty sure `Channel`s are still persistent for the entire life of a `Volume`, though `MeshVolume`s and other objects I'm not thinking of might not be.

toloudis (approved):

ascibisz (approved):
Not gonna lie, I had to google the word "tweakpane" because I hadn't heard of that before 😆 . It looks to me like your code does what you described, so I'll approve, but my understanding of this code is certainly limited!

PR #204, opened 2024-04-09
optimize data arriving at viewer's main thread

toloudis:
Estimated time to review:  10 min or less.

One of the most expensive operations when playing through time series is shuffling data around when it arrives at the viewer.
The following two profiles show 30 seconds of ome-zarr playback of one of our nucmorph colonies, playing in our testbed viewer.
In the following profile: `packToAtlas` is pretty expensive.

Before:
DevTools - localhost9021 2024-04-09 13-15-37

Turns out packToAtlas is iterating pixel by pixel over the data and reorganizing it into a different buffer. 
At the innermost part of the loop, we could actually do this with a block-wise array copy rather than iterating one pixel at a time.  
After this optimization, the profile shows packToAtlas consisting of a much smaller time slice!

After:
DevTools - localhost9021 2024-04-09 13-18-03

(You may notice that now `onChannelLoaded` is starting to dominate the time spent.  However, that computation is nearly entirely eliminated by #203 .)

### DIFF
--- a/src/Channel.ts
+++ b/src/Channel.ts
@@ -242,9 +242,10 @@ export default class Channel {
       tileoffset = tilex * x + tiley * y * atlasrow;
       for (let j = 0; j < y; ++j) {
         tilerowoffset = j * atlasrow;
-        for (let k = 0; k < x; ++k) {
-          volimgdata[tileoffset + tilerowoffset + k] = this.volumeData[i * (x * y) + j * x + k];
-        }
+        volimgdata.set(
+          this.volumeData.subarray(i * (x * y) + j * x, i * (x * y) + j * x + x),
+          tileoffset + tilerowoffset
+        );
       }
     }
 
--- a/src/Channel.ts
+++ b/src/Channel.ts
@@ -166,7 +166,8 @@ export default class Channel {
     let tilex = 0,
       tiley = 0,
       tileoffset = 0,
-      tilerowoffset = 0;
+      tilerowoffset = 0,
+      destOffset = 0;
     for (let i = 0; i < z; ++i) {
       // tile offset
       tilex = i % numXtiles;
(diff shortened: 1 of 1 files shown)


src/Channel.ts
```diff
@@ -174,9 +174,10 @@ export default class Channel {
       tileoffset = tilex * x + tiley * y * atlasrow;
       for (let j = 0; j < y; ++j) {
         tilerowoffset = j * atlasrow;
-        for (let k = 0; k < x; ++k) {
-          this.volumeData[i * (x * y) + j * x + k] = volimgdata[tileoffset + tilerowoffset + k];
-        }
+        this.volumeData.set(
+          volimgdata.subarray(tileoffset + tilerowoffset, tileoffset + tilerowoffset + x),
+          i * (x * y) + j * x
+        );
```
toloudis, line 180: this one is for when we load from our old atlased-png format and need to create a 3d volume from it.  This is the less-critical path because we aren't trying to animate playback of these.

src/Channel.ts
```diff
@@ -242,9 +243,10 @@ export default class Channel {
       tileoffset = tilex * x + tiley * y * atlasrow;
       for (let j = 0; j < y; ++j) {
         tilerowoffset = j * atlasrow;
-        for (let k = 0; k < x; ++k) {
-          volimgdata[tileoffset + tilerowoffset + k] = this.volumeData[i * (x * y) + j * x + k];
-        }
+        volimgdata.set(
+          this.volumeData.subarray(i * (x * y) + j * x, i * (x * y) + j * x + x),
+          tileoffset + tilerowoffset
+        );
```
toloudis, line 249: This is the code that turns volume arrays into a texture atlased 2d layout for rendering.  The ome-zarr code path goes through this function.

frasercl: could/should `i * (x * y) + j * x` be factored out to make the arguments to `subarray` something like `(offset, offset + x)`?

src/Channel.ts
frasercl: same question applies here I suppose, though it's just addition happening here

frasercl (approved):

src/Channel.ts
toloudis: I have no problem with that... I'll do it

meganrm (approved):

PR #205, opened 2024-04-17
error is now a warning

toloudis:
Time to review: 5 min

2 key fixes:

1. when matching data from different sources, we will consider the physical scale to be optional in favor of just matching the XYZ array dims.   This is important for our nucmorph deliverable in which they are having trouble applying the correct physical scale metadata to the segmentation zarrs.
2. The RawArrayLoader had some bad imports

### DIFF
--- a/src/loaders/zarr_utils/utils.ts
+++ b/src/loaders/zarr_utils/utils.ts
@@ -119,7 +119,7 @@ function compareZarrArraySize(
   }
 }
 
-const EPSILON = 0.0000001;
+const EPSILON = 0.00001;
 const aboutEquals = (a: number, b: number): boolean => Math.abs(a - b) < EPSILON;
 
 function scaleTransformsAreEqual(aSrc: ZarrSource, aLevel: number, bSrc: ZarrSource, bLevel: number): boolean {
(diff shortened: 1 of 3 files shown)


ShrimpCryptid (approved):

src/loaders/zarr_utils/utils.ts
```diff
@@ -168,7 +168,10 @@ export function matchSourceScaleLevels(sources: ZarrSource[]): void {
         // Now we know the arrays are equal, but they may still be invalid to match up because...
         // ...they have different scale transformations
         if (!scaleTransformsAreEqual(smallestSrc, scaleIndexes[smallestIdx], currentSrc, scaleIndexes[currentIdx])) {
-          throw new Error("Incompatible zarr arrays: scale levels of equal size have different scale transformations");
+          // today we are going to treat this as a warning.
+          // For our implementation it is enough that the xyz pixel ranges are the same.
+          // Ideally scale*arraysize=physical size is really the quantity that should be equal, for combining two volume data sets as channels.
+          console.warn("Incompatible zarr arrays: scale levels of equal size have different scale transformations");
```
vianamp, line 174: Perfect for now.

vianamp (approved):

src/loaders/zarr_utils/utils.ts
```diff
@@ -119,7 +119,7 @@ function compareZarrArraySize(
   }
 }
 
-const EPSILON = 0.0000001;
```
frasercl, line 122: Just curious: why shrink this?

frasercl (approved):
👍

src/loaders/zarr_utils/utils.ts
toloudis: It's not a great reason, and I can undo it, but it's because Matheus showed me this data
```
SEG
"scale": [
                                5.003550305,
                                1.0,
                                0.5349045238095238,
                                0.270825,
                                0.270825

RAW
scale": [
                                5.003550305,
                                1.0,
                                0.53,
                                0.2708333333333333,
                                0.2708333333333333
```

frasercl: :shrug: works for me. If our arbitrary fudge factor to reflect real-world conditions doesn't match real-wold conditions, there's no rigorous reason not to change it.

PR #207, opened 2024-04-22
multi-source ome zarr dimension matching:  loosen one more level comparison criterion

toloudis:
When we are combining zarr channels from two different data sources, we have this loop that tries to find matching array sizes ("shape") among the multiresolution levels of the two zarr sources.

When shapes differ by increasing one dimension but decreasing others, we return an undefined sort value.  This is fine but it also means the levels are unmatched.  We don't need to crash on this - we know the combination is not a match and can continue iterating.

### DIFF
--- a/src/loaders/zarr_utils/utils.ts
+++ b/src/loaders/zarr_utils/utils.ts
@@ -163,7 +163,12 @@ export function matchSourceScaleLevels(sources: ZarrSource[]): void {
       if (!ordering) {
         // Arrays are equal, or they are uncomparable
         if (ordering === undefined) {
-          throw new Error("Incompatible zarr arrays: pixel dimensions are mismatched");
+          // We don't need to Error on this at this time.
+          // This just means that the dimensions aren't all changing in the same direction.
+          // We can just skip this as a failed check and continue.
+          console.warn("Incompatible zarr arrays: pixel dimensions are mismatched");
+          allEqual = false;
+          continue;
         }
         // Now we know the arrays are equal, but they may still be invalid to match up because...
         // ...they have different scale transformations
--- a/src/test/zarr_utils.test.ts
+++ b/src/test/zarr_utils.test.ts
@@ -304,9 +304,9 @@ describe("zarr_utils", () => {
       expectSourcesEqual(testSources, refSources);
     });
 
-    it("throws an error if the size of two scale levels are mismatched", async () => {
+    it("Does not throw an error if the size of two scale levels are mismatched", async () => {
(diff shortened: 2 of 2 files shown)


src/loaders/zarr_utils/utils.ts
```diff
@@ -163,7 +163,12 @@ export function matchSourceScaleLevels(sources: ZarrSource[]): void {
       if (!ordering) {
         // Arrays are equal, or they are uncomparable
         if (ordering === undefined) {
-          throw new Error("Incompatible zarr arrays: pixel dimensions are mismatched");
+          // We don't need to Error on this at this time.
+          // This just means that the dimensions aren't all changing in the same direction.
+          // We can just skip this as a failed check and continue.
+          console.warn("Incompatible zarr arrays: pixel dimensions are mismatched");
+          allEqual = false;
+          continue;
```
frasercl, line 171: Okay, I've thought back through why I decided to error here, and I think more might need to happen to make it sound not to error.

Recall that the inner loop of this function is advancing through successively smaller scale levels of each provided multiscale simultaneously. Informally, every loop iteration has only one scale level it's "currently looking at" per multiscale. Then without all the ceremony, it basically looks like this:

```
allEqual = true
smallestMultiscale = multiscales[0]

For each multiscale m:
    If (current scale level of m) != (current scale level of smallestMultiscale):
        allEqual = false
        If (current scale level of m) < (current scale level of smallestMultiscale):
            smallestMultiscale = m

If allEqual:
    save this set of scale levels as matching
    advance the current scale level by one for every multiscale
else:
    advance the current scale level by one for every multiscale except smallestMultiscale
```

That last line is the important one: if one of the comparisons in the loop is undefined, which multiscales should be advanced? Which should be held constant?

The strategy here is to just let `smallestMultiscale` default to whichever multiscale came first. That's fine and the function won't error, but if the function ends up finding a match after encountering a comparison like this, it will to some extent be due to an accident of how the multiscales were ordered. I'm okay with naively doing our best here, but it might also be a good idea to ask if we can refine how the comparison works, or do something smarter when it's undefined.

toloudis:
Putting this back to draft in order to work out the best way to handle it.

src/loaders/zarr_utils/utils.ts
toloudis: Thanks! I figured my approach was probably too naive. I'll take a harder look at this algorithm and try to figure out something smarter.

toloudis:
closing this PR until we have an actual bug to fix

PR #209, opened 2024-04-29
allow differing numbers of time steps

### LINKED ISSUE #208: multi-source: allow number of T to be different
## Use Case

We have data sets where the segmentations are fewer timepoints than the raw.  We would like to be able to show these in the viewer anyway.

## Solution

It would be fine to truncate the time series at the minimum T of all the sources.  We can console-log as a warning that this is happening.

toloudis:
resolves #208 

The use case is that our segmentations have fewer t samples than the raw images.  We would like to just warn in the console and quietly display as if the time sequence had the min of the two times.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -316,9 +316,10 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
       times = shapeLv[t];
       for (let i = 0; i < this.sources.length; i++) {
         const shape = this.sources[i].scaleLevels[levelToLoad].shape;
-        if (shape[t] < times) {
-          console.warn("The number of time points is not consistent across sources: ", shape[t], times);
-          times = shape[t];
+        const tindex = this.sources[i].axesTCZYX[0];
+        if (shape[tindex] < times) {
+          console.warn("The number of time points is not consistent across sources: ", shape[tindex], times);
+          times = shape[tindex];
         }
       }
     }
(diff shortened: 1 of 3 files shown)


interim17 (approved):
From what I know about vv this seems to make sense, but I haven't tested it as I'm not sure what datasets to use? If you want more thorough testing I'm happy to do it, otherwise its a go from me :)

ShrimpCryptid (approved):

src/loaders/OmeZarrLoader.ts
```diff
@@ -309,7 +309,19 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+    // use the min size of T
+    let times = 1;
+    if (hasT) {
+      times = shapeLv[t];
+      for (let i = 0; i < this.sources.length; i++) {
+        const shape = this.sources[i].scaleLevels[levelToLoad].shape;
+        if (shape[t] < times) {
+          console.warn("The number of time points is not consistent across sources: ", shape[t], times);
+          times = shape[t];
+        }
```
frasercl, line 322: `this.sources[i]` may not have the same dimension order as `source0`. Rather than indexing `shape` with `t`, grab the source's time index at `this.sources[i].axesTCZYX[0]`.

src/loaders/zarr_utils/utils.ts
```diff
@@ -177,7 +177,11 @@ export function matchSourceScaleLevels(sources: ZarrSource[]): void {
         const largestT = smallestSrc.axesTCZYX[0] > -1 ? smallestArr.shape[smallestSrc.axesTCZYX[0]] : 1;
         const currentT = currentSrc.axesTCZYX[0] > -1 ? currentArr.shape[currentSrc.axesTCZYX[0]] : 1;
         if (largestT !== currentT) {
-          throw new Error("Incompatible zarr arrays: different numbers of timesteps");
+          console.warn(
+            "Incompatible zarr arrays: different numbers of timesteps: ",
+            smallestArr.shape,
+            currentArr.shape
+          );
```
frasercl, line 184: might add a bit of extra text to relate these numbers

frasercl (approved):

src/loaders/OmeZarrLoader.ts
toloudis: Thank you!

toloudis: This type of thing makes me want to have an internal shape representation that's always TCZYX

PR #212, opened 2024-05-08
feat: Added error for zarr files that exceed max atlas size

### LINKED ISSUE #174: handle zarrs that are too big
## Use Case

Accounting for all multiresolution levels, if:
* 2D mode: zarr has data in which a single slice is larger than our texture size
* 3D mode: zarr has data in which the XYZ dimensions do not tile into our preferred texture size

then we either have to error out in some way or downsample in the viewer

## Solution

Error out gracefully - perhaps find a way to propagate error out to clients, even if it's just `throw new Error("Zarr too big")`

## Alternatives
1. show some cropped subset of the data which is easy to display quickly.
2. Downsample on the fly.  Has performance concerns naively downsampling potentially very large data.  Potentially fraught with other details to deal with.

ShrimpCryptid:
Closes #174, handling zarr files that are too large.

*Estimated review size: small, 10 minutes*

## Changes
- Adds a check in VolumeLoaderUtils for Zarr files that exceed our current max atlas dimension (4096x4096), and throws an error if it cannot be loaded.

TODO noted by Dan: For certain datasets, it may be possible to open the Z-slice data but not the 3D data. Should we do parallel checks for both, and switch to 2D mode if it's the only view mode we can support?
TODO: Enforce maximum on the number of channels that can be enabled? 

## Validation
1. Run volume-viewer locally.
2. Switch to the new "Fly Brain" Zarr in the dropdown.
3. Press Ctrl + Alt + 1 to open the developer menu. 
4. Change the maxAtlasEdge to 512. There should be an error message for the 3D view.
5. Switch to Z-slice mode. The error message should disappear and a slice should appear.

### DIFF
--- a/src/loaders/VolumeLoaderUtils.ts
+++ b/src/loaders/VolumeLoaderUtils.ts
@@ -2,6 +2,7 @@ import { Box3, Vector2, Vector3 } from "three";
 
 import { ImageInfo } from "../Volume.js";
 import { LoadSpec } from "./IVolumeLoader.js";
+import { VolumeLoadError, VolumeLoadErrorType } from "./VolumeLoadError.js";
 
 export const MAX_ATLAS_EDGE = 4096;
 
@@ -77,8 +78,24 @@ export function computePackedAtlasDims(z: number, tw: number, th: number): Vecto
   return new Vector2(nrows, ncols);
 }
 
+function doesSpatialDimensionFitInAtlas(
+  spatialDimZYX: [number, number, number],
+  maxAtlasEdge = MAX_ATLAS_EDGE
+): boolean {
+  // Estimate atlas size
+  const x = spatialDimZYX[2];
+  const y = spatialDimZYX[1];
+  const z = spatialDimZYX[0];
+  const xtiles = Math.floor(maxAtlasEdge / x);
+  const ytiles = Math.floor(maxAtlasEdge / y);
+  return xtiles * ytiles >= z;
+}
+
 /** Picks the largest scale level that can fit into a texture atlas with edges no longer than `maxAtlasEdge`. */
-export function estimateLevelForAtlas(spatialDimsZYX: [number, number, number][], maxAtlasEdge = MAX_ATLAS_EDGE) {
+export function estimateLevelForAtlas(
+  spatialDimsZYX: [number, number, number][],
+  maxAtlasEdge = MAX_ATLAS_EDGE
+): number | undefined {
   if (spatialDimsZYX.length <= 1) {
     return 0;
   }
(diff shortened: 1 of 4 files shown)


ShrimpCryptid:
Marking this ready for review but @frasercl @toloudis let me know if there's additional checks or debug features I should include. (We had talked about a few but I wasn't sure if they were requirements or just nice-to-haves.)

src/loaders/VolumeLoaderUtils.ts
```diff
@@ -130,8 +141,33 @@ export function scaleMultipleDimsToSubregion(subregion: Box3, dims: ZYX[]): ZYX[
...
+    levelToLoad = spatialDimsZYX.length - 1;
+  }
+  const smallestDims = spatialDimsZYX[levelToLoad];
+  console.error(
+    `Volume is too large; no multiscale level found that fits in preferred memory footprint. Selected level ${levelToLoad}  has dimensions `,
+    smallestDims,
+    `. Max atlas edge allowed is ${loadSpec.maxAtlasEdge}.`
+  );
+  console.log("All available levels: ", spatialDimsZYX);
+  throw new VolumeLoadError(`Volume is too large; multiscale level does not fit in preferred memory footprint.`, {
```
toloudis, line 168: If they are getting this error for 3D/YZ/ZX, we may want to suggest to users that they switch to XY mode in which their data might be viewable.  I think we need to work out how to handle that in the best way.  Maybe we need a future ticket for this.  We also need to consider how to handle this gracefully on the front end (w3cv) so maybe the 3d view stays empty but they can still switch to XY and use the viewer.

src/loaders/VolumeLoaderUtils.ts
```diff
@@ -130,8 +141,33 @@ export function scaleMultipleDimsToSubregion(subregion: Box3, dims: ZYX[]): ZYX[
  */
 export function pickLevelToLoadUnscaled(loadSpec: LoadSpec, spatialDimsZYX: ZYX[]): number {
   const optimalLevel = estimateLevelForAtlas(spatialDimsZYX, loadSpec.maxAtlasEdge);
-  const levelToLoad = Math.max(optimalLevel + (loadSpec.scaleLevelBias ?? 0), loadSpec.multiscaleLevel ?? 0);
-  return Math.max(0, Math.min(spatialDimsZYX.length - 1, levelToLoad));
+  let levelToLoad = optimalLevel;
```
frasercl, line 144: Looks like there's not really a need for `optimalLevel` now that `levelToLoad` is immediately set to its value

frasercl (approved):

src/loaders/VolumeLoaderUtils.ts
ShrimpCryptid: Oops, holdover from when I was also reporting optimalLevel!

Issue #223, opened 2024-07-10
last z slice can fail to display


toloudis:
## Description

When loading some data sets, and going into XY mode and moving slider to last slice, sometimes the last slice shows up completely black.    ( A preliminary debugging shows that the single-slice shader thinks the Z value is out of bounds (greater than 1) in this scenario )

## Expected Behavior

Expect to be able to see every Z slice when in XY mode.

## Reproduction

https://allen-cell-animated.github.io/website-3d-cell-viewer/viewer?url=https%3A%2F%2Fdev-aics-dtp-001.int.allencell.org%2F%2Fassay-dev%2Fcomputational%2Fdata%2Frelease%2Femt-data%2Fmain%2F3500006256_20X_P12_B10%2Fraw.ome.zarr%2F
Go to XY mode.
move z slider to max value.
Data will never show up.

toloudis:
See this line in the shader:
pos.xyz = (pos.xyz - SUBSET_OFFSET) / SUBSET_SCALE;

We might have some floating point error that pushes this out of bounds when you have a 1x1 atlas
Removing that line makes this data show up again.  Haven’t tested on any other images.
For this data, SUBSET_OFFSET.z = 0.9667 and SUBSET_SCALE.z = 0.0333 (approx 29/30 and 1/30)

Confirmed that was pushing pos.z > 1.0
Changing the bounds check in sampleAtlas to have an epsilon also fixes it:
    pos[2] >= 0.0 && pos[2] <= 1.0001); // used to be 1.0

toloudis:
(1.0-0.9666)/(0.0333) > 1.0

frasercl:
Closed by #224

PR #224, opened 2024-07-10
remove subset adjustment from slice shader

toloudis:
Fixing #223 
This extra adjustment was incurring floating point errors which make the last z slice go out of bounds.
Rather than add a bounds epsilon, I found that removing this code seemed to fix the bug without causing any visual discrepancies.

@frasercl would love your insight as to whether this is totally safe.  I think I tested a variety of files but not sure I know why it was there in the first place.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -603,8 +603,8 @@ function updateZSliceUI(volume: Volume) {
   const zInput = document.getElementById("zValue") as HTMLInputElement;
 
   const totalZSlices = volume.imageInfo.volumeSize.z;
-  zSlider.max = `${totalZSlices}`;
-  zInput.max = `${totalZSlices}`;
+  zSlider.max = `${totalZSlices - 1}`;
+  zInput.max = `${totalZSlices - 1}`;
 
   zInput.value = `${Math.floor(totalZSlices / 2)}`;
   zSlider.value = `${Math.floor(totalZSlices / 2)}`;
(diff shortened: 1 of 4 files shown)


public/index.ts
```diff
@@ -603,8 +603,8 @@ function updateZSliceUI(volume: Volume) {
   const zInput = document.getElementById("zValue") as HTMLInputElement;
 
   const totalZSlices = volume.imageInfo.volumeSize.z;
-  zSlider.max = `${totalZSlices}`;
-  zInput.max = `${totalZSlices}`;
+  zSlider.max = `${totalZSlices - 1}`;
+  zInput.max = `${totalZSlices - 1}`;
```
toloudis, line 607: this is just a bonus fix to fix the slider range in the testbed viewer

src/constants/shaders/slice.frag
```diff
@@ -87,7 +86,6 @@ void main() {
   vec4 pos = vec4(vUv, 
     (SLICES==1.0 && Z_SLICE==0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0), 
     0.0);
-  pos.xyz = (pos.xyz - SUBSET_OFFSET) / SUBSET_SCALE;
```
toloudis, line 90: removing this line is the bug fix.  For example: when pos.z = 1.0, we could get into a situation where this math was doing (1.0 - 0.9666) /  0.0333 which is slightly greater than 1.  Then later there is bounds checking code which needs to ensure that pos is never greater than 1.

rugeli (approved):

frasercl (approved):

Issue #225, opened 2024-07-10
playback needs to be faster


toloudis:
## Use Case

Many scientists just want to play through their time series at a reasonable frame rate and are willing to sacrifice resolution to do it.

## Solution

Lower res for playback especially in Z.

## Alternatives

drop down in res only while playing and then load higher res.

Smarter buffering idea: This is just one possible implementation but not nearly the only one:
Pre-request 5 frames ahead at very low res, next 5 frames at one res higher, and next 5 even higher.    Then if the playback catches up, cancel high res requests and re-request lower.

toloudis:
look into ways to properly cancel using AbortSignals
https://zarrita.dev/packages/storage.html#specific-request-options

PR #226, opened 2024-07-12
tuneable settings for data loading optimization

toloudis:
Add instrumentation for testing various settings, via the advanced settings tweakpane.  Should not change any functionality but allows testing and evaluation of different parameters by changing things at runtime.

Goal: high framerate playthrough of single slices along T or Z (T is more important). (see #225)

Options added:
- during playthrough, ONLY prefetch along "priority direction".  This means don't prefetch along Z while playing through T.
- select number of chunks to prefetch
- Allow throttling the arriving channeldata so that you don't accidentally get many frames at once and "hitches" in framerate.  

All options need evaluation and testing.
Interesting things to test:  
* explicit selection of multiresolution level
* turn prefetching down to 0
* try different powers of 2 for the max texture atlas size

Note: in the main app, website-3d-cell-viewer, we still apply a playback bias which is incompatible with these settings.  Some code change is still required for the main app to respect these tweakpane settings.

Also of note: In general, multichannel playback is a lot slower since it has to wait for more data.   Maybe we can reduce resolution even more when extra channels are enabled?


public/index.ts
```diff
@@ -889,9 +889,9 @@ function onChannelDataArrived(v: Volume, channelIndex: number) {
   view3D.updateActiveChannels(currentVol);
   view3D.updateLuts(currentVol);
 
-  if (currentVol.isLoaded()) {
-    console.log("currentVol with name " + currentVol.name + " is loaded");
-  }
+  // if (currentVol.isLoaded()) {
+  //   console.log("currentVol with name " + currentVol.name + " is loaded");
+  // }
```
meganrm, line 894: do you mean to have this commented out? delete?

src/Volume.ts
```diff
@@ -260,6 +272,7 @@ export default class Volume {
   /** Returns `true` iff differences between `loadSpec` and `loadSpecRequired` indicate new data *must* be loaded. */
   private mustLoadNewData(): boolean {
     return (
+      this.loadSpec.useExplicitLevel !== this.loadSpecRequired.useExplicitLevel || // explicit vs automatic level changed
       this.loadSpec.time !== this.loadSpecRequired.time || // time point changed
```
meganrm, line 276: nit but these could be constants so the final line reads something like `return explicitVsAutoLevelChanged || timePointChanged` so you don't need the comments

meganrm (approved):

src/View3d.ts
```diff
@@ -35,10 +36,21 @@ export interface View3dOptions {
...
+  throttleArrivingChannelData: true,
+};
+
 /**
  * @class
  */
 export class View3d {
+  // TODO because View3d is basically a top level entrypoint for volume-viewer,
+  // maybe it should create the VolumeLoaderContext with options passed in.
+  // (instead of having the loaderContext created externally)
```
frasercl, line 51: Q to ask first: will we ever want the option to prevent View3d from creating a webworker and force all work to happen on the main thread?

(A: probably not)

toloudis: I think it's fine if this viewer only ever loads data via a worker.  

However, I also think it's potentially extremely useful if you can _easily_ write code to load volume data that isn't necessarily tied to rendering.  Such as for scripting some intensity calculation like you did several weeks back.  I don't care whether code like that goes through a worker behind the scenes, as long as we have a nice simple and clean interface for passing in a url and some loadspec parameters and getting data back.

frasercl: That sounds like a great argument for giving View3d control of its own load context _and_ a solid endorsement of how loading is set up otherwise. Making the actual loaders used behind the scenes available to client packages is good, but requiring clients to engage with them when they just want to load an image isn't necessary when it's easy enough to abstract away.

src/View3d.ts
```diff
@@ -35,10 +36,21 @@ export interface View3dOptions {
   useWebGL2?: boolean;
 }
 
+const allGlobalLoadingOptions = {
+  numChunksToPrefetchAhead: 5,
```
rugeli, line 40: Is this value always 5? Would making it dynamically determined provide more flexibility for different use cases?

toloudis: Yes and we will probably be adding some better user-controls for this in the future.

src/loaders/OmeZarrLoader.ts
```diff
@@ -385,6 +392,18 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+      const dims = new VolumeDims();
+
+      dims.spaceUnit = spatialUnit;
+      dims.timeUnit = timeUnit;
+      dims.shape = this.orderByTCZYX(level.shape, 1);
+      dims.spacing = this.getScale(i);
+      dims.dataType = level.dtype;
+
+      alldims.push(dims);
+    });
```
frasercl, line 406: Might be missing something but if this is a `map`, wouldn't this work?

```suggestion
    const alldims = source0.scaleLevels.map((level, i) => {
      const dims = new VolumeDims();

      dims.spaceUnit = spatialUnit;
      dims.timeUnit = timeUnit;
      dims.shape = this.orderByTCZYX(level.shape, 1);
      dims.spacing = this.getScale(i);
      dims.dataType = level.dtype;

      return dims;
    });
```

src/loaders/TiffLoader.ts
```diff
@@ -184,6 +184,15 @@ class TiffLoader extends ThreadableVolumeLoader {
...
       multiscaleLevel: 0,
+      multiscaleLevelDims: [
+        {
+          shape: [dims.sizet, dims.sizec, dims.sizez, dims.sizey, dims.sizex],
+          spacing: [1, 1, dims.pixelsizez, dims.pixelsizey, dims.pixelsizex],
+          spaceUnit: dims.unit || "",
+          timeUnit: "",
+          dataType: dims.pixeltype || "uint8",
+        },
+      ],
```
frasercl, line 195: Kind of odd to see this structure being explicitly made inline in all these loaders, when `IVolumeLoader` requires a different method explicitly for getting them. I get why you did it this way - calling a loader's `loadDims` could re-fire whatever `async` operation grabbed that metadata in the first place - but if `createImageInfo` is going to effectively produce a superset of what `loadDims` gives you, it makes me wonder if we still need `loadDims` in the first place.

More generally, I think there's some growing unresolved confusion around what our different image metadata types (`VolumeDims` vs `ImageInfo` vs `LoadSpec`) are for, though I'm fine leaving it unresolved for this PR.

src/workers/VolumeLoaderContext.ts
```diff
@@ -193,6 +207,14 @@ class VolumeLoaderContext {
...
     return this.activeLoader;
   }
+
+  setThrottleChannelData(throttle: boolean): void {
+    this.workerHandle.setThrottleChannelData(throttle);
+  }
+
+  getActiveLoader(): WorkerLoader | undefined {
+    return this.activeLoader;
+  }
```
frasercl, line 217: This class is designed around the possibility of supporting more than one "active" loader at a time in the future. This method wouldn't make sense in that world, but removing it and replacing whatever depends on it with something that can live with multiple loaders should be about as hard as anything else we'd need to do to support that anyways.

src/workers/VolumeLoaderContext.ts
```diff
@@ -94,7 +98,13 @@ class SharedLoadWorkerHandle {
   private receiveMessage<T extends WorkerMsgType>({ data }: MessageEvent<WorkerResponse<T>>): void {
     if (data.responseResult === WorkerResponseResult.EVENT) {
       if (data.eventType === WorkerEventType.CHANNEL_LOAD) {
-        this.onChannelData?.(data);
+        if (this.onChannelData) {
+          if (this.throttleChannelData) {
+            throttle(() => (this.onChannelData ? this.onChannelData(data) : {}));
```
frasercl, line 103: why throttle here rather than e.g. when requests are sent, or before data is transferred from the worker?

src/View3d.ts
```diff
@@ -35,10 +36,21 @@ export interface View3dOptions {
...
+};
+
 /**
  * @class
  */
 export class View3d {
+  // TODO because View3d is basically a top level entrypoint for volume-viewer,
+  // maybe it should create the VolumeLoaderContext with options passed in.
+  // (instead of having the loaderContext created externally)
+  public loaderContext?: VolumeLoaderContext;
```
frasercl, line 52: Makes me think that `View3d` maybe ought to have a `createLoader` method of its own that calls our existing `createLoader` function within this context. Gets closer to having `View3d` represent the "world" in which all these things exist, which seems to be a relatively common paradigm.

frasercl (approved):

src/loaders/OmeZarrLoader.ts
toloudis: haha yes I have no idea why I didn't do it like that.  will update

src/loaders/TiffLoader.ts
toloudis: I agree it's a good candidate for a further change after this

src/workers/VolumeLoaderContext.ts
toloudis: View3d.setupGui sets up a tweakpane handler that calls `      this.loaderContext?.getActiveLoader()?.updateFetchOptions({...
`

toloudis: (just needed some way to push these values through from the tweakpane layer)

src/workers/VolumeLoaderContext.ts
toloudis: my thought here was that I mainly want to allow for a redraw in between onChannelDatas.  The throttle code lets them get automatically queued up with a time delay of my choosing.

src/View3d.ts
toloudis: I do like this idea.  Someday there might also be a View3d.createScene and getScene ... and that leads to thoughts like better separation of the "viewer" from the actual viewing panel/canvas

src/Volume.ts
```diff
@@ -60,6 +60,9 @@ export type ImageInfo = Readonly<{
 
   /** Number of scale levels available for this volume */
   numMultiscaleLevels: number;
+  /** Dimensions of each scale level, at original size, from the first data source */
+  // TODO THIS DATA IS SOMEWHAT REDUNDANT WITH SOME OF THE OTHER FIELDS IN HERE
+  multiscaleLevelDims: VolumeDims[];
   /** The scale level from which this image was loaded, between `0` and `numMultiscaleLevels-1` */
   multiscaleLevel: number;
```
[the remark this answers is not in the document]
toloudis: I think this is important enough that I'll do it in a quick followup PR but not confuse this one any more.

PR #227, opened 2024-07-12
Two lut enhancements

frasercl:
Review time: small-medium (~10-20min)

Implements one and a half small wishlist items related to lut/control point remapping I had while working on allen-cell-animated/website-3d-cell-viewer#294. This causes no breaking changes to the API, so this code should be fine to update into website-3d-cell-viewer without any further changes.

## Change 1: nudge start and end control points when remapping

When control points are remapped, the first and last control points are almost always pushed either outside the 0-255 range, or into the middle of the range so that the control points no longer technically define a function with a domain of 0-255. The lut generator doesn't at all mind generating a lut with points like this, but they make the transfer function editor look sloppy:

| Before remap | After remap (👎) |
| ----- | ----- |
| image | image |

I added some code to the remapping function that gets a bit smarter about handling these edge points, while still prioritizing preserving information over keeping points within range:

| Before remap | After remap (👍) |
| ----- | ----- |
| image | image |

## Change 1.5: don't generate a lut until necessary

I reverted most of this change in favor of just exporting the `remapControlPoints` function that I need, but some of its fingerprints are still on this PR.

Originally, I made changes to `Lut` that allowed its `lut` field to be `null` on construction so that it only needed to fill it in when necessary. This was to allow `Lut` to be used as a simple control point container without incurring the overhead of filling in a `Uint8Array`. This was motivated by the fact that I was having to create and use a `Lut` to manually remap control points in allen-cell-animated/website-3d-cell-viewer#294, but again the simpler solution here is to just export the underlying function.

Again, most of this change is gone now, but the commits are still there and waiting to be re-added if it sounds useful for other reasons.

### DIFF
--- a/src/Lut.ts
+++ b/src/Lut.ts
@@ -475,7 +475,8 @@ export function remapControlPoints(
   oldMin: number,
   oldMax: number,
   newMin: number,
-  newMax: number
+  newMax: number,
+  noNudge = false
 ): ControlPoint[] {
   if (controlPoints.length === 0) {
     return controlPoints;
@@ -502,6 +503,10 @@ export function remapControlPoints(
       color: [cp.color[0], cp.color[1], cp.color[2]],
     };
     newControlPoints.push(newCP);
+  }
+
+  if (noNudge) {
+    return newControlPoints;
   }
 
   // Commonly (e.g. in the output of most of the LUT generators above), the first and last control points define a line
--- a/src/Lut.ts
+++ b/src/Lut.ts
@@ -182,10 +182,10 @@ export class Lut {
         const [first, last] = this.controlPoints;
         if (first.x === 0 && last.x === 255 && first.opacity === 0 && last.opacity === 1) {
           this.#lut = createFullRangeLut();
+          return this.#lut;
         }
-      } else {
-        this.#lut = controlPointsToLut(this.controlPoints);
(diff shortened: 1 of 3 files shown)


src/Lut.ts
```diff
@@ -479,5 +484,48 @@ export function remapControlPoints(
...
+  const EPSILON = 0.0001;
+  const first = controlPoints[0];
+  const second = controlPoints[1];
+  const secondLast = controlPoints[controlPoints.length - 2];
+  const last = controlPoints[controlPoints.length - 1];
+
+  if (Math.abs(first.opacity - (second?.opacity ?? Infinity)) < EPSILON) {
+    if (first.x < 0) {
+      // control point is now out of bounds - clamp it to 0 (or as close as we can get without losing information)
+      first.x = Math.min(0, second.x - 1);
```
toloudis, line 513: good catch for if both first.x and second.x are less than zero

toloudis (approved):
I think this looks good!  Eager to see it working.  

I wish the unit tests were nicer overall - I kinda feel like I made terrible ones but it's hard to make it totally obvious what's even happening.  Once we are using true intensities in the domain of the control points I feel things will be a bit simpler.

ascibisz (approved):

PR #231, opened 2024-07-23
feature: Camera transform getters + setters

ShrimpCryptid:
Adds methods for getting and setting the camera transform on the View3D; this helps with the implementation of https://github.com/allen-cell-animated/website-3d-cell-viewer/issues/283, saving/loading camera position from the URL.

*Estimated review size: small, 5-10 minutes*

## Changes
- Added a new exported type, `CameraTransform`, which contains the position, rotation, up vector, and target position of the camera + camera controller.
- Adds getters and setters for the camera transform from the `ThreeJsPanel`, which are exposed via the `View3d`.
  - Note that all data is passed via arrays, so clients of `View3d` don't have to import ThreeJs directly.

### DIFF
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -28,12 +28,12 @@ const DEFAULT_PERSPECTIVE_CAMERA_FAR = 20.0;
 
 const DEFAULT_ORTHO_SCALE = 0.5;
 
-export interface CameraTransform {
+export type CameraTransform = {
   position: [number, number, number];
   rotation: [number, number, number];
   up: [number, number, number];
   target: [number, number, number];
-}
+};
 
 export class ThreeJsPanel {
   public containerdiv: HTMLDivElement;
@@ -624,7 +624,7 @@ export class ThreeJsPanel {
 
   setCameraTransform(transform: Partial<CameraTransform>) {
     const currentTransform = this.getCameraTransform();
-    // Fill in any missing properties with current
+    // Fill in any missing properties with current transform
     const newTransform = { ...currentTransform, ...transform };
 
     this.camera.up = new Vector3().fromArray(newTransform.up).normalize();
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -33,6 +33,7 @@ export type CameraTransform = {
   rotation: [number, number, number];
   up: [number, number, number];
   target: [number, number, number];
+  orthoScales: [number, number, number];
 };
 
 export class ThreeJsPanel {
@@ -619,6 +620,7 @@ export class ThreeJsPanel {
       rotation: this.camera.rotation.toArray() as [number, number, number],
       up: this.camera.up.toArray(),
       target: this.controls.target.toArray(),
+      orthoScales: [this.orthoControlsX.scale, this.orthoControlsY.scale, this.orthoControlsZ.scale],
     };
   }
 
@@ -631,6 +633,14 @@ export class ThreeJsPanel {
     this.camera.position.set(...newTransform.position);
     this.camera.setRotationFromEuler(new Euler().fromArray(newTransform.rotation));
     this.controls.target = new Vector3().fromArray(newTransform.target);
+    // Update orthographic cameras
+    const orthoControls = [this.orthoControlsX, this.orthoControlsY, this.orthoControlsZ];
+    const orthoCameras = [this.orthographicCameraX, this.orthographicCameraY, this.orthographicCameraZ];
+    for (let i = 0; i < orthoControls.length; i++) {
+      const scale = newTransform.orthoScales[i];
+      orthoControls[i].scale = scale;
+      orthoCameras[i].zoom = 0.5 / scale;
+    }
   }
 
   render(): void {
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -21,6 +21,7 @@ import Timing from "./Timing.js";
 import scaleBarSVG from "./constants/scaleBarSVG.js";
 import { isOrthographicCamera, ViewportCorner, isTop, isRight } from "./types.js";
 import { formatNumber } from "./utils/num_utils.js";
+import { CameraTransform } from "../public/types.js";
 
 const DEFAULT_PERSPECTIVE_CAMERA_DISTANCE = 5.0;
 const DEFAULT_PERSPECTIVE_CAMERA_NEAR = 0.001;
@@ -606,15 +607,16 @@ export class ThreeJsPanel {
     return this.renderer.getContext().canvas.height;
   }
 
-  getCameraTransform(): { position: Vector3; rotation: Euler; up: Vector3 } {
+  getCameraTransform(): CameraTransform {
     return {
       position: this.camera.position.clone(),
       rotation: this.camera.rotation.clone(),
       up: this.camera.up.clone(),
+      target: this.controls.target.clone(),
     };
   }
 
-  setCameraTransform(transform: Partial<{ position: Vector3; rotation: Euler; up: Vector3 }>) {
+  setCameraTransform(transform: Partial<CameraTransform>) {
     const currentTransform = this.getCameraTransform();
     // Fill in any missing properties with current
     const newTransform = { ...currentTransform, ...transform };
@@ -622,6 +624,7 @@ export class ThreeJsPanel {
     const newPos = newTransform.position;
     this.camera.position.set(newPos.x, newPos.y, newPos.z);
     this.camera.setRotationFromEuler(newTransform.rotation);
+    this.controls.target = newTransform.target;
   }
 
   render(): void {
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -60,4 +60,5 @@ export {
   RENDERMODE_PATHTRACE,
   RENDERMODE_RAYMARCH,
   SKY_LIGHT,
+  CameraTransform,
```
ShrimpCryptid, line 63: @frasercl will this cause any issues if this type export doesn't have the `type` annotation in front of it? I get a babel warning if I include it, regardless of whether `CameraTransform` is an interface or type.

ShrimpCryptid:
Converting to draft because there are still some parameters missing

src/index.ts
```diff
@@ -60,4 +60,5 @@ export {
   RENDERMODE_PATHTRACE,
   RENDERMODE_RAYMARCH,
   SKY_LIGHT,
+  type CameraTransform,
```
ShrimpCryptid, line 63: This throws a babel warning, any idea why it might be triggering off of this and not the similar declarations on lines 47-49?

frasercl: I'm not seeing a similar error, and don't have any great theories on why you would be

src/ThreeJsPanel.ts
```diff
@@ -605,6 +614,36 @@ export class ThreeJsPanel {
...
+
+  setCameraTransform(transform: Partial<CameraTransform>) {
+    const currentTransform = this.getCameraTransform();
+    // Fill in any missing properties with current transform
+    const newTransform = { ...currentTransform, ...transform };
+
+    this.camera.up = new Vector3().fromArray(newTransform.up).normalize();
+    this.camera.position.set(...newTransform.position);
+    this.camera.setRotationFromEuler(new Euler().fromArray(newTransform.rotation));
+    this.controls.target = new Vector3().fromArray(newTransform.target);
```
frasercl, line 635: 1. Any reason `position` is set in-place but `up` and `target` get new vector objects?
2. I believe you can accomplish the same thing you're doing to `position` without the spread (`...`) by using `fromArray`
3. If no on 1, you could set them all this way

src/ThreeJsPanel.ts
```diff
@@ -27,6 +28,14 @@ const DEFAULT_PERSPECTIVE_CAMERA_FAR = 20.0;
 
 const DEFAULT_ORTHO_SCALE = 0.5;
 
+export type CameraTransform = {
```
frasercl, line 31: "transform" seems like an odd name for this type - it makes me think of a "transformation matrix" rather than the decomposed settings captured by this object - but I'm having a hard time coming up with a better one... "Position" isn't general enough to capture everything this type contains, "settings" is too general... 🤷

frasercl (approved):
Looks great overall, excited to see you got this working!

src/ThreeJsPanel.ts
ShrimpCryptid: `camera.position` is a readonly property, so I can't assign it to a new reference. 

Made some changes, see if this looks better!

frasercl: I like it!

src/ThreeJsPanel.ts
toloudis: I agree that camera transform isn't quite right.  There are a few weird things here.
1. Just be aware: having rotation and target exist independently means they can become inconsistent, which is bad, but it also is true that "target" is overspecifying rotation.  The location of target is for the camera CONTROLS, not just the camera view.    Basically if you provide position and target, then rotation is fully determined. But ultimately, I guess it's ok to provide rotation and target separately.. I might mull this over a bit longer.
2. For perspective cameras, you need to capture the field of view angle too.  Put a comment or something to say whether it's half the fov or the full fov, and whether it describes horizontal or vertical fov. (Or put it in the name like fullVerticalFovDegrees - overkill....)
3. Is orthoScales supposed to be representing the 3 separate ortho cameras?  Or will you record 4 different CameraTransform objects, one for each of the 4 viewing cameras?  This feels weirdly overloaded.

I'd call this data structure, with the fov angle added, something more like CameraData, CameraInfo, or Camera.

src/index.ts
toloudis: > This throws a babel warning, any idea why it might be triggering off of this and not the similar declarations on lines 47-49?

maybe because you didn't use `import type` to import it?

src/ThreeJsPanel.ts
toloudis: Part of my comment above is wondering whether you mean to capture only the active viewing camera, or cameras for all other view modes too.

toloudis (commented):
I think the main data structure can be refined/futureproofed a little

src/ThreeJsPanel.ts
ShrimpCryptid: @3: `orthoScales` captures the current camera scales but the rest of the CameraTransform only captures details for the currently active camera. Do you think it's important that we sync all four cameras when we share links?

toloudis: Well I think we should or we shouldn't.  No point in capturing orthoScales unless the current camera is ortho, and then you only need one of them right?  Most important is to capture the "current" view. So I don't think it's super critical to save all 4 cameras.   I suggest making the data structure for all the data needed to record one single camera, and then later on if we want to save them for multiple views ,we save whole entire cameras.

src/index.ts
ShrimpCryptid: That was it, thank you!

src/ThreeJsPanel.ts
ShrimpCryptid: Also great point on target + rotation being redundant! I originally only stored position + rotation and it led to some bugged camera behavior until I also added in the target value-- I didn't go back to check if I could remove the rotation after.

src/ThreeJsPanel.ts
```diff
@@ -13,6 +13,7 @@ import {
   NormalBlending,
   WebGLRenderer,
   Scene,
+  Euler,
```
toloudis, line 16: can remove, says the linter

src/ThreeJsPanel.ts
```diff
@@ -605,6 +616,35 @@ export class ThreeJsPanel {
...
+  }
+
+  setCameraState(state: Partial<CameraState>) {
+    const currentState = this.getCameraState();
+    // Fill in any missing properties with current state
+    const newState = { ...currentState, ...state };
+
+    this.camera.up.fromArray(newState.up).normalize();
+    this.camera.position.fromArray(newState.position);
+    this.controls.target.fromArray(newState.target);
```
toloudis, line 636: do you need to do something like controls.update() to make sure the camera gets the correct rotation? Or maybe should set camera rotation here with a little math.
Also one thing I would be slightly worried about is if the camera is being overridden in an incorrect way (like XY mode being given a perspective camera?).  Is there a way to error-check that?  (Maybe make sure ortho modes are getting ortho cameras, and make sure the direction of the camera is consistent with the ortho mode.  3d mode is more free in terms of direction)

src/ThreeJsPanel.ts
```diff
@@ -605,6 +616,35 @@ export class ThreeJsPanel {
...
+
+    // Update fields by camera type
+    if (this.camera.type === "OrthographicCamera") {
```
toloudis, line 639: prefer this.camera.isOrthographicCamera:  https://threejs.org/docs/#api/en/cameras/OrthographicCamera.isOrthographicCamera

toloudis (commented):
I have one or two cleanup things and one bigger error checking thing to consider

src/ThreeJsPanel.ts
ShrimpCryptid: `isOrthographicCamera()` is only implemented on `OrthographicCamera` and not on `PerspectiveCamera`... so it's not actually a valid call unless I cast `this.camera`, and then the TypeScript conditional logic breaks down.

toloudis: We've already tackled that... you can call this func:  https://github.com/allen-cell-animated/volume-viewer/blob/98ff141423c4a94a8d8e8073e10fd0d5f2d4d708/src/types.ts#L89

ShrimpCryptid: Thanks!

src/ThreeJsPanel.ts
```diff
@@ -605,6 +622,48 @@ export class ThreeJsPanel {
     return this.renderer.getContext().canvas.height;
   }
 
+  getCameraState(): CameraState {
+    return {
+      position: this.camera.position.toArray(),
+      up: this.camera.up.toArray(),
+      target: this.controls.target.toArray(),
+      orthoScale: this.camera.type === "OrthographicCamera" ? this.controls.scale : undefined,
```
toloudis, line 630:
```suggestion
      orthoScale: isOrthographicCamera(this.camera) ? this.controls.scale : undefined,
```

toloudis (approved):
Looks great - could do one small cleanup of the ortho/perspective camera checking but otherwise good to go.

PR #233, opened 2024-08-06
feature: Format timestamp

ShrimpCryptid:
Will close https://github.com/allen-cell-animated/website-3d-cell-viewer/issues/265 ("time in minutes should automatically change to hours/days") once changes are released and merged.

*Estimated review size: medium, 20 minutes*

## Changes
- Changes the timestamp formatter to recognize days, hours, minutes, seconds, or milliseconds, and to format these units into a `d:hh:mm:ss.sss`-style timestamp.
  - All units smaller than the provided unit are dropped.
  - The total timestamp is used to determine the largest time unit to show.

Changes, as seen in 3DVV: 
image

### DIFF
--- a/src/utils/num_utils.ts
+++ b/src/utils/num_utils.ts
@@ -1,5 +1,3 @@
-import { use } from "chai";
-
 import { parseTimeUnit, TimeUnit } from "../constants/time.js";
 
 export const DEFAULT_SIG_FIGS = 5;
--- a/src/utils/num_utils.ts
+++ b/src/utils/num_utils.ts
@@ -89,15 +89,15 @@ export function formatNumber(value: number, sigFigs = DEFAULT_SIG_FIGS, sciSigFi
 }
 
 export function timeToMilliseconds(time: number, unit: TimeUnit): number {
-  if (unit == TimeUnit.Millisecond) {
+  if (unit == TimeUnit.MILLISECOND) {
     return time;
-  } else if (unit == TimeUnit.Second) {
+  } else if (unit == TimeUnit.SECOND) {
     return time * SECONDS_IN_MS;
-  } else if (unit == TimeUnit.Minute) {
+  } else if (unit == TimeUnit.MINUTE) {
     return time * MINUTES_IN_MS;
-  } else if (unit == TimeUnit.Hour) {
+  } else if (unit == TimeUnit.HOUR) {
     return time * HOURS_IN_MS;
-  } else if (unit == TimeUnit.Day) {
+  } else if (unit == TimeUnit.DAY) {
     return time * DAYS_IN_MS;
   } else {
     throw new Error("Unrecognized time unit");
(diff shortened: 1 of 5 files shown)


src/utils/num_utils.ts
```diff
@@ -1,5 +1,12 @@
+import { parseTimeUnit, TimeUnit } from "../constants/time.js";
+
 export const DEFAULT_SIG_FIGS = 5;
 
+const SECONDS_IN_MS = 1000;
+const MINUTES_IN_MS = 1000 * 60;
+const HOURS_IN_MS = 1000 * 60 * 60;
+const DAYS_IN_MS = 1000 * 60 * 60 * 24;
```
toloudis, line 8: Minor Nit:
```suggestion
const DAYS_IN_MS = HOURS_IN_MS * 24;
```
and same for other lines above... slightly less error prone

toloudis (approved):
✅ 
So many LoC!  Is this mostly borrowed from TFE?

ShrimpCryptid:
> ✅ So many LoC! Is this mostly borrowed from TFE?

Some of the logic is borrowed, but it's a little more involved to handle many different units! TFE only accepts time units in seconds, but VV can take any number of time units.

src/utils/num_utils.ts
```diff
@@ -78,3 +85,118 @@ export function formatNumber(value: number, sigFigs = DEFAULT_SIG_FIGS, sciSigFi
...
+  } else if (unit == TimeUnit.MINUTE) {
+    return time * MINUTES_IN_MS;
+  } else if (unit == TimeUnit.HOUR) {
+    return time * HOURS_IN_MS;
+  } else if (unit == TimeUnit.DAY) {
+    return time * DAYS_IN_MS;
+  } else {
+    throw new Error("Unrecognized time unit");
+  }
+}
```
rugeli, line 103: Maybe use an object to handle the time unit here to avoid multiple `else if`? 
something like:
```
const timeUnitToMilliseconds = {
  [TimeUnit.MILLISECOND]: 1,
  [TimeUnit.SECOND]: SECONDS_IN_MS,
  [TimeUnit.MINUTE]: MINUTES_IN_MS,
  [TimeUnit.HOUR]: HOURS_IN_MS,
  [TimeUnit.DAY]: DAYS_IN_MS,
};

export function timeToMilliseconds(time: number, unit: TimeUnit): number {
  const timeUnitFactor = timeUnitToMilliseconds[unit];
  if (timeUnitFactor === undefined) {
    throw new Error("Unrecognized time unit");
  }
  return time * timeUnitFactor;
}
```

rugeli (approved):
This change is so well-organized with testing, a constants file and a separate utility file! Plus, the comments are great as always. Seems there are some merge conflicts on this branch, and I have one suggestion for a refactor, otherwise lgtm!

src/utils/num_utils.ts
ShrimpCryptid: Makes sense, thank you!

PR #234, opened 2024-08-08
safer type check for perspective camera

toloudis:
time to review: tiny

This change just adds a more robust type check (for whether the camera is in a perspective projection mode), that is in accordance with threejs documentation.  Alternately we could use `!isOrthographicCamera` but why not be even more specific.


ShrimpCryptid (approved):

interim17 (approved):
Briefly thought the property check on .isPerspectiveCamera was an infinitely looping call of the isPerspectiveCamera function and felt scared

PR #237, opened 2024-08-29
Volume-isosurface intersections in raymarch mode

### LINKED ISSUE #106: combine mesh and volume rendering
## Use Case

Viewer supports concurrent display of volume and isosurface mesh but the rendering is not accurate.  We wish to be able to render both in a more accurate way. This is a key piece to have solved before merging volume rendering into simularium viewer. 

## Solution

Meshes render first, into depth buffer.  Volume shader reads from depth buffer once at each fragment and uses the value to modify the termination of the ray march if a mesh has a depth value that would be inside the volume.  In this concept, the mesh surface blocks the volume completely.

frasercl:
Resolves #106: adds proper intersections between volumes and isosurfaces in raymarch mode!

Works as described in the linked issue: meshes render first, then volumes sample the resulting depth texture while rendering.
Uses three's [layers](https://threejs.org/docs/#api/en/core/Layers) system to separate the scene into meshes and volumes.

Additionally, adds a `RenderToBuffer` class like Simularium has for render stages that effectively skip the vertex shader by just rendering a fullscreen quad and passing through uv. (A supplemental PR will apply this class to two render steps in `PathTracedVolume` which also take this approach.) This is required because the depth texture is inaccessible unless you render to a three.js `RenderTarget`, so meshes must be rendered to a `RenderTarget` then copied to the canvas in an additional render step.

## Screenshots

| Before | After |
| ----- | ----- |
| image | image |

To reproduce the above images:
- `npm start` and open the test viewer
- Use the dropdown immediately below the viewport to switch to "OME Zarr (Variance)"
- Modify channel settings using the strip of controls to the right:
  - In "Channel Membrane," uncheck "enabled," check "isosurface," and set "isovalue" to 14
  - In "Channel Labeled Structure," click the "pct50_98" button
  - In "Channel DNA," uncheck "enabled"


src/RenderToBuffer.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.scene.add(this.mesh);
+
+    this.camera = new OrthographicCamera(-1, 1, 1, -1, 0, 1);
+  }
+
+  public render(renderer: WebGLRenderer, target?: WebGLRenderTarget) {
+    renderer.setRenderTarget(target ?? null);
+    renderer.render(this.scene, this.camera);
+    renderer.setRenderTarget(null);
+  }
```
ShrimpCryptid, line 44: When would you want to render without a render target?

src/constants/shaders/raymarch.frag
```diff
@@ -280,10 +279,27 @@ void main() {
...
+  float depth = texture2D(textureDepth, vUv).r;
+  // If there's a depth-contributing mesh at this fragment, we may need to terminate the ray early
+  if (depth < 1.0) {
+    // Get a projection space position from depth and uv, and unproject back to object space
+    vec4 meshProj = vec4(vUv * 2.0 - 1.0, depth * 2.0 - 1.0, 1.0);
+    vec4 meshView = inverseProjMatrix * meshProj;
+    vec4 meshObj = inverseModelViewMatrix * vec4(meshView.xyz / meshView.w, 1.0);
+
+    // Derive a t value for the mesh intersection
+    float tMesh = (meshObj.z - eyeRay_o.z) / eyeRay_d.z;
```
ShrimpCryptid, line 292: What do `eyeRay_o` and `eyeRay_d` represent here? Is it origin and depth?

ShrimpCryptid (approved):
Have some basic questions but I'm delighted by how small this code change ended up being! Nice work!

src/constants/shaders/raymarch.frag
frasercl: origin and direction

src/RenderToBuffer.ts
frasercl: `setRenderTarget(null)` means "render to the default canvas"

src/ThreeJsPanel.ts
```diff
@@ -677,7 +703,19 @@ export class ThreeJsPanel {
...
+
+    this.meshRenderToBuffer.material.uniforms.image.value = this.meshRenderTarget.texture;
+    this.meshRenderToBuffer.render(this.renderer);
+
+    this.camera.layers.set(0);
+    this.renderer.setRenderTarget(null);
+    this.renderer.autoClear = false;
+    this.renderer.render(this.scene, this.camera);
+    this.renderer.autoClear = true;
+
```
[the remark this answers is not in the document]
frasercl: Your understanding is right - I've added some docs here!

src/RenderToBuffer.ts
ShrimpCryptid: Gotcha. Maybe include a method comment explaining that?

src/ThreeJsPanel.ts
```diff
@@ -97,6 +108,20 @@ export class ThreeJsPanel {
     }
 
     this.scene = new Scene();
+    this.meshRenderTarget = new WebGLRenderTarget(this.canvas.width, this.canvas.height, {
+      minFilter: NearestFilter,
+      magFilter: NearestFilter,
+      format: RGBAFormat,
+      type: FloatType,
```
toloudis, line 115: is float type really what is wanted here?  does this cause it to create a RGBA32F buffer?  For this basic mesh-rasterizing renderer probably RGBA8U is plenty.

src/constants/shaders/raymarch.frag
toloudis, line 292: a comment here about potential division by 0 and what you might do to avoid it, could be helpful to future programmers looking at this code

src/ThreeJsPanel.ts
frasercl: After a bit more research, it looks like this is possible, but requires an update to three.js to pick up [this change](https://github.com/mrdoob/three.js/pull/22351)

src/constants/lights.ts
```diff
@@ -1,7 +1,7 @@
 import { MathUtils } from "three";
 
 const spotlightSettings = Object.freeze({
-  angle: 6 * MathUtils.DEG2RAD,
```
frasercl, line 4: This fixes an unrelated visual artifact @toloudis and I found while looking at a bunch of isosurfaces:
image

src/ThreeJsPanel.ts
```diff
@@ -677,7 +700,24 @@ export class ThreeJsPanel {
       }
     }
 
+    // RENDERING
+    // Step 1: Render meshes, e.g. isosurfaces, separately to a render target. (Meshes are all on
+    //   layer 1.) This is necessary to access the depth buffer.
+    this.camera.layers.set(1);
```
toloudis, line 706: I'd suggest using a named constant like MESH_LAYER and VOLUME_LAYER so this value can be tracked across files, like in view3d as well

ShrimpCryptid: Seconding this

toloudis (approved):
It's generally good to have clip near and clip far as tight around the scene as possible, usually for purposes of rasterized depth precision, so I like the CLIP_NEAR/CLIP_FAR changes, too.

PR #238, opened 2024-08-30
Refactor: use `RenderToBuffer` class; organize & dedup shaders

frasercl:
Review time: medium (20-30min)

#237 introduced the `RenderToBuffer` class, which abstracts out some cumbersome setup for render pipelines that just draw a fullscreen quad. I created that class because `PathTracedVolume` already had two instances of this pattern, and the linked PR added a third, which felt like a critical mass. But that PR didn't integrate that class into `PathTracedVolume`. That's what this PR is for.

While I was messing with how shaders are invoked, I did a bit of extra shader code organization. Picking up where #119 left off, now truly all shader code (that actually gets used) lives in GLSL files rather than inline in TypeScript, with one exception: a couple very rudimentary shaders that get used more than once have their own home in basicShaders.ts to ensure they aren't imported into TypeScript-land as literal strings more than once.


src/constants/denoiseShader.ts
```diff
@@ -1,152 +1,44 @@
...
-  float LerpQ = (fCount > gDenoiseLerpThreshold) ? gDenoiseLerpC : 1.0f - gDenoiseLerpC;
-
-  clr.rgb = mix(clr.rgb, clr00.rgb, LerpQ);
-  clr.rgb = clamp(clr.rgb, 0.0, 1.0);
-
-  pc_fragColor = vec4(clr.rgb, clr00.a);
-}
-`;
+const DENOISE_WINDOW_RADIUS = 3;
+export const denoiseShaderUniforms = () => ({
```
toloudis, line 7: what happened to UniformsUtils.merge?

frasercl: The point of `UniformsUtils.merge` seems to be to:

1. Make a new, deep clone of the given uniforms objects
2. Merge all the cloned objects into one new object

2 is unnecessary, since the function was just getting an array with one element anyways, and 1 is handled by putting the object in an arrow function. I checked in with the [docs](https://threejs.org/docs/#api/en/renderers/shaders/UniformsUtils) and [source](https://github.com/mrdoob/three.js/blob/17abdb5e4b88368a6c1528743d014e641b32cf85/src/renderers/shaders/UniformsUtils.js#L53) to make sure there wasn't anything else special it was doing before removing it.

src/PathTracedVolume.ts
```diff
@@ -132,148 +110,24 @@ export default class PathTracedVolume implements VolumeRenderImpl {
...
-
-    this.screenTextureMaterial = new ShaderMaterial({
-      uniforms: this.screenTextureShader.uniforms,
-      vertexShader: this.screenTextureShader.vertexShader,
-      fragmentShader: this.screenTextureShader.fragmentShader,
-      depthWrite: false,
-      depthTest: false,
+    this.pathTracingRenderToBuffer = new RenderToBuffer(pathTracingFragmentShaderSrc, this.pathTracingUniforms);
+    this.screenTextureRenderToBuffer = new RenderToBuffer(copyImageFragShader, {
+      image: { value: this.pathTracingRenderTarget.texture },
```
interim17, line 122: If I'm following the general move in this PR is to take a lot of the code and shaders removed from this file and have those tasks handled here by the instances of `RenderToBuffer`

interim17 (approved):
Lot's of moving code! But the primary change of using `RenderToBuffer` as a utility to simplify `PathTracedVolume` LGTM.

Seems to be running well when set to `Pathtrace` locally. Is there anything else you'd like for verification from reviewer? Approving, but lmk if you'd like a check on anything else.

src/PathTracedVolume.ts
frasercl: Pretty much! The objects removed from this class and replaced by these `RenderToBuffer` instances were all related to the boilerplate hoops-jumping you have to do to pretend you have an object you want to render (scene, camera, geometry, material, mesh...) when all you're really trying to tell the renderer to do is "run this shader code for every screen pixel and render the result." Now `RenderToBuffer` puts the hoops-jumping in one place instead of three.

The shader organization element is somewhat separate - I could have swapped in `RenderToBuffer` and still left the shader code it runs in big JS string arrays with `.join("\n");` at the end. But that's not nice either. if your code is instead in its own GLSL file, you can install an editor extension and edit with syntax highlighting like a real programming language.

> Seems to be running well when set to Pathtrace locally. Is there anything else you'd like for verification from reviewer?

Nope! This should be the only spot with a chance of meaningful behavior problems.

rugeli (approved):
Thank you for the helpful conversation above-it helped me better understand the context behind the shader organization. LGTM!

PR #240, opened 2024-09-07
fix: Allow `loadSpecOverride` to override the volume's loadSpec

ShrimpCryptid:
*Estimated review size: tiny, 1 minute*

Fixes a bug where the `loadSpecOverride` didn't actually do what it said it did! I flipped around the spread operators so the override is actually being applied now.

This is blocking https://github.com/allen-cell-animated/website-3d-cell-viewer/pull/321 so a quick review would be appreciated!


frasercl (approved):
whoops oops

PR #247, opened 2024-09-20
Fix/chunks span multiple channels

### LINKED ISSUE #246: cant load ome-zarr when chunks have multiple channels per chunk
## Description

When chunks can span more than 1 channel, our loop over all channels on OmeZarrLoader results in chunks being re-requested.
SubscribableRequestQueue considers the re-request to be an error and cause for rejection.

## Expected Behavior

This type of chunking should "just work" and zarr should load correctly.

## Reproduction

https://allen-genetic-tools.s3.amazonaws.com/epifluorescence/1116405308/ome_zarr_conversion/1116405308.zarr

toloudis:
fixes #246 
The main issue was re-queueing on SubscribableRequestQueue because the viewer wants to organize things by Channel as a high level way of fetching data.  Because of this it is possible to request the same chunk several times at once.

There's also a small bug fix for indexing into the scale values.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -383,9 +383,9 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
     });
 
     // for physicalPixelSize, we use the scale of the first level
-    const scale5d = this.getScale(0);
+    const scale5d: TCZYX<number> = this.getScale(0);
     // assume that ImageInfo wants the timeScale of level 0
-    const timeScale = hasT ? scale5d[t] : 1;
+    const timeScale = hasT ? scale5d[0] : 1;
 
     const imgdata: ImageInfo = {
       name: source0.omeroMetadata?.name || "Volume",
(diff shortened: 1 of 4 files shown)


src/test/SubscribableRequestQueue.test.ts
```diff
@@ -106,17 +106,18 @@ describe("SubscribableRequestQueue", () => {
       expect(result2).to.equal("bar");
     });
 
-    it("rejects and reissues a request when one subscriber queues the same key twice", async () => {
+    it("alows a subscriber to queue the same key twice", async () => {
```
ShrimpCryptid, line 109:
```suggestion
    it("allows a subscriber to queue the same key twice", async () => {
```

src/loaders/OmeZarrLoader.ts
```diff
@@ -395,7 +395,7 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
       volumeSize: pxSizeLv,
       subregionSize: pxSizeLv.clone(),
       subregionOffset: new Vector3(0, 0, 0),
-      physicalPixelSize: new Vector3(scale5d[x], scale5d[y], hasZ ? scale5d[z] : Math.min(scale5d[x], scale5d[y])),
+      physicalPixelSize: new Vector3(scale5d[4], scale5d[3], hasZ ? scale5d[2] : Math.min(scale5d[4], scale5d[3])),
```
ShrimpCryptid, line 398: What was the bug that this fixes?

toloudis: if the data was TZYX then the zyx indices can be wrong (as opposed to TCZYX).  The scale5d variable is always in TCZYX order, so you can use explicit indexing.  That's why the variable has "5d" in its name and also why I added a little type hint where the variable is declared above.

ShrimpCryptid: Gotcha, thanks for the explanation!

interim17 (approved):
Thanks to @frasercl your comment was clarifying. Haven't done much to test this, but in general LGTM

frasercl (approved):
Looking good now!

PR #252, opened 2024-10-07
Feature/redundant imageinfo fields

### LINKED ISSUE #251: remove redundant imageinfo fields
## Use Case

From @frasercl :
So it's in writing which fields are specifically redundant: volumeSize, numChannels, and times should now be derivable from multiscaleLevelDims and multiscaleLevel (which I actually kind of like better). and originalSize might just be multiscaleLevelDims[0].shape depending on whether level 0 is always considered the "original" (currently not the case with JSON vols, and maybe not with ZARR either)

## Solution

remove redundancies in ImageInfo

toloudis:
REFACTOR:  resolves #251 in which after adding multiscaleLevelDims, several fields in ImageInfo were totally redundant.

This should not change any behaviors in usage of the viewer.

Changes:

* Move `ImageInfo` and `VolumeDims` into their own modules.
* Remove redundant fields from `ImageInfo` and use convenience functions in a wrapper class to return the relevant data back out.
* the dtype in `VolumeDims` gets better typechecking.
* New `ImageInfo` uses only array types instead of Vector2/Vector3.  Note that the `transform` falls under that group.  This is really useful for effectively and efficiently passing the ImageInfo object across worker boundaries.

Aside/discussion:

This shows that even though there is one single `ImageInfo` per `Volume`, that doesn't take into account the following:
* multiple data sources
* subregion selection (See commented-out code in omezarrloader)
* some of the physical size variables stored in Volume seem redundant or unnecessary
So there is still some thinking to do with respect to what ImageInfo really is supposed to contain, and whether it should be split apart, perhaps along the lines of read-only info from the data sources vs app-selections of what to load.

Tested for compatibility with main branch of website-3d-cell-viewer.

### DIFF
--- a/src/ImageInfo.ts
+++ b/src/ImageInfo.ts
@@ -1,5 +1,5 @@
 import { type VolumeDims2, volumeSize, physicalPixelSize } from "./VolumeDims";
-import { Vector3 } from "three";
+import { Vector3, Vector2 } from "three";
 
 export type ImageInfo2 = Readonly<{
   name: string;
@@ -13,9 +13,9 @@ export type ImageInfo2 = Readonly<{
   atlasTileDims: [number, number];
   /** Size of the volume (current level), in pixels */
   //volumeSize: Vector3;
-  /** Size of the currently loaded subregion, in pixels */
+  /** Size of the currently loaded subregion, in pixels, in XYZ order */
   subregionSize: [number, number, number];
-  /** Offset of the loaded subregion into the total volume, in pixels */
+  /** Offset of the loaded subregion into the total volume, in pixels, in XYZ order */
   subregionOffset: [number, number, number];
   /** Size of a single *original* (not downsampled) pixel, in spatial units */
   //physicalPixelSize: Vector3;
@@ -26,7 +26,7 @@ export type ImageInfo2 = Readonly<{
   combinedNumChannels: number;
   /** The names of each channel */
   channelNames: string[];
-  /** Optional overrides to default channel colors, in 0-255 range */
+  /** Optional overrides to default channel colors, in 0-255 range, RGB order */
   channelColors?: [number, number, number][];
 
   /** Number of timesteps in the time series, or 1 if the image is not a time series */
@@ -61,7 +61,7 @@ export type ImageInfo2 = Readonly<{
   multiscaleLevel: number;
 
   transform: {
-    /** Translation of the volume from the center of space, in volume voxels */
+    /** Translation of the volume from the center of space, in volume voxels in XYZ order */
     translation: [number, number, number];
     /** Rotation of the volume in Euler angles, applied in XYZ order */
     rotation: [number, number, number];
@@ -99,6 +99,7 @@ export function defaultImageInfo(): ImageInfo2 {
 
 export class CImageInfo {
   imageInfo: ImageInfo2;
+
   constructor(imageInfo?: ImageInfo2) {
     this.imageInfo = imageInfo || defaultImageInfo();
   }
@@ -136,4 +137,36 @@ export class CImageInfo {
   get numMultiscaleLevels(): number {
     return this.imageInfo.multiscaleLevelDims.length;
   }
+
+  get channelNames(): string[] {
+    return this.imageInfo.channelNames;
+  }
+  get channelColors(): [number, number, number][] | undefined {
+    return this.imageInfo.channelColors;
+  }
+  get subregionSize(): Vector3 {
+    return new Vector3(...this.imageInfo.subregionSize);
+  }
+  get subregionOffset(): Vector3 {
+    return new Vector3(...this.imageInfo.subregionOffset);
+  }
+  get multiscaleLevel(): number {
+    return this.imageInfo.multiscaleLevel;
+  }
+  get atlasTileDims(): Vector2 {
+    return new Vector2(...this.imageInfo.atlasTileDims);
+  }
+  get transform(): { translation: Vector3; rotation: Vector3 } {
+    return {
+      translation: new Vector3(...this.imageInfo.transform.translation),
+      rotation: new Vector3(...this.imageInfo.transform.rotation),
+    };
+  }
+}
+
+export function computeAtlasSize(imageInfo: ImageInfo2): [number, number] {
+  const { atlasTileDims } = imageInfo;
+  const volDims = imageInfo.multiscaleLevelDims[imageInfo.multiscaleLevel];
+  // TCZYX: 4 = x, 3 = y
+  return [atlasTileDims[0] * volDims[4], atlasTileDims[1] * volDims[3]];
 }
--- a/src/ImageInfo.ts
+++ b/src/ImageInfo.ts
@@ -22,7 +22,10 @@ export type ImageInfo2 = Readonly<{
   /** Symbol of physical spatial unit used by `pixelSize` */
   //spatialUnit: string;
 
-  /** Number of channels in the image, accounting for convergence of multiple sources */
+  /** Number of channels in the image, accounting for convergence of multiple sources.
+   * Because of multiple sources, which is not accounted for in ImageInfo2,
+   * that this could be different than the number of channels in the multiscaleLevelDims.
+   */
   combinedNumChannels: number;
   /** The names of each channel */
   channelNames: string[];
@@ -54,7 +57,6 @@ export type ImageInfo2 = Readonly<{
   /** Number of scale levels available for this volume */
   //numMultiscaleLevels: number;
   /** Dimensions of each scale level, at original size, from the first data source */
-  // TODO THIS DATA IS SOMEWHAT REDUNDANT WITH SOME OF THE OTHER FIELDS IN HERE
   multiscaleLevelDims: VolumeDims2[];
 
   /** The scale level from which this image was loaded, between `0` and `numMultiscaleLevels-1` */
--- a/src/ImageInfo.ts
+++ b/src/ImageInfo.ts
@@ -1,4 +1,4 @@
-import { type VolumeDims2, volumeSize, physicalPixelSize } from "./VolumeDims.js";
+import { type VolumeDims, volumeSize, physicalPixelSize } from "./VolumeDims.js";
 import { Vector3, Vector2 } from "three";
 
 export type ImageInfo = Readonly<{
@@ -58,7 +58,7 @@ export type ImageInfo = Readonly<{
   /** Number of scale levels available for this volume */
   //numMultiscaleLevels: number;
   /** Dimensions of each scale level, at original size, from the first data source */
-  multiscaleLevelDims: VolumeDims2[];
+  multiscaleLevelDims: VolumeDims[];
 
   /** The scale level from which this image was loaded, between `0` and `numMultiscaleLevels-1` */
   multiscaleLevel: number;
@@ -107,7 +107,7 @@ export class CImageInfo {
     this.imageInfo = imageInfo || defaultImageInfo();
   }
 
-  get currentLevelDims(): VolumeDims2 {
+  get currentLevelDims(): VolumeDims {
     return this.imageInfo.multiscaleLevelDims[this.imageInfo.multiscaleLevel];
   }
 
--- a/src/ImageInfo.ts
+++ b/src/ImageInfo.ts
@@ -4,23 +4,15 @@ import { Vector3, Vector2 } from "three";
 export type ImageInfo = Readonly<{
   name: string;
 
-  /** XYZ size of the *original* (level 0) volume, in pixels */
-  //originalSize: Vector3;
   /**
    * XY dimensions of the texture atlas used by `RayMarchedAtlasVolume` and `Atlas2DSlice`, in number of z-slice
    * tiles (not pixels). Chosen by the loader to lay out the 3D volume in the squarest possible 2D texture atlas.
    */
   atlasTileDims: [number, number];
-  /** Size of the volume (current level), in pixels */
-  //volumeSize: Vector3;
   /** Size of the currently loaded subregion, in pixels, in XYZ order */
   subregionSize: [number, number, number];
   /** Offset of the loaded subregion into the total volume, in pixels, in XYZ order */
   subregionOffset: [number, number, number];
-  /** Size of a single *original* (not downsampled) pixel, in spatial units */
-  //physicalPixelSize: Vector3;
-  /** Symbol of physical spatial unit used by `pixelSize` */
-  //spatialUnit: string;
 
   /** Number of channels in the image, accounting for convergence of multiple sources.
    * Because of multiple sources, which is not accounted for in ImageInfo,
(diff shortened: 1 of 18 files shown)


src/ImageInfo.ts
```diff
@@ -0,0 +1,223 @@
...
+  subregionSize: [number, number, number];
+  /** Offset of the loaded subregion into the total volume, in pixels, in XYZ order */
+  subregionOffset: [number, number, number];
+  /** Size of a single *original* (not downsampled) pixel, in spatial units */
+  //physicalPixelSize: Vector3;
+  /** Symbol of physical spatial unit used by `pixelSize` */
+  //spatialUnit: string;
+
+  /** Number of channels in the image, accounting for convergence of multiple sources.
+   * Because of multiple sources, which is not accounted for in ImageInfo2,
```
toloudis, line 26:
```suggestion
   * Because of multiple sources, which is not accounted for in ImageInfo,
```

src/ImageInfo.ts
```diff
@@ -0,0 +1,223 @@
...
+  subregionOffset: [number, number, number];
+  /** Size of a single *original* (not downsampled) pixel, in spatial units */
+  //physicalPixelSize: Vector3;
+  /** Symbol of physical spatial unit used by `pixelSize` */
+  //spatialUnit: string;
+
+  /** Number of channels in the image, accounting for convergence of multiple sources.
+   * Because of multiple sources, which is not accounted for in ImageInfo,
+   * that this could be different than the number of channels in the multiscaleLevelDims.
+   * NOTE Currently there is one ImageInfo2 per Volume, not per source.
```
toloudis, line 28:
```suggestion
   * NOTE Currently there is one ImageInfo per Volume, not per source.
```

src/Volume.ts
```diff
@@ -1,170 +1,28 @@
...
-    "rotation": new Vector3(0, 3.14159, 1.57),
-  },
-  };
- */
-
 /**
  * A renderable multichannel volume image with 8-bits per channel intensity values.
  * @class
- * @param {ImageInfo} imageInfo
+ * @param {ImageInfo2} imageInfo
```
toloudis, line 22:
```suggestion
 * @param {ImageInfo} imageInfo
```

src/workers/util.ts
```diff
@@ -14,15 +14,15 @@ export function rebuildLoadSpec(spec: LoadSpec): LoadSpec {
...
+    //originalSize: new Vector3().copy(imageInfo.originalSize),
+    //atlasTileDims: new Vector2().copy(imageInfo.atlasTileDims),
+    //volumeSize: new Vector3().copy(imageInfo.volumeSize),
+    //subregionSize: new Vector3().copy(imageInfo.subregionSize),
+    //subregionOffset: new Vector3().copy(imageInfo.subregionOffset),
+    //physicalPixelSize: new Vector3().copy(imageInfo.physicalPixelSize),
+    //transform: {
+    //  translation: new Vector3().copy(imageInfo.transform.translation),
+    //  rotation: new Vector3().copy(imageInfo.transform.rotation),
+    //},
```
toloudis, line 26: rebuildImageInfo can go away?

toloudis: maybe the same for LoadSpec in separate PR

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -67,46 +67,59 @@ type JsonImageInfo = {
...
-      : new Vector3(0, 0, 0),
-    rotation: json.transform?.rotation ? new Vector3().fromArray(json.transform.rotation) : new Vector3(0, 0, 0),
-  },
 
-  userData: json.userData,
-});
+    userData: {
+      ...json.userData,
+      // for metadata display reasons
+      originalVolumeSize: [json.width, json.height, json.tiles],
```
toloudis, line 119: put physical_pixel_size here too

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -145,11 +158,13 @@ class JsonImageInfoLoader extends ThreadableVolumeLoader {
...
     const jsonInfo = await this.getJsonImageInfo(loadSpec.time);
 
-    const d = new VolumeDims();
-    d.shape = [jsonInfo.times || 1, jsonInfo.channels, jsonInfo.tiles, jsonInfo.tile_height, jsonInfo.tile_width];
-    d.spacing = [1, 1, jsonInfo.pixel_size_z, jsonInfo.pixel_size_y, jsonInfo.pixel_size_x];
-    d.spaceUnit = jsonInfo.pixel_size_unit || "μm";
-    d.dataType = "uint8";
+    const d: VolumeDims = {
+      shape: [jsonInfo.times || 1, jsonInfo.channels, jsonInfo.tiles, jsonInfo.tile_height, jsonInfo.tile_width],
+      spacing: [1, 1, jsonInfo.pixel_size_z, jsonInfo.pixel_size_y, jsonInfo.pixel_size_x],
```
toloudis, line 163: rescale these values!!

src/loaders/TiffLoader.ts
```diff
@@ -167,36 +169,37 @@ class TiffLoader extends ThreadableVolumeLoader {
...
+      //timeScale: 1,
+      //timeUnit: "",
 
-      numMultiscaleLevels: 1,
+      //numMultiscaleLevels: 1,
       multiscaleLevel: 0,
       multiscaleLevelDims: [
         {
           shape: [dims.sizet, dims.sizec, dims.sizez, dims.sizey, dims.sizex],
           spacing: [1, 1, dims.pixelsizez, dims.pixelsizey, dims.pixelsizex],
```
toloudis, line 192: make sure scale is consistent with shape

src/loaders/TiffLoader.ts
```diff
@@ -140,11 +139,14 @@ class TiffLoader extends ThreadableVolumeLoader {
...
-    d.shape = [dims.sizet, dims.sizec, dims.sizez, dims.sizey, dims.sizex];
-    d.spacing = [1, 1, dims.pixelsizez, dims.pixelsizey, dims.pixelsizex];
-    d.spaceUnit = dims.unit ? dims.unit : "micron";
-    d.dataType = dims.pixeltype ? dims.pixeltype : "uint8";
+    const d: VolumeDims = {
+      shape: [dims.sizet, dims.sizec, dims.sizez, dims.sizey, dims.sizex],
+      spacing: [1, 1, dims.pixelsizez, dims.pixelsizey, dims.pixelsizex],
+      spaceUnit: dims.unit ? dims.unit : "micron",
+      // TODO reconcile OMEDims pixel type with NumberType
+      dataType: dims.pixeltype ? (dims.pixeltype as NumberType) : "uint8",
```
toloudis, line 147: confirm that the strings are compatible

src/ImageInfo.ts
```diff
@@ -0,0 +1,223 @@
...
+  /** Number of channels in the image */
+  get numChannels(): number {
+    return this.imageInfo.combinedNumChannels;
+    // // 1 is C
+    // return this.currentLevelDims.shape[1];
+  }
+
+  /** XYZ size of the *original* (not downsampled) volume, in pixels */
+  get originalSize(): Vector3 {
+    return volumeSize(this.imageInfo.multiscaleLevelDims[0]);
```
toloudis, line 123: account for subregion? return subregion size equivalent in level 0 pixels???

src/loaders/OmeZarrLoader.ts
```diff
@@ -358,11 +359,14 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
     if (!this.maxExtent) {
       this.maxExtent = loadSpec.subregion.clone();
     }
-    const pxDims0 = convertSubregionToPixels(
-      loadSpec.subregion,
-      new Vector3(shape0[x], shape0[y], hasZ ? shape0[z] : 1)
-    );
-    const pxSize0 = pxDims0.getSize(new Vector3());
+    // TODO IS IT SAFE TO IGNORE THIS SUBREGION STUFF?
```
toloudis, line 362: probably not ok but need to figure out where to store the subregion-scaled volume size

meganrm (approved):
This looks like a real improvement to me, with the caveat that I don't feel familiar enough with this to catch anything that could cause problems down the line

src/loaders/OpenCellLoader.ts
```diff
@@ -1,23 +1,18 @@
...
-  ThreadableVolumeLoader,
-  LoadSpec,
-  RawChannelDataCallback,
-  VolumeDims,
-  LoadedVolumeInfo,
-} from "./IVolumeLoader.js";
-import { ImageInfo } from "../Volume.js";
+import { ThreadableVolumeLoader, LoadSpec, RawChannelDataCallback, LoadedVolumeInfo } from "./IVolumeLoader.js";
+import { computeAtlasSize, ImageInfo } from "../ImageInfo.js";
+import { VolumeDims } from "../VolumeDims.js";
```
frasercl, line 3: sorry
```suggestion
import { computeAtlasSize, type ImageInfo } from "../ImageInfo.js";
import type { VolumeDims } from "../VolumeDims.js";
```

src/loaders/RawArrayLoader.ts
```diff
@@ -4,11 +4,11 @@ import {
   ThreadableVolumeLoader,
   type LoadSpec,
   type RawChannelDataCallback,
-  VolumeDims,
   type LoadedVolumeInfo,
 } from "./IVolumeLoader.js";
 import { computePackedAtlasDims } from "./VolumeLoaderUtils.js";
-import { ImageInfo } from "../Volume.js";
+import { ImageInfo } from "../ImageInfo.js";
+import { VolumeDims } from "../VolumeDims.js";
```
frasercl, line 11: sorry
```suggestion
import type { ImageInfo } from "../ImageInfo.js";
import type { VolumeDims } from "../VolumeDims.js";
```

src/loaders/TiffLoader.ts
```diff
@@ -49,6 +48,24 @@ class OMEDims {
...
+    uint32: "uint32",
+    int8: "int8",
+    int16: "int16",
+    int32: "int32",
+    float: "float32",
+  };
+  const dtype = mapping[omepixeltype];
+  if (dtype === undefined) {
+    // TODO consider throwing an error?
+    return "uint8";
```
frasercl, line 64: Might be good to log this at least? "unrecognized TIFF pixel type; defaulting to 'uint8'" so that if that guess is wrong and causes an error later, folks are more likely to know where to actually look

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -67,46 +67,55 @@ type JsonImageInfo = {
...
-    translation: json.transform?.translation
-      ? new Vector3().fromArray(json.transform.translation)
-      : new Vector3(0, 0, 0),
-    rotation: json.transform?.rotation ? new Vector3().fromArray(json.transform.rotation) : new Vector3(0, 0, 0),
-  },
+const convertImageInfo = (json: JsonImageInfo): ImageInfo => {
+  const [px, py, pz] = rescalePixelSize(json);
+  // translation is in pixels that are in the space of json.width, json.height.
+  // We need to convert this to the space of the tile_width and tile_height.
+  const tr: [number, number, number] = json.transform?.translation ? json.transform.translation : [0, 0, 0];
```
frasercl, line 84: can use nullish coalescing operator here?

```suggestion
  const tr: [number, number, number] = json.transform?.translation ?? [0, 0, 0];
```

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -145,11 +154,15 @@ class JsonImageInfoLoader extends ThreadableVolumeLoader {
...
-    d.spaceUnit = jsonInfo.pixel_size_unit || "μm";
-    d.dataType = "uint8";
+    const [px, py, pz] = rescalePixelSize(jsonInfo);
+
+    const d: VolumeDims = {
+      shape: [jsonInfo.times || 1, jsonInfo.channels, jsonInfo.tiles, jsonInfo.tile_height, jsonInfo.tile_width],
+      spacing: [1, 1, pz, py, px],
+      spaceUnit: jsonInfo.pixel_size_unit || "μm",
+      dataType: "uint8",
+      timeUnit: jsonInfo.time_unit || "s",
```
frasercl, line 164: some of these might want to be `??` rather than `||` too, though `jsonInfo.times` is at least a bit subtle because `0` is also a problematic value

frasercl (approved):
Looks good overall! Always love seeing redundancies get cleaned up.

Sorry for all the suggestions to put `type` in imports... seems like maybe TS should just take care of that for you huh

src/workers/util.ts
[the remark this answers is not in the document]
toloudis: out of scope but I agree for another small PR. will add issue so I don't forget

src/test/volume.test.ts
```diff
@@ -1,25 +1,20 @@
...
-  subregionSize: new Vector3(204, 292, 65),
-  subregionOffset: new Vector3(0, 0, 0),
-  physicalPixelSize: new Vector3(0.065, 0.065, 0.29),
-  spatialUnit: "",
-
-  numChannels: 9,
+  atlasTileDims: [7, 10],
+  subregionSize: [204, 292, 65],
+  subregionOffset: [0, 0, 0],
+  combinedNumChannels: 9,
```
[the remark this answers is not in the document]
toloudis: It's mostly helping me remember that there is only one ImageInfo for the whole Volume, but the ImageInfo is not correctly reflecting that there could have been multiple channel data sources.  To handle it correctly we would need separate multiscaledims for all sources, or at least trim them to the intersection or something.  But VolumeDims is meant to represent the whole dims of one zarr.

I could rename it back to numChannels if you like - but that might make one think that numChannels should be a value extracted from multiscaledims, which would be wrong!

src/loaders/VolumeLoaderUtils.ts
```diff
@@ -235,12 +237,13 @@ export function buildDefaultMetadata(imageInfo: ImageInfo): Record<string, unkno
     y: imageInfo.physicalPixelSize.y + imageInfo.spatialUnit,
     z: imageInfo.physicalPixelSize.z + imageInfo.spatialUnit,
   };
-  metadata["Multiresolution levels"] = imageInfo.multiscaleLevelDims;
-  metadata["Channels"] = imageInfo.numChannels;
+  metadata["Multiresolution levels"] = rawImageInfo.multiscaleLevelDims;
+  // TODO decide???? combined or not?
+  metadata["Channels"] = rawImageInfo.combinedNumChannels; //imageInfo.numChannels;
```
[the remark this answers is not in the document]
toloudis: a per-source ImageInfo (or at least multiscale dims), or possibly having ImageInfo account for multiple sources inside it, is a good idea.  not sure how to organize it all yet, but it seems like that would move the code in a positive direction in a future PR

src/loaders/JsonImageInfoLoader.ts
toloudis: Yea I think maybe we want to intentionally always have at least numT=1.  What do you think about handling empty string for time_unit and pixel_size_unit?

frasercl: I think it seems possible that a user might want to have no space/time unit, and would express that with an empty string? Though I know unit strings are read and interpreted in a couple places (@ShrimpCryptid's unit formatter, my unit abbreviator), and can't necessarily predict what either would do with those.

src/Volume.ts
```diff
@@ -179,9 +37,13 @@ export default class Volume {
   public channelNames: string[];
   public channelColorsDefault: [number, number, number][];
 
+  // The maximum of the measurements of 3 axes in physical units (pixels*physicalSize)
   public physicalScale: number;
+  // The physical size of a voxel in the original level 0 volume
   public physicalPixelSize: Vector3;
+  // The physical dims of the whole volume (not accounting for subregion)
   public physicalSize: Vector3;
+  // Normalized physical size of the whole volume (not accounting for subregion)
```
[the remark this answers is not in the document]
toloudis: done!

src/loaders/JsonImageInfoLoader.ts
toloudis: okay `??` it is then. at least for the unit strings.

PR #255, opened 2024-10-23
Changes to support Simularium integration

frasercl:
Review time: medium (20-30min). Most of the diff is package-lock.json!

Misc. changes required to incorporate volume rendering from this codebase into Simularium, including:

- In addition to translation and rotation, volumes may now set their scale. This supports setting the size of volume agents with the Simularium format's "collision radius" field. I also added a field to `ImageInfo.transform` to correspond to this, even though fields of that object are never read within the core package.
- Package now exports formerly internal `VolumeDrawable` type.
- The common `doRender` method of `VolumeRenderImpl` no longer depends on our internal `ThreeJsPanel` class, and instead accepts the three.js objects it was previously getting from it.
- `RayMarchedAtlasVolume` now generates its own fallback empty depth texture on its first render, to allow its render method to be called without a depth texture. (Future work should allow this class to accept the full view space position buffer generated by Simularium in its first render pass in place of a depth texture.) Producing this depth texture currently generates errors in Simularium, which is running a slightly older version of three.
- Updates a bunch of babel-related dependencies and adds `plugin-transform-runtime`, to resolve some conflicts.


src/PathTracedVolume.ts
```diff
@@ -582,11 +588,12 @@ export default class PathTracedVolume implements VolumeRenderImpl {
 
   // 0..1 ranges as input
   updateClipRegion(xmin: number, xmax: number, ymin: number, ymax: number, zmin: number, zmax: number): void {
+    console.log("clip region");
```
ShrimpCryptid, line 591: Delete?

src/RayMarchedAtlasVolume.ts
```diff
@@ -25,14 +30,23 @@ import {
...
 import { VolumeRenderSettings, SettingsFlags } from "./VolumeRenderSettings.js";
 
 const BOUNDING_BOX_DEFAULT_COLOR = new Color(0xffff00);
 
+function createEmptyDepthTexture(renderer: WebGLRenderer): DepthTexture {
+  const depthTexture = new DepthTexture(2, 2);
+  const target = new WebGLRenderTarget(2, 2);
+  target.depthTexture = depthTexture;
+  renderer.setRenderTarget(target);
+  renderer.clear(false, true, false);
```
ShrimpCryptid, line 45: Can you add a comment about what these boolean flags are setting?

src/RayMarchedAtlasVolume.ts
```diff
@@ -73,20 +88,22 @@ export default class RayMarchedAtlasVolume implements VolumeRenderImpl {
...
     this.updateSettings(settings, SettingsFlags.ALL);
+    // TODO this is doing *more* redundant work! Fix?
+    this.updateVolumeDimensions();
   }
 
   public updateVolumeDimensions(): void {
     const { normPhysicalSize, normRegionSize } = this.volume;
     // Set offset
-    this.geometryMesh.position.copy(this.volume.getContentCenter());
+    this.geometryMesh.position.copy(this.volume.getContentCenter().multiply(this.settings.scale));
```
ShrimpCryptid, line 100: Why does a volume's scale affect its offset position? Is the scaling origin always the origin of the scene?

ShrimpCryptid (commented):
Looking really good overall, had a couple questions but once they're clarified I'm good to approve 👍

src/PathTracedVolume.ts
frasercl: sure does

src/RayMarchedAtlasVolume.ts
frasercl: The offset position is relative to the center of the volume, which for this viewer should almost always be the origin of the scene. It's there to compensate for an irregular subregion. E.g. if the subregion in the x-axis were 0.5 - 1.0, `getContentCenter` would return 0.25 in the x-axis, because the center of the mesh has to move from the center of the volume bounding box to halfway to its edge:
image
Since that offset is in world units, it also has to account for the scale of the volume.

ShrimpCryptid (approved):
LGTM!

meganrm (approved):

PR #259, opened 2024-11-12
Volume-mesh intersections with view position texture

frasercl:
review time: small (~5min)

#237 added proper intersections between volumes and meshes. This works in the fragment shader by:

1. sampling the depth buffer
2. converting the depth value and uv to a position in view space
3. converting that position back to object space
4. terminating a ray early if it would run past that object space position

Getting volume-mesh intersections right was motivated in part by plans to incorporate volume rendering into Simularium. But Simularium's initial render pass generates a buffer of view space positions, which let us skip the first two steps above. This change adds support for using that view space buffer.

### DIFF
--- a/src/constants/shaders/raymarch.frag
+++ b/src/constants/shaders/raymarch.frag
@@ -21,6 +21,7 @@ uniform float CLIP_FAR;
 uniform sampler2D textureAtlas;
 uniform sampler2D textureAtlasMask;
 uniform sampler2D textureDepth;
+uniform int usingPositionTexture;
 uniform int BREAK_STEPS;
 uniform float SLICES;
 uniform float isOrtho;
--- a/src/constants/shaders/raymarch.frag
+++ b/src/constants/shaders/raymarch.frag
@@ -280,28 +280,28 @@ void main() {
   float clipNear = 0.0;//-(dot(eyeRay_o.xyz, eyeNorm) + dNear) / dot(eyeRay_d.xyz, eyeNorm);
   float clipFar  = 10000.0;//-(dot(eyeRay_o.xyz,-eyeNorm) + dFar ) / dot(eyeRay_d.xyz,-eyeNorm);
 
-  // Sample the depth texture
-  vec4 depthSample = texture2D(textureDepth, vUv);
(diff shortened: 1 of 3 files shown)


src/constants/shaders/raymarch.frag
```diff
@@ -279,14 +280,22 @@ void main() {
   float clipNear = 0.0;//-(dot(eyeRay_o.xyz, eyeNorm) + dNear) / dot(eyeRay_d.xyz, eyeNorm);
   float clipFar  = 10000.0;//-(dot(eyeRay_o.xyz,-eyeNorm) + dFar ) / dot(eyeRay_d.xyz,-eyeNorm);
 
-  // Sample the depth texture
-  float depth = texture2D(textureDepth, vUv).r;
+  // Sample the depth/position texture
+  // This is EITHER a view space position or a depth value we'll convert to a view space position
+  vec4 meshViewPos = texture2D(textureDepth, vUv);
```
toloudis, line 285: naming is a little weird here as this value might not be in view space yet

src/constants/shaders/raymarch.frag
```diff
@@ -279,14 +280,22 @@ void main() {
...
+  bool hasDepthValue = usingPositionTexture == 0 && meshViewPos.r < 1.0;
+
   // If there's a depth-contributing mesh at this fragment, we may need to terminate the ray early
-  if (depth < 1.0) {
-    // Get a projection space position from depth and uv, and unproject back to object space
-    vec4 meshProj = vec4(vUv * 2.0 - 1.0, depth * 2.0 - 1.0, 1.0);
-    vec4 meshView = inverseProjMatrix * meshProj;
-    vec4 meshObj = inverseModelViewMatrix * vec4(meshView.xyz / meshView.w, 1.0);
+  if (hasDepthValue || (usingPositionTexture == 1 && meshViewPos.a > 0.0)) {
```
toloudis, line 289: I think this could use an extra comment up above to say what r,g,b,a of the position texture are expected to represent.   You might even mention that this is a data format contract with Simularium viewer.

toloudis (approved):
Good stuff!  Looks like you didn't even have to manipulate the value. I recommend a clarifying comment and maybe a variable rename.

PR #260, opened 2024-11-18
naive first attempt not working due to need for range requests

toloudis:
@frasercl FYI for ome-zarr-0.5 (including zarrv3)

I tried updating the metadata parsing code only, and zarrita complained that our WrappedStore doesn't support range requests.  Did a very naive attempt to just add getRange but things are still broken.   At least the metadata does load properly.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -5,7 +5,7 @@ import { get as zarrGet, slice, Slice } from "@zarrita/indexing";
 import { AbsolutePath } from "@zarrita/storage";
 // Importing `FetchStore` from its home subpackage (@zarrita/storage) causes errors.
 // Getting it from the top-level package means we don't get its type. This is also a bug, but it's more acceptable.
-import { FetchStore } from "zarrita";
+import { FetchStore, Store } from "zarrita";
 
 import type { ImageInfo } from "../ImageInfo.js";
 import type { VolumeDims } from "../VolumeDims.js";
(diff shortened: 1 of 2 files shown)


src/loaders/OmeZarrLoader.ts
```diff
@@ -112,6 +114,35 @@ const DEFAULT_FETCH_OPTIONS = {
...
+        });
+      }
+    } else {
+      throw new VolumeLoadError("OME-Zarr metadata missing 'multiscales' or 'ome' field", {
+        type: VolumeLoadErrorType.INVALID_METADATA,
+      });
+    }
+  }
+  return { multiscales: attrs.multiscales as OMEMultiscale[], omero: attrs.omero as OmeroTransitionalMetadata };
+}
```
frasercl, line 144: fwiw a lot of the validation this function does is already done by `validateOMEZarrMetadata` below. This could probably be replaced at the callsite with something like `const metadata = group.attrs.ome ?? group.attrs;` plus the version check (which could also go in `validateOMEZarrMetadata`).

toloudis: Thinking ahead to future versions, I thought it would be good to isolate this in its own function.  And if the schemas diverge even more, we may have to read the data into our own internal structure.  But for now, maybe it can be combined with the validate function somehow?

frasercl: I definitely like the idea of keeping any sort of parsing separate. My thought was that long-term, if we want to do any more complex parsing, that parsing would go in the current validation module, which would get renamed to reflect its new responsibilities. (In a sense, validation is just a special case of parsing where we don't do any transformation.) That could also happen now though. The current type of `validateOMEZarrMetadata` is `(data: unknown, ...) => asserts data is OMEZarrMetadata` - i.e. returns nothing, but if control flow passes through it safely then `data` has been validated. If it's also responsible for that simple transformation, it can just as easily guarantee that you have properly formatted metadata by passing it back like `(data: unknown, ...) => OMEZarrMetadata`.

toloudis:
@frasercl I think this PR is obsolete, right? Can we just close this?

PR #261, opened 2024-11-20
Replace depth texture fallback with position texture

frasercl:
review time: tiny (~5min)

#237 gave the raymarch renderer the ability to clip volumes against meshes, using a depth texture generated during mesh rendering.

#255 made a bunch of changes to the API to support ongoing Simularium integration work, including making this depth texture optional. To keep the raymarcher working without an external depth texture, it gained a dummy depth texture for cases when no depth texture was passed in. But creating an empty depth texture involves a fair bit of annoying hoops-jumping.

#259 allowed the raymarch renderer to accept Simularium's position buffer for the purposes of clipping volumes against meshes, rather than a depth texture. Now we have the option to make ourselves a dummy normal texture rather than a dummy depth texture, and remove the hoops-jumping.

This PR removes the hoops-jumping.

### DIFF
--- a/src/RayMarchedAtlasVolume.ts
+++ b/src/RayMarchedAtlasVolume.ts
@@ -5,6 +5,7 @@ import {
   BufferAttribute,
   BufferGeometry,
   Color,
+  DataTexture,
   DepthTexture,
   Group,
   LineBasicMaterial,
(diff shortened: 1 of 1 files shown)


toloudis (approved):
LGTM

rugeli (approved):
Looks good! Sorry for the delay in getting to this one.

PR #266, opened 2024-12-03
Bug fix for Json images, and remove image

toloudis:
Two very small changes toward fixing some issues in CFE with loading images by clicking Gallery.

1. Volumes were never being truly removed from the View3d.  The problem with that is that you could still call view3d calls that would operate on the view's image even though it had already been "removed".   

2. The JsonImageInfoLoader was requesting the wrong channels when a subset of channels was provided to it.  
2a.  I also added code to respect the syncChannels setting so that time series playback of multi-frame json works better.  One of our test data files uses this.

I also updated a comment to reflect the correct conditions for required channels.

**Note: there will be a companion PR in website-3d-cell-viewer.  These changes would cause breakage without it.**

### DIFF
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -226,6 +226,7 @@ export class View3d {
     if (this.image) {
       this.removeVolume(this.image.volume);
     }
+    this.image = undefined;
   }
 
   /**
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -118,7 +118,8 @@ export default class VolumeDrawable {
   }
 
   /**
-   * Updates whether a channel's data must be loaded for rendering, based on if its volume or isosurface is enabled.
+   * Updates whether a channel's data must be loaded for rendering,
+   * based on if its volume or isosurface is enabled, or whether it is needed for masking.
    * Calls `Volume.updateRequiredData` to update the list of required channels if necessary.
    */
   private updateChannelDataRequired(channelIndex: number): void {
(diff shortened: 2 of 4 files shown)


src/loaders/JsonImageInfoLoader.ts
```diff
@@ -192,7 +197,7 @@ class JsonImageInfoLoader extends ThreadableVolumeLoader {
     const requestedChannels = loadSpec.channels;
     if (requestedChannels) {
       // If only some channels are requested, load only images which contain at least one requested channel
-      images = images.filter(({ channels }) => channels.some((ch) => ch in requestedChannels));
+      images = images.filter(({ channels }) => channels.some((ch) => requestedChannels.includes(ch)));
```
toloudis, line 200: this is the key bugfix for getting the right channels to load

src/View3d.ts
```diff
@@ -226,6 +226,7 @@ export class View3d {
     if (this.image) {
       this.removeVolume(this.image.volume);
     }
+    this.image = undefined;
```
toloudis, line 229: 🎉

src/VolumeDrawable.ts
```diff
@@ -128,12 +129,12 @@ export default class VolumeDrawable {
 
     if (requiredChannels.includes(channelIndex)) {
       if (!channelIsRequired) {
-        // This channel is currently marked required, but both its volume and isosurface are off. Remove it!
+        // This channel is currently marked required, but both its volume and isosurface are off, and it's not a mask. Remove it!
```
frasercl, line 132: nit/optional: width >120

src/loaders/JsonImageInfoLoader.ts
frasercl: (shaking fist at the sky) javascriiiiiipt!

frasercl (approved):

src/loaders/JsonImageInfoLoader.ts
ShrimpCryptid: I had to look this one up, so in this case it was trying to check if `ch` was an index for `requestedChannels` instead of if `ch` was a value?

ShrimpCryptid (approved):

src/loaders/JsonImageInfoLoader.ts
toloudis: right.  "in" roughly means "is a property key for", which for arrays just means the array indices.  This took me a while to track down, because ordinarily we request all channels but I made a change in w3cv to request only required channels - I kept asking for channels [0,1,2,5,7] and it consistently refused to load channel 7!

PR #268, opened 2024-12-06
bump three to r171

frasercl:
Review time: tiny

Bumps three to revision 171.


ShrimpCryptid (approved):

toloudis (approved):
Always a little scary to update three.  Their change logs are extensive.  Just be sure to test isosurfaces and pathtrace as well.

It'll be good to get all our code on the same three version.

PR #270, opened 2024-12-13
fix: Default range causes volumes to have incorrect LUT values

ShrimpCryptid:
Fix for https://github.com/allen-cell-animated/cell-feature-explorer/issues/223, "Volume channels are sometimes very blown out when switching to Full Field mode."

The bug was being caused by the array loaders always assuming a value range from 0 to 255 for uint8 value types. This no longer matched changes to behavior made in #197, where the Lut and Histogram were changed to be relative to a range of min/max value for the raw data. 

This meant that volumes where there were no values for certain numbers would have a mismatch between the onscreen histogram and the actual rendered values. In the case of the described bugged volume, there were no pixels with values from 0-18, which resulted in the rendered ranges looking incorrect.

*Estimated review size: small, 10 minutes*

## Changes
- Changed all array loaders previously depending on `DATARANGE_UINT8` to use `getDataRange()`, a new helper method.
- Added explanatory comments to `Histogram.ts`.

Before:
image

After:
image

### DIFF
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -21,11 +21,11 @@ export default class Histogram {
   private min: number;
   /** Max value in the original raw data. */
   private max: number;
-  /** Size of each histogram bin in the scale of the original raw data. */
+  /** Size of each histogram bin in the scale of the original data. */
   private binSize: number;
-  /** Index of the first nonzero bin with at least 1 sample */
+  /** Index of the first bin (other than 0) with at least 1 value. */
   private dataMinBin: number;
-  /** Index of the last nonzero bin with at least 1 sample */
+  /** Index of the last bin (other than 0) with at least 1 value. */
   private dataMaxBin: number;
   private pixelCount: number;
   public maxBin: number;
@@ -47,6 +47,8 @@ export default class Histogram {
     this.binSize = hinfo.binSize;
 
     // track the first and last nonzero bins with at least 1 sample
+    // TODO: If `this.min` is not 0, should `this.dataMinBin` be `0` (instead of starting at 1)
+    // since the first bin doesn't represent zero values?
     for (let i = 1; i < this.bins.length; i++) {
       if (this.bins[i] > 0) {
         this.dataMinBin = i;
@@ -105,20 +107,19 @@ export default class Histogram {
   }
 
   /**
-   * Returns the minimum bin index with a non-zero value.
+   * Returns the first bin index with at least 1 value, other than the 0th bin.
    * @return {number}
    */
   getMin(): number {
-    // Note that this will currently always return 0, unless NBINS changes.
     return this.dataMinBin;
   }
 
   /**
-   * Returns the maximum bin index with a non-zero value.
+   * Returns the last bin index with at least 1 value, other than the 0th bin.
    * @return {number}
    */
   getMax(): number {
-    // Note that this will always return 255, unless NBINS changes.
+    // Note that this will always return `NBINS - 1`.
     return this.dataMaxBin;
   }
 
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -21,18 +21,12 @@ export default class Histogram {
   private min: number;
   /** Max value in the original raw data. */
   private max: number;
-  /** Size of each histogram bin in the scale of the original raw data. */
+  /** Size of each histogram bin in the scale of the original data. */
   private binSize: number;
-  /** Index of the first nonzero bin with at least 1 sample */
-  private dataMinBin: number;
-  /** Index of the last nonzero bin with at least 1 sample */
-  private dataMaxBin: number;
   private pixelCount: number;
   public maxBin: number;
 
   constructor(data: TypedArray<NumberType>) {
-    this.dataMinBin = 0;
-    this.dataMaxBin = 0;
     this.maxBin = 0;
     this.bins = new Uint32Array();
     this.min = 0;
@@ -45,20 +39,6 @@ export default class Histogram {
     this.min = hinfo.min;
     this.max = hinfo.max;
     this.binSize = hinfo.binSize;
-
-    // track the first and last nonzero bins with at least 1 sample
-    for (let i = 1; i < this.bins.length; i++) {
-      if (this.bins[i] > 0) {
-        this.dataMinBin = i;
-        break;
-      }
-    }
-    for (let i = this.bins.length - 1; i >= 1; i--) {
-      if (this.bins[i] > 0) {
-        this.dataMaxBin = i;
-        break;
-      }
-    }
 
     this.pixelCount = data.length;
 
@@ -105,21 +85,19 @@ export default class Histogram {
   }
 
   /**
-   * Returns the minimum bin index with a non-zero value.
+   * Returns the minimum bin index.
    * @return {number}
    */
   getMin(): number {
-    // Note that this will currently always return 0, unless NBINS changes.
-    return this.dataMinBin;
+    return 0;
   }
 
   /**
-   * Returns the maximum bin index with a non-zero value.
+   * Returns the maximum bin index.
    * @return {number}
    */
   getMax(): number {
-    // Note that this will always return 255, unless NBINS changes.
-    return this.dataMaxBin;
+    return NBINS - 1;
   }
 
   getNumBins(): number {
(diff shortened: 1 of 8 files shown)


src/PathTracedVolume.ts
```diff
@@ -492,11 +492,11 @@ export default class PathTracedVolume implements VolumeRenderImpl {
...
         i,
-        this.volume.channels[channel].histogram.getMax() / 255.0
+        this.volume.channels[channel].histogram.getDataMax() / 255.0
       );
       this.pathTracingUniforms.gIntensityMin.value.setComponent(
         i,
-        this.volume.channels[channel].histogram.getMin() / 255.0
+        this.volume.channels[channel].histogram.getDataMin() / 255.0
       );
     }
```
ShrimpCryptid, line 501: I made a change since this seemed to be relying on actual intensity values, but I did notice that `gIntensityMax` is totally unused in the shaders. @toloudis should this be removed or out of scope?

src/utils/num_utils.ts
```diff
@@ -230,3 +230,13 @@ export function constrainToAxis(
...
 }
+
+export function getDataRange(data: ArrayLike<number>): [number, number] {
+  let min = Infinity;
+  let max = -Infinity;
+  for (let i = 0; i < data.length; i++) {
+    min = Math.min(min, data[i]);
+    max = Math.max(max, data[i]);
+  }
+  return [min, max];
```
ShrimpCryptid, line 241: Note that this will fail for `NaN` values. Do we ever expect those to be in loaded volume data?

src/Histogram.ts
```diff
@@ -17,10 +17,15 @@ type HistogramData = {
 export default class Histogram {
   // no more than 2^32 pixels of any one intensity in the data!?!?!
   private bins: Uint32Array;
+  /** Min value in the original raw data. */
   private min: number;
+  /** Max value in the original raw data. */
   private max: number;
+  /** Size of each histogram bin in the scale of the original raw data. */
   private binSize: number;
+  /** Index of the first nonzero bin with at least 1 sample */
```
toloudis, line 26: what does "nonzero bin with at least 1 sample" mean?  Are "nonzero" and "at least 1 sample" meant to mean the same thing?

src/Histogram.ts
```diff
@@ -17,10 +17,15 @@ type HistogramData = {
 export default class Histogram {
   // no more than 2^32 pixels of any one intensity in the data!?!?!
   private bins: Uint32Array;
+  /** Min value in the original raw data. */
   private min: number;
+  /** Max value in the original raw data. */
   private max: number;
+  /** Size of each histogram bin in the scale of the original raw data. */
```
toloudis, line 24: I think restating "original raw data" is unnecessary here in the above comments. You could just say "data".  A Histogram is being given data values and the histogram is a collection of statistics about those data values.

src/Histogram.ts
```diff
@@ -83,27 +88,37 @@ export default class Histogram {
...
     return this.max;
   }
 
   /**
-   * Return the min data value
+   * Returns the minimum bin index with a non-zero value.
    * @return {number}
    */
   getMin(): number {
+    // Note that this will currently always return 0, unless NBINS changes.
```
toloudis, line 112: why would this return something other than 0 if NBINS changed?

src/PathTracedVolume.ts
toloudis: The change above here seems like it would alter the rendering results for pathtrace. It's being divided by 255 which makes me think it's expected to be in the 0-255 range...  Did you test the before/after?

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -262,14 +262,15 @@ class JsonImageInfoLoader extends ThreadableVolumeLoader {
...
             // if we are synchronizing channels, we need to keep track of the data
             resultChannelIndices.push(chindex);
             resultChannelDtype.push("uint8");
-            resultChannelData.push(new Uint8Array(cacheResult));
-            resultChannelRanges.push(DATARANGE_UINT8);
+            resultChannelData.push(channelData);
+            resultChannelRanges.push(getDataRange(channelData));
           } else {
-            onData([chindex], ["uint8"], [new Uint8Array(cacheResult)], [DATARANGE_UINT8]);
+            onData([chindex], ["uint8"], [channelData], [getDataRange(channelData)]);
```
toloudis, line 273: future idea: it could save a bit of computation if we cached the min and max!

src/loaders/JsonImageInfoLoader.ts
```diff
@@ -308,10 +309,16 @@ class JsonImageInfoLoader extends ThreadableVolumeLoader {
       }
 
       // extract the data
+      const rawRange: [number, number][] = [];
```
toloudis, line 312: I feel like `data` (or `channel`) is a better prefix than `raw` here... raw implies to me that some postprocessing will be done to make it less raw later.

src/utils/num_utils.ts
toloudis: another nice way to do this loop is to initialize min and max to data[0] and then iterate from 1 to length.

toloudis: NaNs are pretty unexpected, I think.  Certainly out of scope for this PR.

src/PathTracedVolume.ts
ShrimpCryptid: The field is only used in two commented-out lines in `pathtrace.frag`...

toloudis: if gIntensityMin and gIntensityMax are totally unused in path trace, then I wouldn't bother to modify this code at all either.

src/Histogram.ts
ShrimpCryptid: This comment was copied from the constructor (`// track the first and last nonzero bins with at least 1 sample`). It used to represent the first bin (excluding the zero-th bin, back when values of `0` were placed into the zero-th bin) that had at least one pixel of that intensity.

Now that I think about it, I think this code is obsolete? `dataMinBin` and `dataMaxBin` can be safely removed, and `getMin()` and `getMax()` can just return 0 and `NBINS - 1` respectively.

ShrimpCryptid: (The comment I had on `getMin()` and `getMax()` were actually incorrect for the current behavior, but returning `0` and `NBINS - 1` *should* be the correct behavior.)

src/Histogram.ts
```diff
@@ -42,6 +47,8 @@ export default class Histogram {
     this.binSize = hinfo.binSize;
 
     // track the first and last nonzero bins with at least 1 sample
+    // TODO: If `this.min` is not 0, should `this.dataMinBin` be `0` (instead of starting at 1)
+    // since the first bin doesn't represent zero values?
```
toloudis, line 51: I guess the original intent of this was: when all the data used to always be 8bit 0-255, and there were always 255 bins, one for each intensity, then you would know that bin 0 always represented only intensity=0.   Now bin 0 could actually have nonzero intensity values in it, so looping from 1 is not necessarily doing the same thing.    

Kinda makes me wonder if we should track the number of "0" data values separately.  Which begs the question: is 0 a special value or not?

frasercl (approved):

src/Histogram.ts
ShrimpCryptid: I'd love to say zero is not a special value... I'll check where this is being used and if it's not really then I'll change it so `dataMinBin` is just `0`.

ShrimpCryptid: Discussion Notes:
- In 8-bit data, the 0th bin always contains only 0 values.
- In 16+ bit data, the 0th bin can contain a mix of 0 and nonzero values.
- If we treat zero values as pixels, this would cause percentile values to behave very differently for certain types of volumes that have a lot of zero-pixels (e.g. segmentations, volumes that have "padding" in the form of zeroes)
  - How do we know if 0 is a "special" value (meaning no data) or a normal value (should be included in data calculations?)
- One solution would be to bin zero pixels separately.

toloudis (approved):

src/Histogram.ts
ShrimpCryptid: After looking at ImageJ, we decided that including the 0th bin should be the correct behavior.

Issue #275, opened 2025-01-15
Multiscene loading


frasercl:
## Use Case

The zarr loader needs to be able to distinguish between different scenes.

## Solution

`setScene(number, url)` callable from the UI thread & results in loading data from the correct scene into the same volume object. (Assume all scenes have same dims). Verify with test data or mock.

PR #279, opened 2025-01-28
Multiple loaders in `VolumeLoaderContext`

frasercl:
# Problem

Since #177, we've supported running volume data loaders in a web worker, to keep loading and rendering from blocking each other. The procedure for this looks like:
1. Create a `VolumeLoaderContext`, which owns the worker.
2. Spawn a loader for some data type (zarr, tiff, etc.) in that context.
3. The context returns a `WorkerLoader`, which has the same interface as any other loader, but just sends messages to/receives messages from the actual loader running in the worker.

`VolumeLoaderContext` was always intended to eventually be a single context in which multiple loaders could run, but the initial implementation in #177 supported just one loader. Now that we're looking at supporting multi-scene images which may combine data from different sources, it's time for an upgrade.

This is most of the solution to #275. The remaining work on that ticket is to come up with some interface that combines multiple loaders into a single logical multiscene data source. That could look like some additional loader combinator, live at the vole-app level, or be handled by the context itself.

# Solution

- Add support for multiple loaders in a single `VolumeLoaderContext`.
- Create a very rudimentary demo of multi-scene loading using this new capability. There's now an additional "scene" input to the right of the time input, and the existing "zarrVariance" test image now has two scenes that can be switched between with this box. I expect to revert this pretty quickly in the next phase of this work (to be replaced with the "proper" solution to multi-scene data sources), but it was useful for verifying that the changes I had made were working properly.

### The scene input box is this one:

image


public/index.ts
```diff
@@ -1094,7 +1121,8 @@ async function loadTestData(testdata: TestDataSpec) {
 
   const loadSpec = new LoadSpec();
   myState.totalFrames = testdata.times;
-  loadVolume(loadSpec, myState.loader);
+  const loader = Array.isArray(myState.loader) ? myState.loader[myState.scene] : myState.loader;
+  loadVolume(loadSpec, loader);
```
meganrm, line 1125: I'm not sure I'm following the full data flow yet (just started reading through this). but I'm seeing a lot of `isArray` checks because it may be a single scene or multiple scenes. But it seems like you should be able to convert the data to an array on entry and just sometimes deal with an array of length 1? I just think it's cleaner to have a consistent data type.

src/workers/VolumeLoaderContext.ts
```diff
@@ -173,9 +174,25 @@ class VolumeLoaderContext {
...
+        } else {
+          loader.onChannelData(e);
+        }
+      } else if (e.eventType === WorkerEventType.METADATA_UPDATE) {
+        loader.onUpdateMetadata(e);
+      }
+    }
+  }
+
+  private;
```
ascibisz, line 194: What does this standalone `private;` do?

frasercl: What indeed?

(Nothing. It's gone.)

ascibisz (approved):

ShrimpCryptid (approved):
Left one small nit but otherwise LGTM! The global message type vs. message type w/ loader IDs is clever.

PR #280, opened 2025-01-29
Bump @types/three to r171

frasercl:
Estimated review time: small (~5min)

# Problem

The version of three (r171) is out of sync with @types/three (r144).

# Solution

Bump @types/three to r171, and fix associated type errors.


ShrimpCryptid (approved):

src/VolumeRenderImpl.ts
```diff
@@ -19,7 +19,7 @@ export interface VolumeRenderImpl {
   doRender: (
     renderer: WebGLRenderer,
     camera: PerspectiveCamera | OrthographicCamera,
-    depthTexture?: DepthTexture | Texture
+    depthTexture?: DepthTexture | Texture | null
```
toloudis, line 22: null and undefined are both possible values for all these?

toloudis (approved):

src/VolumeRenderImpl.ts
frasercl: three.js now represents the absence of a depth texture on a render target as `null`, but passing in depth textures is now also part of the public interface used by the simularium volumes implementation, and since leaving this property entirely missing has the same effect it makes sense to keep that option in there.

PR #289, opened 2025-02-13
doc: Updated README text and added NPM links, removed outdated publishing steps

ShrimpCryptid:
Added a link to our NPM release and removed some outdated steps in the `DEV.md` documentation.

You can view the new README here: https://github.com/allen-cell-animated/vole-core/blob/7b3a6113bf4824846aef97a715ae1b4ac35fd400/README.md

*Estimated review size: tiny, <5 minutes*

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -13,8 +13,9 @@
 
 - Requires that you are listed as a maintainer for this npm module
 
-* Commit and push changes
-* Login: `npm login`
 * Make sure you can run build successfully: `npm run build`
 * Update version: `npm version patch`
-* Publish: `npm publish`
+* Push to main: `git push origin main`
+* Push tags: `git push origin --tags vX.X.X`
+
+You do not need to run `npm publish` as the CI/CD pipeline will automatically publish the package to npm.
--- a/DEV.md
+++ b/DEV.md
@@ -11,11 +11,13 @@
 
 ## Publishing
 
-- Requires that you are listed as a maintainer for this npm module
+Requires that you have write access to `main` in this repository on GitHub. If
+you do not have write access, ask someone who does to publish a new version for
+you.
 
-* Make sure you can run build successfully: `npm run build`
-* Update version: `npm version patch`
-* Push to main: `git push origin main`
-* Push tags: `git push origin --tags vX.X.X`
+- Make sure you can run build successfully: `npm run build`
+- Update version: `npm version {major|minor|patch}`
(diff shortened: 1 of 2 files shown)


DEV.md
```diff
@@ -13,8 +13,9 @@
 
 - Requires that you are listed as a maintainer for this npm module
 
-* Commit and push changes
-* Login: `npm login`
 * Make sure you can run build successfully: `npm run build`
 * Update version: `npm version patch`
-* Publish: `npm publish`
+* Push to main: `git push origin main`
+* Push tags: `git push origin --tags vX.X.X`
```
rugeli, line 19: two questions: 
1. Is the main branch protected? Should we create PRs for pushing tags, or is it safe to push directly to main? 
2. Should we specify the commands for pushing tags:
    - push all tags: `git push origin --tags`
    - push a single tag: `git push origin vX.X.X`

DEV.md
```diff
@@ -13,8 +13,9 @@
 
 - Requires that you are listed as a maintainer for this npm module
 
-* Commit and push changes
-* Login: `npm login`
 * Make sure you can run build successfully: `npm run build`
 * Update version: `npm version patch`
```
frasercl, line 17: Should this indicate other version types you could create? `npm version {major|minor|patch}`?

DEV.md
```diff
@@ -13,8 +13,9 @@
 
 - Requires that you are listed as a maintainer for this npm module
```
frasercl, line 14: Does this mean npm package or GitHub repo?

frasercl (approved):
My comments are not on your changes &mdash; this is still an improvement if it merges as-is.

DEV.md
frasercl: I know these are the instructions to publish to NPM, but it seems like this is a statement about which permissions you need to do that? If so, I don't know that I've ever been granted special privileges via NPM, but I imagine you need to be a maintainer on this project to run this action.

ShrimpCryptid: The action runs automatically on a new version tag being pushed to this repo.

https://github.com/allen-cell-animated/vole-core/blob/main/.github/workflows/publish.yml

Though I'm wondering if anyone can push a tag and have the NPM version update... Let me test this

DEV.md
ShrimpCryptid: 1. Main is protected.
2. Good point! Changed this to push a single tag.

PR #291, opened 2025-02-21
Cache uncompressed chunks

### LINKED ISSUE #264: cache uncompressed chunks
## Use Case

The current chunk cache stores the fetched blobs in cache. These chunks are typically compressed.  Whenever we have a cache hit, we have to pay for decompression every time.
One way we can make playback mode faster is to store chunks decompressed.
Accept a cost of memory consumption, for the tradeoff of faster cache hits at run time.

## Solution

We use a Store class that inherits from Zarrita's Store. 
We preliminarily think this might require forking Zarrita to add the hooks or expose the internal modules we need.

Measure performance before and after.

frasercl:
Review time: small-medium (~15-20min)

# Problem

Closes #264: moves OME-Zarr chunk caching from before chunks are decompressed to after, making playback of cached time series volumes significantly faster!

# Quick background

zarrita.js, the library we use to read zarr datasets, has a few important abstraction layers:
- `Store`s are the most basic &mdash; just a plain source of data keyed by strings
- `Array`s wrap `Store`s and add awareness of the individual sizes, shapes, chunking strategies, and (importantly) encodings of specific arrays. They allow querying single chunks by their coordinates.
- The `get` function from `@zarrita/indexing` accepts an `Array` and can extract an arbitrary subset of data from it, by querying many chunks and assembling data from them into a larger typed array.

Our previous strategy for adding caching slotted in between the `Store` and the `Array`, where chunks were not yet decoded. To cache decoded data, we instead need to get between the `Array` and the indexing function.

# Solution

This PR removes the cache connection in `WrappedStore` (that object sticks around to queue requests, though zarrita has a place for a queue in the indexing function, so `WrappedStore` could now go away entirely in a future cleanup PR?) and introduces a function called `cachingArray`, which connects an `Array` to the cache. This works via an ES `Proxy` object, which is a bit awkward, but seemed to be the least bad option:

- Extending the `Array` class wasn't great because arrays are mainly produced by methods on other objects from zarrita, not their own constructors, and we can't make those methods create instances of our special custom extended array instead
- Creating a wrapper class with exactly the same interface as the base `Array` class would have involved a lot of boilerplate and been more vulnerable to interface changes later down the line
- Proxies have a bit of subtlety when [combined with objects with private fields](https://lea.verou.me/blog/2023/04/private-fields-considered-harmful/), but this can be worked around for this case

This PR also adds some additional guardrails around a possible infinite loop condition in our current prefetching implementation. Future work should try to make this more robust.

## Steps to Verify:

1. Select a volume with a time series (e.g. "OME-Zarr (NucMorph)").
2. Play through time a bit to get data into the cache.
3. Skip back to the beginning and play again. *This should be noticeably faster on this branch than on main.*

### DIFF
--- a/src/VolumeCache.ts
+++ b/src/VolumeCache.ts
@@ -1,7 +1,7 @@
 import { Chunk, DataType } from "@zarrita/core";
 
 type MaybeCacheEntry = CacheEntry | null;
-type CacheData = ArrayBuffer | Chunk<DataType>;
+export type CacheData = ArrayBuffer | Chunk<DataType>;
 type CacheEntry = {
   /** The data contained in this entry */
   data: CacheData;
--- a/src/VolumeCache.ts
+++ b/src/VolumeCache.ts
@@ -1,5 +1,7 @@
 import { Chunk, DataType } from "@zarrita/core";
 
+export const isChunk = (data: CacheData): data is Chunk<DataType> => (data as Chunk<DataType>).data !== undefined;
+
 type MaybeCacheEntry = CacheEntry | null;
 export type CacheData = ArrayBuffer | Chunk<DataType>;
 type CacheEntry = {
--- a/src/VolumeCache.ts
+++ b/src/VolumeCache.ts
@@ -1,7 +1,10 @@
+import { Chunk, DataType } from "@zarrita/core";
+
 type MaybeCacheEntry = CacheEntry | null;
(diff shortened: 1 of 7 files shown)


toloudis:
If you still have the screen grabs, could you put the profile run results in here for posterity?

toloudis:
I think getting rid of WrappedStore is a great idea if you're saying we don't need it.. in a subsequent PR

PR #295, opened 2025-03-04
Update zarrita to v0.4

frasercl:
Review time: small (~5min)

Updates to zarrita's [new v0.4 release](https://github.com/manzt/zarrita.js/releases/tag/zarrita%400.4.0). This release deletes a number of subpackages from which we were explicitly importing, while seemingly resolving the errors that motivated us to import from the subpackages in the first place. So a lot of imports have changed, but not much else is different here.

This release also seems to fix the remaining errors which were preventing us from loading Zarr v3 (and therefore OME-Zarr v0.5) datasets, and therefore resolves #263. To verify:
- in this repo: `npm i`; `npm run build`; `npm link`
- in vole-app: `npm link @aics/vole-core`; `npm start`
- open http://localhost:9020/viewer?url=https://uk1s3.embassy.ebi.ac.uk/idr/share/ome2024-ngff-challenge/0.0.5/6001240.zarr

Issue #263, opened 2024-11-25
load OME-Zarr v0.5 (Zarr V3)


toloudis:
## Use Case

OME-Zarr is moving forward to adopt Zarr V3.  Zarr 3 has a huge key benefit with sharding that has big implications on storage and fetching chunks.

## Solution

OMEZarrLoader can load OME-Zarr v0.5 in addition to v0.4.  

See baby steps in #260 - basically we need our caching WrappedStore to support sharded range requests.
Can extend that branch or start anew.

PR #296, opened 2025-03-04
Prefetch/queueing/caching cleanup

frasercl:
Review time: small-medium (~15-20min)

# Problem

Our current prefetch and request queueing code does not play nice with sharded Zarr v3 volumes, and a previous PR introduced some additional problems.

# Solution

#291 moved chunk caching from the `WrappedStore` object to a new array wrapper object. That PR's description includes more context around why we'd want to do that &mdash; the short version is that instead of dealing with string-keyed data storage locations coming in and raw bytes going out, we can deal with chunk coordinates in and unencoded chunk data out. This is particularly relevant to Zarr v3, where thanks to the optional "sharding" feature multiple chunks can reside within a single storage key. So currently, the prefetch system receives the coordinates of shards rather than chunks, producing incoherent behavior; and the request queue may try to re-queue multiple requests for different chunks which reside in the same shard, and therefore have the same path. This PR moves these features into the new array wrapper, and removes the `WrappedStore` object.

I also spotted that chunks were not being keyed with their full URL in the cache, so that's fixed here too.

Finally, since this PR takes major steps towards loading Zarr v3, I spent some time testing Zarr v3 volumes and fixed a parsing error with v3 metadata. So this PR gets us almost all the way to resolving #263; the remaining necessary changes are coming in #295.


src/loaders/zarr_utils/wrapArray.ts
frasercl, line 1: Renamed from CachingArray.ts, and beefed up with most of the code from the now-deleted WrappedStore.ts.

src/loaders/zarr_utils/types.ts
```diff
@@ -92,7 +92,13 @@ export type OMEZarrMetadata = {
   omero: OmeroTransitionalMetadata;
 };
 
-export type NumericZarrArray = zarr.Array<zarr.NumberDataType, WrappedStore<RequestInit>>;
+export type CachingArrayOpts = {
```
toloudis, line 95: change name to WrapArrayOpts maybe

src/loaders/zarr_utils/wrapArray.ts
```diff
@@ -0,0 +1,67 @@
...
+  array: ZarrArray<T, Store>,
+  basePath: string,
+  cache?: VolumeCache,
+  queue?: SubscribableRequestQueue
+): ZarrArray<T, AsyncReadableExt<Opts>> {
+  const path = basePath.endsWith("/") ? basePath.slice(0, -1) : basePath;
+  const keyBase = path + array.path + (array.path.endsWith("/") ? "" : "/");
+
+  const getChunk = async (coords: number[], opts?: Parameters<AsyncReadableExt<Opts>["get"]>[1]): Promise<Chunk<T>> => {
+    if (pathIsToMetadata(array.path)) {
```
toloudis, line 25: seems like you could totally remove this

toloudis (approved):

src/loaders/OmeZarrLoader.ts
```diff
@@ -170,29 +167,32 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
-      const store = new WrappedStore<RequestInit>(new FetchStore(url), queue);
+      const store = new FetchStore(url);
       const root = zarr.root(store);
 
       const group = await zarr
         .open(root, { kind: "group" })
         .catch(wrapVolumeLoadError(`Failed to open OME-Zarr data at ${url}`, VolumeLoadErrorType.NOT_FOUND));
 
+      // OME-Zarr v0.5+ metadata has a similar shape to v0.4 (for our purposes), but is stored under the `ome` key
+      const meta = (group.attrs as { ome?: unknown }).ome ?? group.attrs;
```
toloudis, line 178: I suggest putting this in a function that returns a OMEZarrMetadata as a starting point to completely separating/encapsulating the metadata parsing.

src/loaders/OmeZarrLoader.ts
```diff
@@ -431,20 +431,13 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
 
-  private async prefetchChunk(
-    scaleLevel: NumericZarrArray,
-    coords: TCZYX<number>,
-    subscriber: SubscriberId
-  ): Promise<void> {
-    const { store, path } = scaleLevel;
-    const separator = path.endsWith("/") ? "" : "/";
-    const key = path + separator + this.orderByDimension(coords).join("/");
+  private prefetchChunk(scaleLevel: NumericZarrArray, coords: TCZYX<number>, subscriber: SubscriberId) {
```
ShrimpCryptid, line 434: Missing return type? Surprised TS didn't warn on this

toloudis (approved):

PR #301, opened 2025-03-28
feat: Added support for synchronous operations (render + setTime)

ShrimpCryptid:
# Problem

TFE is now using `vole-core`, but it requires some operations to be synchronous.

During an export operation, TFE will asynchronously request a frame load for each frame in a video. Once the frame load Promise resolves, TFE's exporter expects that it can synchronously call `render()` on the canvas, and then immediately get the canvas image and save it to the video buffer. 

Currently, the only access to rendering is via the `redraw()` function on `View3d`, which hooks into the asynchronous `requestAnimationFrame` loop. Additionally, `setTime` is not asynchronous either, which is necessary for both time playback and for Export support.

*Estimated review size: tiny, <5 minutes*

# Solution

- Changed `setTime` to return a `Promise` that resolves once a frame is loaded.
- Added access to synchronous rendering with `renderer.render()`.
- Added a new `getCanvasDOMElement` method for directly getting the `canvas` that the renderer is rendering to. 

## Type of change

### DIFF
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -146,6 +146,10 @@ export class View3d {
     return this.canvas3d.containerdiv;
   }
 
+  getCanvasDOMElement(): HTMLCanvasElement {
+    return this.canvas3d.renderer.domElement;
+  }
+
   getCameraState(): CameraState {
     return this.canvas3d.getCameraState();
   }
@@ -160,6 +164,10 @@ export class View3d {
    */
   redraw(): void {
     this.canvas3d.redraw();
+  }
+
+  render(): void {
+    this.canvas3d.render();
   }
 
   unsetImage(): VolumeDrawable | undefined {
@@ -259,10 +267,11 @@ export class View3d {
     this.loadErrorHandler = handler;
   }
 
-  setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): void {
+  async setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): Promise<void> {
     const timeClamped = Math.max(0, Math.min(time, volume.imageInfo.times - 1));
-    volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
+    const loadPromise = volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
     this.updateTimestepIndicator(volume);
+    await loadPromise;
   }
 
   /**
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -270,7 +270,7 @@ export class View3d {
     this.loadErrorHandler = handler;
   }
 
-  async setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): Promise<void> {
+  setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): Promise<void> {
     const timeClamped = Math.max(0, Math.min(time, volume.imageInfo.times - 1));
     const loadPromise = volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
     this.updateTimestepIndicator(volume);
--- a/src/View3d.ts
+++ b/src/View3d.ts
@@ -161,13 +161,16 @@ export class View3d {
 
   /**
    * Force a redraw.
+   * @param synchronous If true, the redraw will be done synchronously. If false (default), the
+   * redraw will be done asynchronously via `requestAnimationFrame`. Redraws should be done async
+   * whenever possible for the best performance.
    */
-  redraw(): void {
-    this.canvas3d.redraw();
-  }
-
-  render(): void {
-    this.canvas3d.render();
+  redraw(synchronous: boolean = false): void {
+    if (synchronous) {
+      this.canvas3d.onAnimationLoop();
+    } else {
+      this.canvas3d.redraw();
+    }
   }
 
   unsetImage(): VolumeDrawable | undefined {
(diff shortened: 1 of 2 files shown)


src/Volume.ts
```diff
@@ -176,7 +176,7 @@ export default class Volume {
     }
 
     if (shouldReload) {
-      this.loadNewData(onChannelLoaded);
+      await this.loadNewData(onChannelLoaded);
```
ShrimpCryptid, line 179: I checked both `vole-core` and `vole-app`, and currently nothing is calling `await updateRequiredData` or is doing anything with the `Promise` it returns. I also consulted with @frasercl and we agreed that doing this seems okay.

src/View3d.ts
```diff
@@ -259,10 +267,11 @@ export class View3d {
     this.loadErrorHandler = handler;
   }
 
-  setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): void {
+  async setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): Promise<void> {
     const timeClamped = Math.max(0, Math.min(time, volume.imageInfo.times - 1));
-    volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
+    const loadPromise = volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
     this.updateTimestepIndicator(volume);
+    await loadPromise;
```
toloudis, line 274: does this mean that setTime will always wait, whereas before it wasn't?

src/View3d.ts
```diff
@@ -162,6 +166,10 @@ export class View3d {
     this.canvas3d.redraw();
   }
 
+  render(): void {
+    this.canvas3d.render();
+  }
```
toloudis, line 171: slightly hesitant to give clients this hook.  
Also, if you look at `redraw()`, the actual function responsible for a full redraw is called `onAnimationLoop`.  Make sure that you don't actually need the extra bits of code that happen in there.  

(Nit) I would maybe consider just passing a flag to the redraw call instead of adding this new function...

src/Volume.ts
toloudis: we absolutely can not harm performance during time series playback in VolE app

src/View3d.ts
ShrimpCryptid: It just means that it returns an optional `Promise` that a client can decide whether to use or `await` on. Once `setTime` hits the `await`, the calling code will resume as normal after the call.

I can `return loadPromise` here to make this behavior more explicit.

src/View3d.ts
ShrimpCryptid: That sounds reasonable to me! I'll make that change.

src/View3d.ts
```diff
@@ -259,10 +270,11 @@ export class View3d {
     this.loadErrorHandler = handler;
   }
 
-  setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): void {
+  async setTime(volume: Volume, time: number, onChannelLoaded?: PerChannelCallback): Promise<void> {
     const timeClamped = Math.max(0, Math.min(time, volume.imageInfo.times - 1));
-    volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
+    const loadPromise = volume.updateRequiredData({ time: timeClamped }, onChannelLoaded);
     this.updateTimestepIndicator(volume);
+    return loadPromise;
```
ShrimpCryptid, line 277: Hmm, okay, so, thinking about this more.

@toloudis I think I could get away with not making `setTime` async? I'd have to do some promise juggling at the TFE level but it's possible if we really don't like this.

ShrimpCryptid: Basically I'd have to create an empty Promise in TFE and save it to the ColorizeCanvas3D state. I'd wait until a callback I pass to `onChannelLoaded` registers that the volume has loaded, and if so I'd resolve that empty Promise.

toloudis: i think in general it's fine if setTime is async.

src/View3d.ts
```diff
@@ -146,6 +146,10 @@ export class View3d {
     return this.canvas3d.containerdiv;
   }
 
+  getCanvasDOMElement(): HTMLCanvasElement {
+    return this.canvas3d.renderer.domElement;
+  }
```
toloudis, line 151: Thinking about this and anticipating future things like exposing the Camera, maybe it would be better if view3d can just export the ThreeJsCanvas, and we tighten up the public interface on that...  to be considered in a future pr.

src/View3d.ts
```diff
@@ -157,9 +161,16 @@ export class View3d {
 
   /**
    * Force a redraw.
+   * @param synchronous If true, the redraw will be done synchronously. If false (default), the
+   * redraw will be done asynchronously via `requestAnimationFrame`. Redraws should be done async
+   * whenever possible for the best performance.
    */
-  redraw(): void {
-    this.canvas3d.redraw();
+  redraw(synchronous: boolean = false): void {
```
toloudis, line 168: linter warning here

toloudis (approved):

src/View3d.ts
frasercl: Came up when we discussed this, but the (minor) worry here is if other state setters that can update what data we need (z slice or channels, for instance) return `Promise`s, then clients start depending on those `Promise`s, then multiple settings like that get set in order and earlier `Promise`s start doing unexpected things. Should be safe as long as it's isolated to just this one setter though, even if it breaks the paradigm a tiny bit.

frasercl (approved):

PR #306, opened 2025-04-08
feat: Added support for per-frame ID offsets during colorizing

ShrimpCryptid:
# Problem

Adds support for https://github.com/allen-cell-animated/timelapse-colorizer/issues/551, "Support frame-local IDs (instead of global IDs only)".

TFE uses big arrays to hold feature, outlier, and other data. To look up data from these arrays, a segmentation has a globally-unique ID which it uses as the index. 

Previously, we also saved globally-unique IDs in our image data. However, ZARR segmentation files typically do NOT have globally-unique IDs, and are formatted so ID numbers start at 1 on each frame. To add compatibility for ZARR data, we have to offset from the raw IDs on each frame to get their global IDs.

*Estimated review size: small, 10-15 minutes*

# Solution

- Added the `timeToIdOffset` lookup in the `ColorizeFeature` colorizing params.
- FuseChannel now tracks the current loaded time of a volume, and uses it to passed an ID offset uniform to the fuse shader, 

## Type of change

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/68325096-f4af-490d-8af7-c461445dafa2

## Keyfiles
- `types.ts`

### DIFF
--- a/src/FusedChannelData.ts
+++ b/src/FusedChannelData.ts
@@ -237,22 +237,23 @@ export default class FusedChannelData {
         const mat = this.getShader(channels[chIndex].dtype, isColorize).clone();
         mat.uniforms.srcTexture.value = channels[chIndex].dataTexture;
         mat.uniforms.highlightedId.value = combination[i].selectedID == undefined ? -1 : combination[i].selectedID;
-        if (isColorize) {
-          mat.uniforms.featureData.value = combination[i].feature?.idsToFeatureValue;
-          mat.uniforms.outlierData.value = combination[i].feature?.outlierData;
-          mat.uniforms.inRangeIds.value = combination[i].feature?.inRangeIds;
-          mat.uniforms.featureColorRampMin.value = combination[i].feature?.featureMin;
-          mat.uniforms.featureColorRampMax.value = combination[i].feature?.featureMax;
-          mat.uniforms.colorRamp.value = combination[i].feature?.featureValueToColor;
-          mat.uniforms.outlineColor.value = combination[i].feature?.outlineColor;
-          mat.uniforms.outlierColor.value = combination[i].feature?.outlierColor;
-          mat.uniforms.outOfRangeColor.value = combination[i].feature?.outOfRangeColor;
-          mat.uniforms.outlierDrawMode.value = combination[i].feature?.outlierDrawMode;
-          mat.uniforms.outOfRangeDrawMode.value = combination[i].feature?.outOfRangeDrawMode;
-          mat.uniforms.hideOutOfRange.value = combination[i].feature?.hideOutOfRange;
+        const feature = combination[i].feature;
+        if (isColorize && feature) {
+          mat.uniforms.featureData.value = feature.idsToFeatureValue;
+          mat.uniforms.outlierData.value = feature.outlierData;
+          mat.uniforms.inRangeIds.value = feature.inRangeIds;
+          mat.uniforms.featureColorRampMin.value = feature.featureMin;
+          mat.uniforms.featureColorRampMax.value = feature.featureMax;
+          mat.uniforms.colorRamp.value = feature.featureValueToColor;
+          mat.uniforms.outlineColor.value = feature.outlineColor;
+          mat.uniforms.outlierColor.value = feature.outlierColor;
+          mat.uniforms.outOfRangeColor.value = feature.outOfRangeColor;
+          mat.uniforms.outlierDrawMode.value = feature.outlierDrawMode;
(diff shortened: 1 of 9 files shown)


ShrimpCryptid:
NOTE: This currently does NOT affect the pick buffer, which means that calls to `hitTest` return raw per-frame IDs instead of global IDs. Is this okay?

ShrimpCryptid:
Converted this to draft briefly because I was considering whether to implement https://github.com/allen-cell-animated/timelapse-colorizer/issues/630, since it would affect a lot of the same logic.

I think it makes sense to keep get this PR reviewed now, since the time-related changes will stay the same.

src/FusedChannelData.ts
```diff
@@ -249,6 +250,10 @@ export default class FusedChannelData {
           mat.uniforms.outlierDrawMode.value = combination[i].feature?.outlierDrawMode;
           mat.uniforms.outOfRangeDrawMode.value = combination[i].feature?.outOfRangeDrawMode;
           mat.uniforms.hideOutOfRange.value = combination[i].feature?.hideOutOfRange;
+          // Offset IDs based on the current frame, for data without
+          // globally-unique IDs.
+          const idOffset = combination[i].feature?.timeToIdOffset[combination[i].time ?? 0] ?? 0;
+          mat.uniforms.idOffset.value = idOffset;
```
frasercl, line 256: might be worth assigning `combination[i].feature` to a temporary variable above to avoid some boilerplate

frasercl (approved):

src/constants/shaders/colorizeUI.frag
```diff
@@ -45,6 +47,15 @@ uvec4 getUintFromTex(usampler2D tex, int index) {
...
 }
+
+uint getId(ivec2 uv) {
+  uint rawId = texelFetch(srcTexture, uv, 0).r;
+  if (rawId == 0u) {
+    return 0u;
+  }
+  return rawId + idOffset;
+}
+
```
toloudis, line 58: nit: pr would be cleaner if this function were still up above where it used to be

src/types.ts
```diff
@@ -56,6 +56,16 @@ export const ARRAY_CONSTRUCTORS = {
...
   featureValueToColor: DataTexture;
+  /**
+   * A mapping from the current frame number to the ID offset for that frame.
+   * This is used for data where the raw IDs are not globally-unique. Adding the
+   * offset to the raw ID gives the unique, global ID that can be used to index
+   * into the `idsToFeatureValue` and other colorize-related data textures.
+   *
+   * For each raw ID `i` at some time `t`, the global ID is `i +
+   * timeToIdOffset[t]`.
+   */
```
toloudis, line 67: so in the case of globally unique ids, we just have to use a zero offset?  it might be worth saying that you have to do it explicitly by passing zeros in...

src/VolumeDrawable.ts
```diff
@@ -433,8 +434,7 @@ export default class VolumeDrawable {
         this.pickRendering = new PickVolume(this.volume, this.settings);
       }
       this.pickRendering.setChannelToPick(channelIndex);
-    }
-    else {
+    } else {
```
toloudis, line 437: did autoformat do this?

src/loaders/IVolumeLoader.ts
```diff
@@ -179,10 +179,11 @@ export abstract class ThreadableVolumeLoader implements IVolumeLoader {
         const dtype = dtypes[i];
         const data = dataArrays[i];
         const range = ranges[i];
+        const time = volume.loadSpec.time;
```
toloudis, line 182: in VolumeDrawable you used loadSpecRequired.  make sure that's consistent and you're using the right one everywhere.  (I can't remember which is which, exactly)

toloudis (approved):
LGTM!  please just double check on the difference between loadSpec.time and loadSpecRequired.time

src/loaders/IVolumeLoader.ts
ShrimpCryptid: This was a really good question! I think from reviewing `Volume`, `loadSpecRequired` represents data that is required and _will_ be loaded, while `loadSpec` is data that _is_ loaded (or at least is currently in the process of being loaded).

Either will probably work but I think `loadSpec` is the right choice.

src/VolumeDrawable.ts
ShrimpCryptid: Yes!

src/loaders/IVolumeLoader.ts
ShrimpCryptid: Oops actually the changes in VolumeDrawable are a leftover from when time was in FuseChannel! Removing :>

PR #313, opened 2025-04-19
feat: Added support for lookups from segmentation IDs to global IDs for sparse data

ShrimpCryptid:
# Problem

Supports https://github.com/allen-cell-animated/timelapse-colorizer/issues/630, "display data with arbitrary segmentation IDs."

This change removes a lot of the assumptions we were making about segmentation IDs in datasets. **The linked issue above has a better explanation of what this looks like.**

With the corresponding PR in TFE, this removes a requirement that all datasets be sorted by time and segmentation ID, and that segmentation IDs have to be monotonically increasing. It fixes a number of bugs in the preview EMT dataset as well that were caused by missing segmentation IDs.

*Estimated review size: small, <10 minutes*

# Solution

- Replaced the `frameToIdOffset` in `ColorizeFeature` with a new parameter, `frameToGlobalIdLookup`.
  - `frameToGlobalIdLookup` is a Map from frame numbers to an object containing a texture and an offset.
- Changed the shader to look up the global ID for each segmentation ID using the new uniforms.
- Fixed a bug where the highlighted ID was off-by-one.
- Added handling for segmentation IDs that have no corresponding global ID/no feature data.

## Type of change

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/ce7c3828-1c1f-4ceb-90ac-8c575b1ee145

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -1,4 +1,4 @@
-import { Color, Vector3 } from "three";
+import { Color, DataTexture, Vector3 } from "three";
 import GUI from "lil-gui";
 
 import { colormaps as colorizercolormaps, features as colorizerfeatures } from "./colorizer";
@@ -1182,7 +1182,7 @@ function getStateColorizeFeature(): ColorizeFeature | null {
       outlierDrawMode: 0,
       outOfRangeDrawMode: 0,
       hideOutOfRange: false,
-      timeToIdOffset: new Uint32Array([0]),
+      frameToGlobalIdLookup: new Map(),
     };
   } else {
     return null;
--- a/public/index.ts
+++ b/public/index.ts
@@ -1,4 +1,4 @@
-import { Color, DataTexture, Vector3 } from "three";
+import { Color, Vector3 } from "three";
 import GUI from "lil-gui";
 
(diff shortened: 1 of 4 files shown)


src/constants/shaders/colorizeUI.frag
```diff
@@ -90,12 +101,12 @@ vec4 getObjectColor(ivec2 sUv, float opacity) {
...
 
   // A segmentation id of 0 represents background
-  if (id == 0u) {
+  if (id == BACKGROUND_ID) {
     return vec4(0, 0, 0, 0);
   }
 
   // color the highlighted object
-  if (id == highlightedId) {
+  if (id - 1u == highlightedId) {
```
ShrimpCryptid, line 109: Bug here from before where we weren't offsetting the ID by 1

rugeli (approved):

public/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Color, Vector3 } from "three";
```
frasercl, line 1: unused import

frasercl (approved):
Looks good!

PR #321, opened 2025-06-13
feat: Add 3D line objects to volume

### LINKED ISSUE #285: TFE 3D - Show track path of selected cells
## Use Case

Replicate existing TFE behavior by adding a track path to the selected track.

## Solution

- [x] Allow external consumers to pass in a list of 3D centroids and a range over which to draw them.
- [x] Render the modified ray march pass to a depth buffer (#282). 
- [x] Render a line object (either using the Three Line object) into the scene, with visible segments determined by the provided range parameter.
- [x] Lines should be occluded/occlude segmentations correctly based on Z-depth.

## Alternatives

ShrimpCryptid:
# Problem

Closes #285, "TFE 3D - Show track path of selected cells." 

This change gives View3d an API for adding and removing 3D lines. There is a corresponding PR in TFE that visualizes the track path as well!

*Estimated review size: medium, 30 minutes*

# Solution

- Adds the new `Line3d` class, which is a wrapper around the `LineSegments2` object in Three.js.
- Added methods for adding, removing, and checking for the existence of line objects.

Misc:
- Fixes a bug where glasbey color ramps were not being rendered correctly.

## Type of change

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/721efcc8-beb5-4284-9e04-a448eb3de3ad

### DIFF
--- a/src/Line3d.ts
+++ b/src/Line3d.ts
@@ -126,10 +126,10 @@ export default class Line3d implements IDrawableObject {
    * @param renderAsOverlay If true, the line will be rendered on top of the
    * volume, ignoring depth.
    */
-  setOverlay(overlay: boolean): void {
-    this.lineMesh.layers.set(overlay ? OVERLAY_LAYER : MESH_LAYER);
-    this.lineMesh.material.depthTest = !overlay;
-    this.lineMesh.material.depthTest = !overlay;
+  setRenderAsOverlay(renderAsOverlay: boolean): void {
+    this.lineMesh.layers.set(renderAsOverlay ? OVERLAY_LAYER : MESH_LAYER);
+    this.lineMesh.material.depthTest = !renderAsOverlay;
+    this.lineMesh.material.depthTest = !renderAsOverlay;
     this.lineMesh.material.needsUpdate = true;
   }
 


src/VolumeDrawable.ts
```diff
@@ -823,6 +848,27 @@ export default class VolumeDrawable {
     this.updateScale();
   }
 
+  addSceneObject(object: IDrawableObject): void {
+    if (!this.sceneObjects.has(object)) {
+      this.sceneObjects.add(object);
+      this.sceneObjectsGroup.add(object.get3dObject());
+      object.setResolution(this.settings.resolution.x, this.settings.resolution.y);
+      object.setFlipAxes(this.settings.flipAxes.x, this.settings.flipAxes.y, this.settings.flipAxes.z);
```
ShrimpCryptid: Dan: Should this be set just at the scene object root level? Are there scene objects that DO NOT want to be flipped with the rest of the volume?

ShrimpCryptid: Also `sceneObjects` are positioned + scaled relative to the volume. Should this be reflected in the API name + default coordinate space

ShrimpCryptid: - Cameron: Does VolumeDrawable make sense as the scene root? Should other elements (bounding box, etc.) be split out from VolumeDrawable, and VolumeDrawable itself becomes a DrawableObject?
- Dan: Right now, everything we draw is a child of the volume. **"addVolumeChildObject"** is a closer name to what we're doing -> but maybe that shouldn't be the API we implement.
  - Opportunity to make a real, general API for arbitrary objects (maybe also render instancing...)
  - `addLinePathObject` as a simple API

src/types.ts
```diff
@@ -88,6 +88,33 @@ export interface ColorizeFeature {
...
+  doRender(): void;
+  get3dObject(): Group;
+  setTranslation(translation: Vector3): void;
+  setScale(scale: Vector3): void;
+  setRotation(eulerXYZ: Euler): void;
+  setFlipAxes(flipX: number, flipY: number, flipZ: number): void;
+  setOrthoThickness(thickness: number): void;
+  setResolution(x: number, y: number): void;
+  setAxisClip(axis: "x" | "y" | "z", minval: number, maxval: number, _isOrthoAxis: boolean): void;
+}
```
ShrimpCryptid: Keep even if it's not used yet

src/types.ts
```diff
@@ -88,6 +88,33 @@ export interface ColorizeFeature {
...
+  FRONT = 1,
+}
+
+export interface LineConfiguration {
+  /** Width in onscreen pixels. */
+  widthPx: number;
+  color: string | Color;
+  // colors?: Float32Array;
+  renderOrder: LineRenderOrder;
+}
```
ShrimpCryptid: Remove

src/VolumeDrawable.ts
```diff
@@ -823,6 +848,27 @@ export default class VolumeDrawable {
...
+
+  hasSceneObject(object: IDrawableObject): boolean {
+    return this.sceneObjects.has(object);
+  }
+
+  removeSceneObject(object: IDrawableObject): void {
+    if (this.sceneObjects.has(object)) {
+      this.sceneObjects.delete(object);
+      this.sceneObjectsGroup.remove(object.get3dObject());
+      // object.cleanup();
```
ShrimpCryptid: Remove

src/VolumeDrawable.ts
ShrimpCryptid: - [x] **Rename to `addLinePath()`**

src/FusedChannelData.ts
```diff
@@ -144,6 +144,7 @@ export default class FusedChannelData {
         colorRamp: {
           value: null,
         },
+        useRepeatingCategoricalColors: { value: false },
```
ShrimpCryptid, line 147: This is functionality that was recently added to TFE 2D to support glasbey color ramps, which use a repeating color ramp with 256 colors. I've made the same changes here.

src/Line3d.ts
```diff
@@ -0,0 +1,185 @@
...
+import { MESH_LAYER, OVERLAY_LAYER } from "./ThreeJsPanel";
+import { LineSegments2 } from "three/addons/lines/LineSegments2";
+import { LineSegmentsGeometry } from "three/addons/lines/LineSegmentsGeometry";
+
+const DEFAULT_VERTEX_BUFFER_SIZE = 1020;
+
+/**
+ * Simple wrapper for a 3D line segments object, with controls for vertex data,
+ * color, width, and segments visible.
+ */
```
ShrimpCryptid, line 13: @frasercl I'm debating a little bit whether I should have TFE deal with all of this, and literally be passing in a LineSegments2 object? But I imagine the extra layer of abstraction is good?

src/VolumeDrawable.ts
```diff
@@ -48,6 +56,15 @@ export default class VolumeDrawable {
...
   private meshVolume: MeshVolume;
+  /**
+   * Group for all child objects of the volume. The group is scaled and
+   * transformed to the normalized volume coordinate space, where the volume's
+   * center is at (0, 0, 0) and the bounds have a range of [-0.5, 0.5] along each
+   * axis.
+   */
+  private childObjectsGroup: Group;
+  /** Set of drawable objects that are children of the volume's transform. */
+  private childObjects: Set<IDrawableObject>;
```
ShrimpCryptid, line 71: The `meshVolume` is now a child of the `childObjectsGroup`, and all the transforms that were previously applied to it specifically are now applied to the entire group.

tsconfig.json
```diff
@@ -9,6 +9,9 @@
     "module": "ES2020",
     "moduleResolution": "node",
     "noImplicitAny": false,
+    "paths": {
+      "three/addons/lines/*": ["./node_modules/@types/three/examples/jsm/lines/*"]
```
ShrimpCryptid, line 13: Gives us types for `LineSegments2`, which has variable line width.

interim17 (approved):
I don't see any issues jumping out when I read this. You tagged @frasercl in a question about whether the extra layer of abstraction is worth it or if you should just use the LineSegments2 that three provides... I would default to _not_ adding abstraction, but haven't really dug into the use case or what LineSegments2 provides enough to say whether the extra code is worth it. Trusting your judgement.

src/Line3d.ts
frasercl: I like the extra abstraction! Already commented basically this on the TFE side, but I think in principle our external API should depend as little on three.js types as possible.

src/VolumeDrawable.ts
frasercl: Certainly doesn't have to happen in this PR, but I think the long-term future of this pattern ought to be that this "container" responsibility gets split off into its own class (say `VolumeGroup`) which owns drawable objects like lines _and also_ `VolumeDrawable` and the drawable objects it currently owns (`MeshVolume`, the bounding box, etc.). Right now, `MeshVolume` is sort of half-in, half-out &mdash; it's a child of this group for three.js purposes but a child of `VolumeDrawable` directly for ours. That's just fine for now, but maybe worth a callout in comments to make that clear?

src/Line3d.ts
```diff
@@ -0,0 +1,188 @@
...
+    this.lineMesh.material.needsUpdate = true;
+  }
+
+  /**
+   * Sets whether a line should be rendered as an overlay (rendered on top of
+   * the volume) instead of a mesh (with depth, intersects with volume).
+   * @param renderAsOverlay If true, the line will be rendered on top of the
+   * volume, ignoring depth.
+   */
+  setOverlay(overlay: boolean): void {
```
frasercl: `setIsOverlay`? `setRenderAsOverlay`? Currently this kind of implies that an overlay is a different object you can apply to a line. Though I got it fast with docs so current name is ok to stick with too

src/Line3d.ts
ShrimpCryptid: Perfect, thanks for the input!

PR #324, opened 2025-07-01
Remove `View3d.loaderContext` property

frasercl:
Review time: xsmall (~5min)

# Problem

`View3d` currently has a stray public property `loaderContext`, to give `View3d` a reference to a `VolumeLoaderContext` object which manages a worker thread and creates volume loaders on it. This seems to stem from an era when we thought `VolumeLoaderContext` might eventually become subsumed into `View3d`. But as we've used the loader API more for multi-scene collections, things haven't really shaken out that way. Instead, given the current flow (create a loader, create an image from the loader, put the image in the viewer), it seems to make more sense that the context be owned by the client which wants to create loaders.

That property currently is only used to bind a control in the secret Tweakpane menu. If we can create another way to get at the currently in-use `VolumeLoaderContext`, we can sand off an odd edge of the `View3d` API.

# Solution

Allow `VolumeLoaders` to store references to their context and to give those references via a thread.

This branch also picked up some unrelated polish changes in `TiffLoader`, which I made to make more values well-typed and which should be relatively uncontroversial.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,6 @@ const loaderContext = new VolumeLoaderContext(CACHE_MAX_SIZE, CONCURRENCY_LIMIT,
 
 // create the viewer.  it will try to fill the parent element.
 const view3D = new View3d(el);
-view3D.loaderContext = loaderContext;
 
 // ensure the loader worker is ready
 await loaderContext.onOpen();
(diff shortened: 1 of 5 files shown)


ShrimpCryptid (approved):

ascibisz (approved):

toloudis:
@frasercl note that the README has an example that shows the loaderContext being set

PR #325, opened 2025-07-03
fix: Fixed bug where volume rendering would crash when the canvas is resized

ShrimpCryptid:
# Problem

Fixes a bug described in https://github.com/allen-cell-animated/timelapse-colorizer/issues/641 where the screen goes black during resize when the pick buffer is enabled.

The underlying cause was that, after a resize call to View3d, the depth texture for the `meshRenderTarget` gets disposed of and is not recreated until the next render. That meant that the disposed/deallocated texture was passed to `fillPickBuffer()` early, which caused a crash.

This change moves `fillPickBuffer` to run after the main render occurs, and adds a comment about why it needs to do so.

*Estimated review size: tiny, 2 minutes*

## Screenshots (optional):

Before:
image

After:

image

### DIFF
--- a/src/ThreeJsPanel.ts
+++ b/src/ThreeJsPanel.ts
@@ -51,6 +51,12 @@ export type CameraState = {
   orthoScale?: number;
 };
 
+type AnimateFunction = (
+  renderer: WebGLRenderer,
+  camera: PerspectiveCamera | OrthographicCamera,
+  depthTexture?: DepthTexture | null
+) => void;
+
 export class ThreeJsPanel {
   public containerdiv: HTMLDivElement;
   private canvas: HTMLCanvasElement;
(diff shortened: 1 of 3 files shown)


src/ThreeJsPanel.ts
```diff
@@ -184,6 +184,8 @@ export class ThreeJsPanel {
     if (parentElement) {
       this.renderer.setSize(parentElement.offsetWidth, parentElement.offsetHeight);
       this.meshRenderTarget.setSize(parentElement.offsetWidth, parentElement.offsetHeight);
+      this.meshRenderTarget.depthTexture.dispose();
+      this.meshRenderTarget.depthTexture = new DepthTexture(parentElement.offsetWidth, parentElement.offsetHeight);
     }
```
frasercl: It looks like this whole `if` statement might be redundant? The depth texture is initially created with dimensions from `this.canvas.(width/height)`, and if `parentElement` exists, those are set to `parentElement.offset(Width/Height)` earlier.

toloudis (approved):
LGTM! Threejs is a little tricky with how it manages the depthTexture and I think this code change is the right way to let threejs do what it needs to do

frasercl (approved):
Makes sense!

PR #326, opened 2025-07-07
Import fixes for `Line3d`

frasercl:
Review time: xsmall

Some imports in Line3d.ts are missing ".js" file extensions to be proper "fully specified" imports, which causes problems when linked into vole-app. I put the extensions in and things started working again.


rugeli (approved):
Nice fix! I'm just wondering how it didn't get `module not found` errors before.

toloudis (approved):
LGTM! thanks!

Issue #218, opened 2024-06-04
Using OME-TIFF files generated from FIJI


JohnGrime:
## Description

When trying to load OME-TIFF files generated from e.g. FIJI, I can't seem to get the VolumeViewer to display the data correctly in the dev server example.

I generated a sample 3-channel 8-bit image stack (200x200x200 pixels) with a single time point, and saved the resultant data using both the "standard" OME-TIFF exporter in FIJI and also using the Bio-Formats exporter plugin.

In both cases, Volume Viewer appears to read the file correctly (three channels show up in the UI drawers) but only the first channel is actually displayed (in purple). I can't seem to get the other channels to display.

I tried this by simply modifying the `index.ts` file to put a URL to my generated file(s) under the `omeTIFF` entry in the `TEST_DATA `map, while keeping the type unchanged (`VolumeFileFormat.TIFF`).

I also tried with a single-channel OME-TIFF file generated in FIJI, and although VolumeViewer appears to load the data and understand that it's a single-channel OME-TIFF file ... I don't see any visual output.

## Expected Behavior

I'd expect all the channel data in the 3- or 1-channel OME-TIFF files to display, as they do for e.g. the `AICS-12_881.ome.tif` file specified in the original test data map.

## Reproduction

Run FIJI and use `File -> New -> Hyperstack`. Specify 8-bit and e.g. 200x200x200 pixels with 3 channels and one "Frames" (i.e., timepoints). Ensure "Label imges" is selected. Then save the resultant data as OME-TIFF via `File -> Save as -> OME-TIFF` or `Plugins -> Bio-Formats -> Bio-formats Exporter`

## Environment

FIJI 1.54f
VolumeViewer 3.8.3 dev (via cloning GitHub repo and `npm dev run`)

toloudis:
one way to check the tiff metadata
exiftool -b -ImageDescription -w "_ImageDescription.xml" "C:\path\to\image.ome.tif"

PR #327, opened 2025-07-08
Load TIFFs without OME metadata

### LINKED ISSUE #328: Load Cell Painting Gallery TIFFs
## Use Case

Cell Painting Gallery would like to load a "normal TIFF" which does not include OME-TIFF metadata.

## Solution

If possible, add support for loading non-OME-TIFFs

frasercl:
Review time: small

# Problem

Currently, we only support TIFFs which conform to the OME-TIFF metadata standard. But for many simple TIFF images, the base TIFF standard provides just enough metadata for us to work with.

This is specifically motivated by an ask from the Broad Institute's Cell Painting Gallery, who have non-OME-TIFFs that they'd like to be able to load in Vol-E.

Resolves #328; may also address #218.

# Solution

- Handle missing or incomplete OME-TIFF metadata a bit more leniently.
- When OME-TIFF metadata is missing, rather than erroring out, fill in what we can from base TIFF tags.

## To verify:

- In this repo: `npm i`; `npm run build`; `npm link`
- In vole-app: `npm link @aics/vole-core`; `npm start`
- Open http://localhost:9020/viewer?url=https://cellpainting-gallery.s3.us-east-1.amazonaws.com/cpg0000-jump-pilot/source_4/images/2020_12_08_CPJUMP1_Bleaching/images/BR00116992E__2020-11-12T01_22_40-Measurement1/Images/r01c01f01p01-ch8sk5fk1fl1.tiff

With this repo on `main`, this should error; with it on this branch, it should work.


src/loaders/TiffLoader.ts
```diff
@@ -121,6 +124,7 @@ function getOMEDims(imageEl: Element): OMEDims {
 }
 
 const getBytesPerSample = (type: string): number => (type === "uint8" ? 1 : type === "uint16" ? 2 : 4);
+const getPixelType = (pxSize: number): string => (pxSize === 1 ? "uint8" : pxSize === 2 ? "uint16" : "uint32");
```
ShrimpCryptid, line 127: Probably out of scope, but should these be constants/enums?

src/loaders/TiffLoader.ts
```diff
@@ -135,21 +139,29 @@ class TiffLoader extends ThreadableVolumeLoader {
 
   private async loadOmeDims(): Promise<OMEDims> {
     if (!this.dims) {
-      const tiff = await fromUrl(this.url, { allowFullFile: true }).catch(
+      const tiff = await fromUrl(this.url, { allowFullFile: true }).catch<GeoTIFF>(
```
ShrimpCryptid: Didn't know that you could cast types in `catch` :0

src/loaders/TiffLoader.ts
```diff
@@ -135,21 +139,29 @@ class TiffLoader extends ThreadableVolumeLoader {
...
-      this.dims = getOMEDims(image0El);
+        .catch<GeoTIFFImage>(wrapVolumeLoadError("Failed to open TIFF image", VolumeLoadErrorType.NOT_FOUND));
+
+      const omeEl = getOME(image.getFileDirectory().ImageDescription);
+
+      if (omeEl !== undefined) {
+        const image0El = omeEl.getElementsByTagName("Image")[0];
+        this.dims = getOMEDims(image0El);
+      } else {
+        console.warn("Could not read OME-TIFF metadata from file. Doing our best with base TIFF metadata.");
```
ShrimpCryptid, line 158:
```suggestion
        console.warn("Could not read OME-TIFF metadata from file. Vol-E will attempt to load the TIFF using base TIFF metadata.");
```
Increase formality in debug statement?

ShrimpCryptid (approved):

src/loaders/TiffLoader.ts
```diff
@@ -135,21 +139,29 @@ class TiffLoader extends ThreadableVolumeLoader {
...
+        this.dims = new OMEDims();
+        this.dims.sizex = image.getWidth();
+        this.dims.sizey = image.getHeight();
+        this.dims.pixeltype = getPixelType(image.getBytesPerPixel());
+        this.dims.channelnames = ["Channel"];
```
toloudis: was there any metadata to be gleaned from the PerkinElmer xml in particular?  Info we could stuff in the metadata panel?  This would probably make the CellPainting folks happy.  We will have to ask them about that when they're here...

frasercl: As far as I could tell that XML at the end of the file was inaccessible from geotiff.js. It may live under a proprietary TIFF tag that geotiff.js doesn't know about, or for all I know it may just be dropped raw into the end of the file.

toloudis:
WIP adding multi-source multichannel capability.  First implementation runs but is wrong.. Fixes coming.

public/types.ts
```diff
@@ -5,7 +5,10 @@ import { IVolumeLoader } from "../src/loaders/IVolumeLoader";
 export interface TestDataSpec {
   type: VolumeFileFormat | "opencell" | "procedural";
   // TODO: replace array here with multi-scene handling at the loader level
-  url: string | string[];
+  // one string is single scene
+  // an array of strings is multiple scenes
+  // (currently for tiff only) an array of arrays is multiple scenes with multiple channel sources
```
ShrimpCryptid, line 10: oh boy

ShrimpCryptid (approved):

public/types.ts
```diff
@@ -5,7 +5,10 @@ import { IVolumeLoader } from "../src/loaders/IVolumeLoader";
...
+  url: string | string[] | string[][];
```
meganrm, line 11: nit, but could this be called something different to make it clear what it is? the `this.url[0]` below without knowing this looks like you're doing some string splitting or something at first glance

src/loaders/TiffLoader.ts
```diff
@@ -121,35 +124,49 @@ function getOMEDims(imageEl: Element): OMEDims {
 }
 
 const getBytesPerSample = (type: string): number => (type === "uint8" ? 1 : type === "uint16" ? 2 : 4);
+const getPixelType = (pxSize: number): string => (pxSize === 1 ? "uint8" : pxSize === 2 ? "uint16" : "uint32");
 
 // Despite the class `TiffLoader` extends, this loader is not threadable, since geotiff internally uses features that
 // aren't available on workers. It uses its own specialized workers anyways.
 class TiffLoader extends ThreadableVolumeLoader {
-  url: string;
+  url: string[];
```
meganrm, line 132: if in this case the url is always an array of urls can you call this `scenes` or `urls`? or is this trying to be consistent with the parent class?

meganrm (approved):
looks good, just some clarity comments

toloudis:
@frasercl I wonder why ImageDescription seems to come back undefined for this data. I didn't dig in yet, but just a quick breakpoint in the debugger seemed to show that.  Is that what you saw too?

frasercl:
> @frasercl I wonder why ImageDescription seems to come back undefined for this data. I didn't dig in yet, but just a quick breakpoint in the debugger seemed to show that. Is that what you saw too?

That's what I saw too. I dug into it a bit and my current theory is it just doesn't exist. I'm only just now getting up to speed on how TIFFs work for this issue, but here's what I've learned so far:

Metadata in TIFFs lives under tags (the "T" in TIFF). These tags are identified by arbitrary 16-bit numbers. The standard gives those tags text identifiers, but they never actually show up in the file. The source of geotiff.js has [a helpful file](https://github.com/geotiffjs/geotiff.js/blob/master/src/globals.js) full of these tag IDs and their identifiers, some standard and some application-specific. `ImageDescription` is one of those tags, and despite the fact that the geotiff.js API privileges it with its own getter and the OME-TIFF standard uses it for its metadata, a TIFF is perfectly well-formed without it. So my hypothesis is that all the XML at the end of the file lives under some nonstandard PerkinElmer tag, rather than `ImageDescription`.

toloudis:
I figured it was in some other type of tag.  One question is whether it's indeed a standard tag that would make sense for us to check, or something totally custom.
Also interesting to note, I believe every image plane (or IFD directory) in a TIFF is allowed to have an ImageDescription but it is convention to put whole image metadata in the first ImageDescription.

PR #329, opened 2025-07-09
feat: Show selected ID with contours

### LINKED ISSUE #284: TFE 3D - Show highlight around selected cell
## Use Case

Replicate existing TFE behavior by adding a highlight to a selected cell segmentation.

## Solution

- Add a method where an external consumer can pass in currently highlighted segmentation ID and an outline color.
- Replicate Simularium behavior-- use pick buffer to color edges of the selected object with the provided color.

## Alternatives

ShrimpCryptid:
# Problem

Closes #284, "TFE 3D - Show outline around selected cell." This change adds in object outlines, using the pick buffer!

*Estimated review size: large, 40 minutes*

# Solution

- Adds a new `ContourPass`, which uses the pick buffer to detect edges around a selected/highlighted object and renders it into the scene as a transparent render pass.
- Removed the old selection behavior (changing the fused volume value to the highlighted color).
- Adjusted ordering of mesh rendering so that line paths do not occlude the pick buffer. 
- Changed pick buffer to stop randomizing ray starting positions (thanks, Dan!)

## Type of change

## Steps to Verify:

1. Checkout branch locally
2. Run `npm run start`
3. Open browser and pan around the scene. You should see object contours appear on the hovered nuclei.

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/bc266885-2aa4-419f-a413-3a2efab7fd6d

### DIFF
--- a/src/ContourPass.ts
+++ b/src/ContourPass.ts
@@ -7,6 +7,7 @@ import {
   Texture,
   Uniform,
   UnsignedByteType,
+  UnsignedIntType,
   WebGLRenderer,
   WebGLRenderTarget,
 } from "three";
@@ -34,7 +35,7 @@ const defaultUniforms: ContourUniforms = {
   outlineColor: new Uniform(new Color(1, 0, 1)),
   outlineAlpha: new Uniform(1.0),
   useGlobalIdLookup: new Uniform(false),
-  segIdToGlobalId: new Uniform(new DataTexture(new Uint32Array([0]), 1, 1, RGBAFormat, UnsignedByteType)),
+  segIdToGlobalId: new Uniform(new DataTexture(new Uint32Array([0]), 1, 1, RGBAFormat, UnsignedIntType)),
   segIdOffset: new Uniform(0),
   devicePixelRatio: new Uniform(1.0),
 };
@@ -51,7 +52,8 @@ export default class ContourPass {
   }
 
   setHighlightedId(id: number) {
-    (this.pass.material.uniforms.highlightedId as IUniform<number>).value = id;
+    this.pass.material.uniforms.highlightedId.value = id;
+    console.log("Set highlighted ID to:", id);
   }
 
   syncGlobalIdLookup() {
@@ -61,7 +63,7 @@ export default class ContourPass {
     if (this.featureInfo) {
       const globalIdLookupInfo = this.featureInfo.frameToGlobalIdLookup.get(this.time);
       if (globalIdLookupInfo) {
-        console.log("Using global ID lookup");
+        console.log("Using global ID lookup with type ", globalIdLookupInfo.texture.type);
         uniforms.useGlobalIdLookup.value = true;
         uniforms.segIdToGlobalId.value = globalIdLookupInfo.texture;
         uniforms.segIdOffset.value = globalIdLookupInfo.minSegId;
--- a/src/ContourPass.ts
+++ b/src/ContourPass.ts
@@ -21,6 +21,7 @@ type ContourUniforms = {
   outlineThickness: IUniform<number>;
   outlineColor: IUniform<Color>;
   outlineAlpha: IUniform<number>;
+  useGlobalIdLookup: IUniform<boolean>;
   segIdToGlobalId: IUniform<Texture>;
   segIdOffset: IUniform<number>;
   devicePixelRatio: IUniform<number>;
@@ -32,6 +33,7 @@ const defaultUniforms: ContourUniforms = {
   outlineThickness: new Uniform(2.0),
   outlineColor: new Uniform(new Color(1, 0, 1)),
   outlineAlpha: new Uniform(1.0),
+  useGlobalIdLookup: new Uniform(false),
   segIdToGlobalId: new Uniform(new DataTexture(new Uint32Array([0]), 1, 1, RGBAFormat, UnsignedByteType)),
   segIdOffset: new Uniform(0),
   devicePixelRatio: new Uniform(1.0),
@@ -39,18 +41,54 @@ const defaultUniforms: ContourUniforms = {
 
 export default class ContourPass {
   private pass: RenderToBuffer;
+  private featureInfo: ColorizeFeature | null;
+  private time: number;
 
   constructor() {
     this.pass = new RenderToBuffer(contourFragShader, defaultUniforms, true);
+    this.featureInfo = null;
+    this.time = 0;
   }
 
   setHighlightedId(id: number) {
     (this.pass.material.uniforms.highlightedId as IUniform<number>).value = id;
   }
 
-  setColorizeFeature(featureInfo: ColorizeFeature): void {
-    // TODO: needs access to time to read out the correct segIdToGlobalId texture....
-    this.pass.material.uniforms.outlineColor.value = featureInfo.outlineColor;
+  syncGlobalIdLookup() {
+    const uniforms = this.pass.material.uniforms as ContourUniforms;
+    console.log("Has featureInfo:", !!this.featureInfo);
+    console.log("Time:", this.time);
+    if (this.featureInfo) {
+      const globalIdLookupInfo = this.featureInfo.frameToGlobalIdLookup.get(this.time);
+      if (globalIdLookupInfo) {
+        console.log("Using global ID lookup");
+        uniforms.useGlobalIdLookup.value = true;
+        uniforms.segIdToGlobalId.value = globalIdLookupInfo.texture;
+        uniforms.segIdOffset.value = globalIdLookupInfo.minSegId;
+      } else {
+        console.log("No global ID lookup for this time, using segmentation IDs directly");
+        uniforms.useGlobalIdLookup.value = false;
+      }
+    } else {
+      uniforms.useGlobalIdLookup.value = false;
+    }
+  }
+
+  setColorizeFeature(featureInfo: ColorizeFeature | null): void {
+    if (this.featureInfo !== featureInfo) {
+      this.featureInfo = featureInfo;
+      if (featureInfo) {
+        this.pass.material.uniforms.outlineColor.value = featureInfo.outlineColor;
+      }
+      this.syncGlobalIdLookup();
+    }
+  }
+
+  setTime(time: number) {
+    if (this.time !== time) {
+      this.time = time;
+      this.syncGlobalIdLookup();
+    }
   }
 
   render(renderer: WebGLRenderer, target: WebGLRenderTarget | null, pickBuffer: WebGLRenderTarget) {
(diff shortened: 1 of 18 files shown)


src/RenderToBuffer.ts
```diff
@@ -22,14 +22,15 @@ export default class RenderToBuffer {
...
-  constructor(fragmentSrc: string, uniforms: { [key: string]: IUniform }) {
+  constructor(fragmentSrc: string, uniforms: { [key: string]: IUniform }, transparent = false) {
     this.scene = new Scene();
     this.geometry = new PlaneGeometry(2, 2);
 
     this.material = new ShaderMaterial({
       vertexShader: renderToBufferVertShader,
       fragmentShader: fragmentSrc,
       uniforms,
+      transparent,
```
ShrimpCryptid: Allow some render passes to be transparent, so they can overwrite the existing scene

src/ContourPass.ts
```diff
@@ -0,0 +1,107 @@
...
+    const uniforms = this.pass.material.uniforms as ContourUniforms;
+    console.log("Has featureInfo:", !!this.featureInfo);
+    console.log("Time:", this.time);
+    if (this.featureInfo) {
+      const globalIdLookupInfo = this.featureInfo.frameToGlobalIdLookup.get(this.time);
+      if (globalIdLookupInfo) {
+        console.log("Using global ID lookup with type ", globalIdLookupInfo.texture.type);
+        uniforms.useGlobalIdLookup.value = true;
+        uniforms.segIdToGlobalId.value = globalIdLookupInfo.texture;
+        uniforms.segIdOffset.value = globalIdLookupInfo.minSegId;
```
ShrimpCryptid: Avoid using `seg` as a term in vole-core?

src/ContourPass.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  setColorizeFeature(featureInfo: ColorizeFeature | null): void {
+    if (this.featureInfo !== featureInfo) {
+      this.featureInfo = featureInfo;
+      if (featureInfo) {
+        this.pass.material.uniforms.outlineColor.value = featureInfo.outlineColor;
+      }
+      this.syncGlobalIdLookup();
+    }
+  }
```
ShrimpCryptid: Consider only adding setters for properties of `featureInfo` that are needed

src/Line3d.ts
```diff
@@ -26,13 +25,13 @@ export default class Line3d implements IDrawableObject {
     geometry.setPositions(new Float32Array(this.bufferSize));
     const material = new LineMaterial({ color: "#f00", linewidth: 2, worldUnits: false });
     this.lineMesh = new LineSegments2(geometry, material);
-    this.lineMesh.layers.set(MESH_LAYER);
+    this.lineMesh.layers.set(MESH_NO_PICK_OCCLUSION_LAYER);
```
ShrimpCryptid, line 35: I moved the line meshes from the mesh layer to a new layer. This is because:
1. Lines need to write depth information so they interact with the volume (so the lines appear to fade into the volume if they intersect)
2. BUT, lines shouldn't interact with the pick buffer, since it can cause unexpected outlining behavior as below:

Image

Now, the lines are rendered after the pick buffer is filled in but before volumes are rendered. There are other approaches to solving this, but this was a low-lift solution that doesn't increase our memory requirements.

src/PickVolume.ts
```diff
@@ -276,7 +277,6 @@ export default class PickVolume implements VolumeRenderImpl {
     this.setUniform("inverseModelViewMatrix", mvm);
     this.setUniform("inverseProjMatrix", camera.projectionMatrixInverse);
 
-    const VOLUME_LAYER = 0;
```
ShrimpCryptid, line 279: Used constant instead of defining this inline

src/ThreeJsPanel.ts
```diff
@@ -709,29 +717,47 @@ export class ThreeJsPanel {
...
+    });
+
+    // Step 3: Render meshes that do not interact with the pick buffer. This
+    // must happen after the pick buffer is rendered so picking isn't occluded
+    // by them, but before the volume renders so that volumes can still depth
+    // test against the lines.
+    this.renderer.autoClear = false;
+    this.camera.layers.set(MESH_NO_PICK_OCCLUSION_LAYER);
+    this.renderer.setRenderTarget(this.meshRenderTarget);
+    this.renderer.render(this.scene, this.camera);
```
ShrimpCryptid, line 738: Again, this ordering fixes this visual bug where outlines were being rendered around lines.

Before:
Image

After:
Image

src/constants/shaders/colorizeUI.frag
```diff
@@ -120,9 +120,9 @@ vec4 getObjectColor(ivec2 sUv, float opacity) {
 
   // color the highlighted object. Note, `highlightedId` is a 0-based index
   // (global ID w/o offset), while `id` is a 1-based index.
-  if (id - 1u == highlightedId) {
-    return vec4(outlineColor, 1.0);
-  }
+  // if (id - 1u == highlightedId) {
+  //   return vec4(outlineColor, 1.0);
+  // }
```
ShrimpCryptid, line 126: Debating whether to just remove this altogether, but `colorizeUI.frag` is mostly adapted from TFE's colorize shader so I thought it might be good to keep them structurally similar?

src/constants/shaders/contour.frag
```diff
@@ -0,0 +1,79 @@
...
+  int R = int(getId(uv + ivec2(thicknessFloat * wStep, 0))) - ID_OFFSET;
+  int L = int(getId(uv + ivec2(-thicknessFloat * wStep, 0))) - ID_OFFSET;
+  int T = int(getId(uv + ivec2(0, thicknessFloat * hStep))) - ID_OFFSET;
+  int B = int(getId(uv + ivec2(0, -thicknessFloat * hStep))) - ID_OFFSET;
+  // if any neighbors are not highlightedId then color this as edge
+  return id != -1 && (R != id || L != id || T != id || B != id);
+}
+
+void main(void) {
+  ivec2 vUv = ivec2(int(gl_FragCoord.x / devicePixelRatio), int(gl_FragCoord.y / devicePixelRatio));
```
ShrimpCryptid, line 69: TODO: Check if devicePixelRatio is still necessary

src/Line3d.ts
ShrimpCryptid, line 35: Dan: Add comments about mesh rendering order into the code here

src/RenderToBuffer.ts
```diff
@@ -22,14 +22,15 @@ export default class RenderToBuffer {
   public mesh: Mesh;
   public camera: OrthographicCamera;
 
-  constructor(fragmentSrc: string, uniforms: { [key: string]: IUniform }) {
+  constructor(fragmentSrc: string, uniforms: { [key: string]: IUniform }, transparent = false) {
```
ShrimpCryptid: Comment this constructor and/or change `transparent` to an enum

src/ThreeJsPanel.ts
```diff
@@ -713,41 +717,53 @@ export class ThreeJsPanel {
...
```
ShrimpCryptid, line 738: Dan: Add an `if` check and skip this step if there are no lines

ShrimpCryptid: @toloudis tinkered around with this and there wasn't really an elegant way to do this, since `ThreeJsPanel` doesn't have direct access to `VolumeDrawable`. I could have `View3D` be responsible for setting/removing a flag in `ThreeJsPanel` (`hasNoPickMeshes = true/false`), do you think it's worth it for the optimization?

src/constants/shaders/contour.frag
```diff
@@ -0,0 +1,79 @@
...
+  vec4 finalColor = vec4(0, 0, 0, 0.0);
+
+  uint rawId = getId(vUv);
+  int id = int(rawId) - ID_OFFSET;
+  if (id == highlightedId && isEdge(vUv, id, outlineThickness)) {
+    finalColor = vec4(outlineColor, outlineAlpha);
+  }
+
+  gl_FragColor = finalColor;
```
frasercl: why reassign `finalColor` in the `if` rather than assigning `gl_FragColor` directly in the `if` and `else` branches?

src/VolumeDrawable.ts
```diff
@@ -513,6 +525,13 @@ export default class VolumeDrawable {
     this.volumeRendering.updateActiveChannels(this.fusion, this.volume.channels);
     // pickRendering only really works with one channel so we don't need to call
     // its updateActiveChannels method
+    if (this.pickRendering) {
+      const pickChannel = this.pickRendering.getChannelToPick();
+      const channelData = this.volume.channels[pickChannel];
+      if (channelData) {
+        this.contourRendering.setTime(channelData.time);
```
frasercl: ...remind me why `Channel` has a `time` property? Do we ever expect it to be out of sync with its parent volume's `time`?

frasercl (approved):
LGTM overall!

src/PickVolume.ts
```diff
@@ -261,11 +265,9 @@ export default class PickVolume implements VolumeRenderImpl {
 
     // this.channelData.gpuFuse(renderer);
 
-    // set up texture from segmentation channel!!!!
-    // we need to know the channel index for this.
-    // ...channel.dataTexture...
-    // TODO TODO TODO FIXME
     this.setUniform("textureAtlas", this.volume.getChannel(this.channelToPick).dataTexture);
+    const channelImage = this.volume.getChannel(this.channelToPick).dataTexture.image;
```
toloudis: nit: move this up 1 line so that the line above can use the same aliased variable

src/VolumeDrawable.ts
ShrimpCryptid: Channel has a `time` property for colorizing. The current time is used to get a lookup from local IDs (seg IDs) to their position in global data arrays for feature colorizing. There's a different lookup for each frame, since seg IDs can be reused between frames.

src/constants/shaders/contour.frag
ShrimpCryptid: Refactored!

src/VolumeDrawable.ts
toloudis: This is something that caught my eye too.  Is there a better word than "time" for what this means?  Because you have `frameToGlobalIdLookup` maybe you should use either "frame" for both or "time" for both.

toloudis: Two more thoughts on this.   

1. Maybe this could have been cleaner if the Channel wasn't responsible for having a `frame`(or time) member variable, because it's not really part of the state of a Channel.   It's more a part of the fusion state? 

2. Or if we want to consider it part of the channel object's state, it will definitely get out of sync with the "time" because it's defaulting to 0 and nothing else is setting it yet.  Maybe all the rest of the code needs to properly pass the value in and it shouldn't be defaulted...

ShrimpCryptid: I was thinking about it as per-channel because different channels can have different timesteps loaded... and you want picking/colorizing to stay in sync with the channel it's set to even if the other channels aren't synchronized to that timestamp yet.

PR #334, opened 2025-07-28
refactor: Ran formatting on fragment shaders

ShrimpCryptid:
# Problem

I ran the formatter on all the `.frag` shaders, so edits to them in the future won't incur huge diffs.

### DIFF
--- a/src/constants/shaders/pathtrace.frag
+++ b/src/constants/shaders/pathtrace.frag
@@ -10,12 +10,12 @@ precision highp sampler3D;
 #define INV_2_PI (0.5/PI)
 #define INV_4_PI (0.25/PI)
 
-const vec3 BLACK = vec3(0,0,0);
-const vec3 WHITE = vec3(1.0,1.0,1.0);
+const vec3 BLACK = vec3(0, 0, 0);
+const vec3 WHITE = vec3(1.0, 1.0, 1.0);
 const int ShaderType_Brdf = 0;
 const int ShaderType_Phase = 1;
 const int ShaderType_Mixed = 2;
-const float MAX_RAY_LEN = 1500000.0f;
+const float MAX_RAY_LEN = 1500000.0;
 
 in vec2 vUv;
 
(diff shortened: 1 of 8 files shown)


src/constants/shaders/pathtrace.frag
```diff
@@ -119,34 +118,29 @@ vec3 RGBtoXYZ(vec3 rgb) {
   return rgb * RGB_2_XYZ;
 }
 
-vec3 getUniformSphereSample(in vec2 U)
-{
-  float z = 1.f - 2.f * U.x;
-  float r = sqrt(max(0.f, 1.f - z*z));
-  float phi = 2.f * PI * U.y;
+vec3 getUniformSphereSample(in vec2 U) {
+  float z = 1. - 2. * U.x;
```
toloudis, line 122: did the formatter know it was running on GLSL code?  what language does it think it was formatting?  Just double checking when i see the removal of the `.f` suffixes on these floating point literals.

ShrimpCryptid: Yes! I'm using this extension here: https://marketplace.visualstudio.com/items?itemName=raczzalan.webgl-glsl-editor

https://wikis.khronos.org/opengl/Data_Type_(GLSL)#Literals
> A numeric literal that uses a decimal is by default of type float. To create a float literal from an integer value, use the suffix f or F as in C/C++.
> Floats can also be defined using exponential notation, using the e or E to separate the base from the exponent. The exponent is always base-10.
> In GLSL 4.00 and above, double-precision floats are available. By default, all floating-point literals are of type float. To create a double-precision float, use the lf or LF suffixes. This will force it to the double type.

ShrimpCryptid: I think it's automatically removing them because not every version of GLSL supports `f` suffixes. If we included a version directive in the shader I think it would add them back in.

https://github.com/racz16/WebGL-GLSL-Editor/issues/46

toloudis: ah, I believe threejs might auto-insert a version directive

toloudis (approved):
I guess this is better than not formatting!

PR #336, opened 2025-08-06
allow arbitrary dtypes for RawArrayLoader

toloudis:
# Problem

In support of nbvv, because channel data now supports any numeric dtype, nbvv no longer needs to convert data to uint8.  However, the RawArrayLoader needs to relax its requirement for uint8 data.

# Solution

Let RawArrayLoader have any dtype.
Update the test app to load procedurals with 3 different dtypes.

The RawArrayData type is no longer required to be uint8 only.  I believe this is not breaking but haven't yet tested against the vole-app.

## Type of change

## Steps to Verify:

Can look at test app and load the 3 procedurals and verify that they look the same.

### DIFF
--- a/src/VolumeMaker.ts
+++ b/src/VolumeMaker.ts
@@ -1,3 +1,5 @@
+import { type NumberType, type TypedArray, ARRAY_CONSTRUCTORS } from "./types.js";
+
 /**
  * Basic utility functions to create sample volume data
  * @class
@@ -9,14 +11,17 @@ export default class VolumeMaker {
    * @param {number} vy
    * @param {number} vz
    * @param {function} sdFunc A function f(x,y,z) that returns a distance. f < 0 will be the interior of the volume, and f>=0 will be outside.
+   * @param {TypedArrayConstructor} arrayConstructor The TypedArray constructor to use
    */
   static createVolume(
     vx: number,
     vy: number,
     vz: number,
-    sdFunc: (px: number, py: number, pz: number) => number
-  ): Uint8Array {
-    const data = new Uint8Array(vx * vy * vz).fill(0);
+    sdFunc: (px: number, py: number, pz: number) => number,
+    dtype: NumberType = "uint8"
+  ): TypedArray<NumberType> {
+    const ctor = ARRAY_CONSTRUCTORS[dtype];
+    const data = new ctor(vx * vy * vz).fill(0);
     const cx = vx / 2;
     const cy = vy / 2;
     const cz = vz / 2;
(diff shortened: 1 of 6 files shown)


src/VolumeMaker.ts
```diff
@@ -9,14 +11,17 @@ export default class VolumeMaker {
...
     vy: number,
     vz: number,
-    sdFunc: (px: number, py: number, pz: number) => number
-  ): Uint8Array {
-    const data = new Uint8Array(vx * vy * vz).fill(0);
+    sdFunc: (px: number, py: number, pz: number) => number,
+    dtype: NumberType = "uint8"
+  ): TypedArray<NumberType> {
+    const ctor = ARRAY_CONSTRUCTORS[dtype];
+    const data = new ctor(vx * vy * vz).fill(0);
```
frasercl, line 24: Should there be a lookup for the max & min values at the given `dtype` below, rather than constant `255`?

src/loaders/RawArrayLoader.ts
```diff
@@ -42,7 +41,26 @@ export interface RawArrayLoaderOptions {
...
+    case "uint32":
+    case "int32":
+    case "float32":
+      return 4;
+    case "float64":
+      return 8;
+    default:
+      throw new Error(`Unsupported dtype: ${dtype}`);
+  }
+}
```
frasercl, line 61: Could/should these string literals be the number types from types.ts? (e.g. `Int8` rather than `"int8"`?) (should `NumberType` be a `const enum`...?)

toloudis: I believe all the cases here should be type checked against NumberType because of the argument's type.  But yeah they could be replaced by the symbolic names. 
Hesitant to change NumberType to enum due to possible farther ranging side effects but I can try it easily enough.  I'll try it.

src/VolumeMaker.ts
toloudis: This did occur to me.  But histogram binning made this work fine for the test cases I added for a couple different dtypes.  (Basically all these volume generation functions are for testing stuff out without having images around, so I didn't take it as seriously as I might for other parts of the code).  I'd suggest maybe adding your idea in a future PR. We could make a ticket for it.  The function arguments could even accept an intensity value.

src/loaders/RawArrayLoader.ts
toloudis: After a quick check, we use the string literals in several other places and rely on TS type checking to keep it consistent.  I think switching to symbolic names will go hand in hand with transforming it to an enum and also will cause some other types to get changed downstream (like the way TypedArray is defined). I'm pretty sure the format we are using here imitates zarrita.  So I would consider changing it in a separate PR but it's kinda out of scope for this one.

ShrimpCryptid (approved):

PR #338, opened 2025-08-12
feat: Fixed bin size calculation, fractional bin calculation

ShrimpCryptid:
# Problem

What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

# Solution

What I/we did to solve this problem

with @pairperson1

## Type of change

Please delete options that are not relevant.

- This change requires a documentation update
- This change requires updated or new tests

## Steps to Verify:

1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

## Screenshots (optional):

1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/Histogram.ts
+++ b/src/Histogram.ts
@@ -84,9 +84,19 @@ export default class Histogram {
     return binIndex;
   }
 
-  // return the bin index of the given data value
+  /** Returns the integer bin index for the given value. */
   public findBinOfValue(value: number): number {
     return Histogram.findBin(value, this.min, this.binSize, NBINS);
+  }
+
+  /**
+   * Returns a fractional bin index for the given value. If a value
+   * is not exactly a bin boundary, returns an interpolated index.
+   * Note that this can return a value outside the range of 0 to NBINS - 1.
+   */
+  public findFractionalBinOfValue(value: number): number {
+    const binIndex = (value - this.min) / this.binSize;
+    return binIndex;
   }
 
   /**
@@ -260,7 +270,7 @@ export default class Histogram {
 
     const bins = new Uint32Array(numBins).fill(0);
 
-    const binSize = (max - min) / numBins === 0 ? 1 : (max - min) / numBins;
+    const binSize = (max - min) / numBins === 0 ? 1 : (max - min) / (numBins - 1);
     for (let i = 0; i < arr.length; i++) {
       const item = arr[i];
 
--- a/src/test/lut.test.ts
+++ b/src/test/lut.test.ts
@@ -3,6 +3,8 @@ import { Lut, remapLut, remapControlPoints } from "../Lut";
 import Histogram from "../Histogram";
 import VolumeMaker from "../VolumeMaker";
 
+import { describe, expect, it } from "vitest";
+
 function clamp(val, cmin, cmax) {
   return Math.min(Math.max(cmin, val), cmax);
 }
@@ -112,6 +114,22 @@ describe("test histogram", () => {
     const secondlut = new Lut().createFromControlPoints(lutObj.controlPoints);
     it("generates consistent lut from control points", () => {
       expect(secondlut.lut).to.eql(lutObj.lut);
+    });
+  });
+
+  describe("maps from values to bins", () => {
+    const histogram = new Histogram(new Uint8Array([0, 127, 128, 255]));
+
+    it("maps values to the correct bins", () => {
+      expect(histogram.findBinOfValue(0)).to.equal(0);
+      expect(histogram.findBinOfValue(127)).to.equal(127);
+      expect(histogram.findBinOfValue(255)).to.equal(255);
+    });
+
+    it("can get interpolated bin values", () => {
+      expect(histogram.findFractionalBinOfValue(-1.5)).to.equal(-1.5);
+      expect(histogram.findFractionalBinOfValue(255)).to.equal(255);
+      expect(histogram.findBinOfValue(255)).to.equal(255);
     });
   });
 


src/Histogram.ts
```diff
@@ -75,18 +75,24 @@ export default class Histogram {
   }
 
   // return the bin index of the given data value
-  public static findBin(dataValue: number, dataMin: number, binSize: number, numBins: number): number {
-    let binIndex = Math.floor((dataValue - dataMin) / binSize);
-    // for values that lie exactly on last bin we need to subtract one
-    if (binIndex === numBins) {
-      binIndex--;
-    }
```
ShrimpCryptid: I changed how the binSize calculation is done, which meant that I could remove this edge case!

src/Histogram.ts
```diff
@@ -260,11 +266,12 @@ export default class Histogram {
 
     const bins = new Uint32Array(numBins).fill(0);
 
-    const binSize = (max - min) / numBins === 0 ? 1 : (max - min) / numBins;
+    // Subtract 1 so the last bin includes the max value.
+    const binSize = max - min === 0 ? 1 : (max - min) / (numBins - 1);
```
ShrimpCryptid: You can see this at work in the unit tests, but this makes it so the last bin has the `max` as the lower bin boundary and lets me remove some special-case logic. `vole-app` was already doing this when converting from bin indices (u8) back to absolute values.

ShrimpCryptid: Little proof:

Let's say we are working with 2-bit integers, and we have a histogram with four bins. The min of our data is 0 and the max is 3.

With the old calculation, our bin size would be `(max - min) / numBins = (3 - 0) / 4 = 0.75`, and our bins would have the following bounds:
- [0, 0.75)
- [0.75, 1.5)
- [1.5, 2.25)
- [2.25, 3)

This means that 3 is not actually included in the final bin, which led to the required special case in `findBin`.

With the new math, our bin size would be `(max - min) / (numBins - 1) = (3 - 0) / (4 - 1) = 1`. Our bins would have the following bounds:
- [0, 1)
- [1, 2)
- [2, 3)
- [3, 4)

src/Histogram.ts
```diff
@@ -75,18 +75,22 @@ export default class Histogram {
...
+  public static findBin(dataValue: number, dataMin: number, binSize: number): number {
+    return Math.floor((dataValue - dataMin) / binSize);
   }
 
-  // return the bin index of the given data value
+  /** Returns the integer bin index for the given value. */
   public findBinOfValue(value: number): number {
-    return Histogram.findBin(value, this.min, this.binSize, NBINS);
+    return Histogram.findBin(value, this.min, this.binSize);
+  }
```
frasercl: could this not just be `Math.floor(this.findFractionalBinOfValue(value))`?

src/Histogram.ts
```diff
@@ -75,18 +75,22 @@ export default class Histogram {
...
   // return the bin index of the given data value
-  public static findBin(dataValue: number, dataMin: number, binSize: number, numBins: number): number {
-    let binIndex = Math.floor((dataValue - dataMin) / binSize);
-    // for values that lie exactly on last bin we need to subtract one
-    if (binIndex === numBins) {
-      binIndex--;
-    }
-    return binIndex;
+  public static findBin(dataValue: number, dataMin: number, binSize: number): number {
+    return Math.floor((dataValue - dataMin) / binSize);
```
frasercl: A bit annoying that the same expression as below is also duplicated here, but that was an odd bit of API design that predates this PR. I'd say this ought to be the single place this calculation gets made, but this version also involves a `Math.floor` and so isn't usable in `findFractionalBinOfValue`...

src/Histogram.ts
toloudis: what if the data is a set of uniformly distributed random floating point numbers in the range [0,3] and we want 4 bins?  I would expect the bins to contain [0-.75, .75-1.5, 1.5-2.25 and 2.25-3] which would give us equal counts in each bin.

toloudis: perhaps an epsilon could help

toloudis: I'll write a more practical case: random integer intensities from 0-255.  We want 8 bins.  The bin bounds should be
```
1 | 0-31
2 | 32-63
3 | 64-95
4 | 96-127
5 | 128-159
6 | 160-191
7 | 192-223
8 | 224-255
```
For this type of data, I think you could write the min of bin 2 as 32-0.5 and the max of bin 1 as 31+0.5

toloudis: One more thing I thought of is that you could pass the dtype in to the Histogram if you had a reason to distinguish between integer values and floating point for the bin shapes.

kmitcham: The core question here is what is the expected, intuitive thing to do when the number of elements isn't a multiple of the number of bins.  Should one bin be truncated? Which one? Should the 'extra' elements be distributed as evenly as possible?  I don't see an obvious answer, but it should at least be explicitly documented.
Note: the number of bins is a constant: const NBINS = 256;
how the remainder is to be handled is still a question, though.

src/Histogram.ts
ShrimpCryptid: Ended up making `findBin` `private static` and reusing it in all the other methods.

src/Histogram.ts
ShrimpCryptid: Okay, added a big code comment and refactored the code to handle the float and integer cases!

src/Histogram.ts
```diff
@@ -74,19 +84,46 @@ export default class Histogram {
...
+   * is not exactly a bin boundary, returns an interpolated index.
+   * Note that this can return a value outside the range of 0 to NBINS - 1.
+   */
+  public findFractionalBinOfValue(value: number): number {
+    return Histogram.findBin(value, this.min, this.max, this.binSize, this.isFloatType, false);
+  }
+
+  /** Returns a raw data value from a given (integer or fractional) bin index. */
+  public getValueFromBin(binIndex: number): number {
+    return this.min + binIndex * this.binSize;
```
ShrimpCryptid: This will replace logic that `vole-app` was doing.

src/Histogram.ts
kmitcham: Not meaning to be difficult, but is all the code around calculating the number of bins needed?  The method is
`private static calculateHistogram(arr: TypedArray<NumberType>, numBins = 1)`
and only called once, 
`// build up the histogram
    const hinfo = Histogram.calculateHistogram(data, NBINS);`
and NBINS is a constant 256.
I think to code is correct, but overbuilt for the use case.  If calculateHistogram is intended for wider use in a later update, then it makes more sense.

kmitcham: EDIT: I was wrong.  This code isn't trying to evenly divide things into bins; it is creating bins of equal size, and not caring about how many items go in each bin.  I'd suggest:
histogram_first_bin_lowest_value = math.floor(min_array_value)
histogram_last_bin_highest_value = math.ceiling(max_array_value)
bin size = ( histogram_last_bin_highest_value - histogram_first_bin_lowest_value ) /NBINS
and remove the number_of_bins argument from the function.

kmitcham (approved):
Overbuilt, but looks to work.

src/Histogram.ts
ShrimpCryptid: I see what you mean about `calculateHistogram()` only ever being called with 256 bins. It's possible in the future we could have a differently-sized Histogram, but for now I *am* using a 4-wide histogram in some of my unit tests to match the code comment example since it's easier to reason about.

I'm not sure `math.floor` and `math.ceil` will work in the float case? For example, if our data range is a very small float range, (ex. `[0.45, 0.46]`), the Histogram would be from the range [0, 1], and most of the range would be wasted.

kmitcham: each bin would contains numbers for 1/256th of the range from 0 to 1

src/Histogram.ts
```diff
@@ -28,6 +28,15 @@ export default class Histogram {
...
   private dataMaxBin: number;
   private pixelCount: number;
+
+  /**
+   * Whether the Histogram represents continuous floating point data.
+   *
+   * If true, the histogram's last bin will have the max value of the data as
+   * its inclusive upper bound. If false, the histogram represents integer data,
+   * and the last bin will have an (exclusive) upper bound of `max + 1`.
+   */
```
toloudis: this comment is basically saying: either way, the max of the last bin is equal to the max of the data range. Maybe it could be clearer if written as:
If isFloatType === true, we have to do a little extra work to make sure the max of the last bin includes the max of the data.  For integer data this is more straightforward.

src/Histogram.ts
```diff
@@ -38,6 +47,7 @@ export default class Histogram {
     this.min = 0;
     this.max = 0;
     this.binSize = 0;
+    this.isFloatType = data instanceof Float32Array || data instanceof Float64Array;
```
toloudis: I don't love this but apart from passing in a dtype:NumberType argument, I guess it's ok.  Might consider dtype because it represents the whole range of types we support.. (Maybe add a isFloatType function to the types file and import it from there?)

src/Histogram.ts
```diff
@@ -74,19 +84,46 @@ export default class Histogram {
...
+    dataMax: number,
+    binSize: number,
+    isFloatType: boolean,
+    castToInt: boolean
+  ): number {
+    const binIndex = (dataValue - dataMin) / binSize;
+    if (!castToInt) {
+      return binIndex;
+    }
+    if (isFloatType && Math.abs(dataValue - dataMax) < 1e-6) {
```
toloudis: Nit: 
Do we ever do this query with dataValue>dataMax?
If it's always values from the histogrammed data array, then you don't need abs here, dataValue should always be <= dataMax.  In that sense, it's an error - we can't return a real bin number because the query doesn't make sense.  It's a query about data that's not a member of the histogram.

src/Histogram.ts
```diff
@@ -74,19 +84,46 @@ export default class Histogram {
...
+      // Max is inclusive, so near-exact matches should return the last bin.
+      return Math.round(binIndex) - 1;
     }
-    return binIndex;
+    if (Math.abs(binIndex) < 1e-6) {
```
toloudis: could this be `if binIndex<0` with same comment about float imprecision?  or am I wrong

src/Histogram.ts
```diff
@@ -74,19 +84,46 @@ export default class Histogram {
...
+    dataValue: number,
+    dataMin: number,
+    dataMax: number,
+    binSize: number,
+    isFloatType: boolean,
+    castToInt: boolean
+  ): number {
+    const binIndex = (dataValue - dataMin) / binSize;
+    if (!castToInt) {
+      return binIndex;
```
toloudis, line 81: related to my other comment, should this make some guarantees about out of range values?

src/Histogram.ts
```diff
@@ -74,19 +84,46 @@ export default class Histogram {
...
+   * boundary, returns an interpolated index. Note that this can return a value
+   * outside the range of 0 to NBINS - 1.
+   */
+  public findFractionalBinOfValue(value: number): number {
+    return Histogram.findBin(value, this.min, this.max, this.binSize, this.isFloatType, false);
+  }
+
+  /** Returns a raw data value from a given (integer or fractional) bin index. */
+  public getValueFromBin(binIndex: number): number {
+    return this.min + binIndex * this.binSize;
```
toloudis, line 109: this does not seem to be guaranteed to be a real data value from the data array that was histogrammed

src/Histogram.ts
```diff
@@ -260,11 +297,44 @@ export default class Histogram {
 
     const bins = new Uint32Array(numBins).fill(0);
 
-    const binSize = (max - min) / numBins === 0 ? 1 : (max - min) / numBins;
+    // In the ideal case for both float and integer data, we want to distribute
+    // values evenly across the bins. For example, if we have a data range of
```
toloudis: we want the bins to be of equal widths and span data min to data max. I think "distribute the values evenly across the bins" could be interpreted as saying you want the counts in each bin to be equal. For arbitrary data, some bins will have higher counts, and some will have less or none.

src/Histogram.ts
ShrimpCryptid: If we're not casting to int, we want an exact fractional bin value, even if it's out of range. I would be doing this math to make sure that the conversion from a value -> bin index -> value is stable.

The use case would be if a user sets up a Lut URL parameter for Vol-E from some value outside of the range (ex, `lut:v0:v100` when the data value range is `[50, 60]`). Clamping to the data value range would be unexpected.

src/Histogram.ts
ShrimpCryptid: Ended up clamping values in `findBin` which handles a lot of these cases, and adding an explicit comment in `findBinOfValue` about clamping behavior.

src/Histogram.ts
ShrimpCryptid: Comment updated

toloudis (approved):
LGTM!

src/Histogram.ts
```diff
@@ -74,19 +75,38 @@ export default class Histogram {
     }
   }
 
-  // return the bin index of the given data value
-  public static findBin(dataValue: number, dataMin: number, binSize: number, numBins: number): number {
-    let binIndex = Math.floor((dataValue - dataMin) / binSize);
-    // for values that lie exactly on last bin we need to subtract one
-    if (binIndex === numBins) {
-      binIndex--;
+  private static findBin(dataValue: number, dataMin: number, binSize: number, castToInt: boolean): number {
```
frasercl, line 78: maybe worth giving `castToInt` a default?

```suggestion
  private static findBin(dataValue: number, dataMin: number, binSize: number, castToInt = true): number {
```

src/Histogram.ts
```diff
@@ -260,11 +280,38 @@ export default class Histogram {
...
+    //
+    // |----|----|----|----|
+    // 0  0.75  1.5  2.25  3 <- inclusive
+    //
+    let binSize: number;
+    if (isFloatTypeArray(arr)) {
+      binSize = max - min === 0 ? 1 : (max - min) / numBins;
+    } else {
+      binSize = max + 1 - min === 0 ? 1 : (max + 1 - min) / numBins;
+    }
```
frasercl: 1. Looks like the only thing that varies between these branches is `max`? If so, that could be broken out into one statement, something like `const realMax = isFloatTypeArray(arr) ? max : max + 1;`. Then you don't have to do the declare-without-initializing dance with `binSize`.
2. Should the condition in those ternary operators be checking for `<= 0`? I suppose it's mainly guarding against dividing zero, though you never quite know with floats if two numbers will come out exactly the same. (Follow this thought far enough and it's a decent argument for adding an `EPSILON` to this check, but Kevin is right that this is on the overbuilt side already)
3. That condition is also logically equivalent to `max === min` (or `max <= min`), though I see how it's semantically clearer to do it this way.

src/Histogram.ts
ShrimpCryptid: Chatted about this, I'm going to still require the boolean flag because I want callers (which should only be internally within Histogram) to be sure about the kind of values they expect.

src/Histogram.ts
ShrimpCryptid: Changed, thank you

PR #343, opened 2025-09-03
Feat/source channel counts

ShrimpCryptid:
# Problem

What is the problem this work solves, including
[Link to story or ticket](https://my-tracking-system.url/ticket-number)

# Solution

What I/we did to solve this problem

with @pairperson1

## Type of change

Please delete options that are not relevant.

- This change requires a documentation update
- This change requires updated or new tests

## Steps to Verify:

1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

## Screenshots (optional):

1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/src/ImageInfo.ts
+++ b/src/ImageInfo.ts
@@ -20,6 +20,7 @@ export type ImageInfo = Readonly<{
    * NOTE Currently there is one ImageInfo per Volume, not per source.
    */
   combinedNumChannels: number;
+  numChannelsPerSource: number[];
   /** The names of each channel */
   channelNames: string[];
   /** Optional overrides to default channel colors, in 0-255 range, RGB order */
@@ -56,6 +57,7 @@ export function defaultImageInfo(): ImageInfo {
     subregionSize: [1, 1, 1],
     subregionOffset: [0, 0, 0],
     combinedNumChannels: 1,
+    numChannelsPerSource: [1],
     channelNames: ["0"],
     channelColors: [[255, 255, 255]],
     multiscaleLevel: 0,
@@ -90,6 +92,11 @@ export class CImageInfo {
   /** Number of channels in the image */
   get numChannels(): number {
     return this.imageInfo.combinedNumChannels;
+  }
+
+  /** Number of channels per source, ordered by source index */
+  get numChannelsPerSource(): number[] {
+    return this.imageInfo.numChannelsPerSource;
   }
 
   /** XYZ size of the *original* (not downsampled) volume, in pixels */
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -729,12 +729,14 @@ export default class VolumeDrawable {
     // TODO only one channel can ever have this?
     if (!featureInfo) {
       this.fusion[channelIndex].feature = undefined;
+      this.contourRendering.setGlobalIdLookup(null);
     } else {
       this.fusion[channelIndex].feature = featureInfo;
+      this.contourRendering.setOutlineColor(featureInfo.outlineColor);
+      this.contourRendering.setGlobalIdLookup(featureInfo.frameToGlobalIdLookup);
     }
     this.volumeRendering.updateSettings(this.settings, SettingsFlags.MATERIAL);
     this.pickRendering?.updateSettings(this.settings, SettingsFlags.MATERIAL);
-    this.contourRendering.setColorizeFeature(featureInfo);
   }
 
   setMaskAlpha(maskAlpha: number): void {
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -487,7 +487,7 @@ export default class VolumeDrawable {
     this.pickRendering?.doRender(renderer, camera, depthTexture);
   }
 
-  drawContours(renderer: WebGLRenderer, camera: PerspectiveCamera | OrthographicCamera): void {
+  drawContours(renderer: WebGLRenderer): void {
     if (!this.pickRendering) {
       return;
     }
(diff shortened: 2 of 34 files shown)


src/ImageInfo.ts
```diff
@@ -14,12 +14,18 @@ export type ImageInfo = Readonly<{
...
   subregionOffset: [number, number, number];
 
-  /** Number of channels in the image, accounting for convergence of multiple sources.
+  /**
+   * Number of channels in the image, accounting for convergence of multiple sources.
    * Because of multiple sources, which is not accounted for in ImageInfo,
    * that this could be different than the number of channels in the multiscaleLevelDims.
    * NOTE Currently there is one ImageInfo per Volume, not per source.
    */
   combinedNumChannels: number;
```
ShrimpCryptid: Not sure if I should remove `combinedNumChannels` since it's derivable from `numChannelsPerSource`?

src/types.ts
```diff
@@ -129,28 +129,34 @@ export interface FuseChannel {
 /** If `FuseChannel.rgbColor` is this value, it is disabled from fusion. */
 export const FUSE_DISABLED_RGB_COLOR = 0;
```
ShrimpCryptid, line 130: Added comments here because I was getting confused by the accepted value range of each option

src/loaders/TiffLoader.ts
```diff
@@ -219,6 +219,7 @@ class TiffLoader extends ThreadableVolumeLoader {
       subregionSize: [tilesizex, tilesizey, dims.sizez],
       subregionOffset: [0, 0, 0],
       combinedNumChannels: dims.sizec,
+      numChannelsPerSource: [dims.sizec],
```
toloudis: tiff has had a recent change to support multi-source (`this.url` is an array) so this is not gonna be quite correct.  Wondering if now is the time to fix this

src/ImageInfo.ts
toloudis: > Not sure if I should remove `combinedNumChannels` since it's derivable from `numChannelsPerSource`?

yes?  not sure how many places it's used but I would prefer having the array

src/types.ts
```diff
@@ -129,28 +129,34 @@ export interface FuseChannel {
...
+  /**
+   * Exponent factor controlling the glossiness ("shininess") of the material. 0
+   * is default.
+   */
   glossiness?: number;
+  /** Whether the isosurface mesh should be rendered for this channel. */
   isosurfaceEnabled?: boolean;
+  /**
+   * Isovalue used to calculate the isosurface mesh, in a [0, 255] range.
+   * Intensity values below this will not be included in the isosurface.
```
toloudis:
```suggestion
   * Isosurface is found at the set of all boundaries between voxels whose intensities span across this isovalue.
```

src/ImageInfo.ts
ShrimpCryptid: Okay, will do. I think it's only used in one place internally, but since it's an externally-accessible property I'll mark this as a breaking change when I do a release later.

src/types.ts
```diff
@@ -129,28 +129,34 @@ export interface FuseChannel {
...
   specularColor?: [number, number, number];
+  /**
+   * RGB color array for emissive (glow) color, with values in the range of [0,
+   * 255].
+   */
   emissiveColor?: [number, number, number];
+  /**
+   * Exponent factor controlling the glossiness ("shininess") of the material. 0
+   * is default.
+   */
```
frasercl: nit: did a formatter want these 80 wide? would love to have them on one line if possible (this is wildly unimportant but I'm always a fan of a single-line doc comment when possible)

src/VolumeDrawable.ts
```diff
@@ -905,6 +905,10 @@ export default class VolumeDrawable {
     return this.childObjects.has(line);
   }
 
+  hasLineObjects(): boolean {
+    return this.childObjects.size > 0;
```
frasercl: Does `childObjects` ever have anything other than lines in it? I'm seeing isosurface meshes get put in there too I think

src/ImageInfo.ts
```diff
@@ -14,12 +14,11 @@ export type ImageInfo = Readonly<{
...
-  /** Number of channels in the image, accounting for convergence of multiple sources.
-   * Because of multiple sources, which is not accounted for in ImageInfo,
-   * that this could be different than the number of channels in the multiscaleLevelDims.
-   * NOTE Currently there is one ImageInfo per Volume, not per source.
+  /**
+   * The number of channels in each source, in source order. The sum of this
+   * array is equal to `combinedNumChannels`.
    */
-  combinedNumChannels: number;
+  numChannelsPerSource: number[];
```
frasercl, line 19: But `combinedNumChannels` went away. Remove? (could be one line then :eyes:)

src/loaders/TiffLoader.ts
```diff
@@ -211,14 +211,15 @@ class TiffLoader extends ThreadableVolumeLoader {
...
+    const numChannelsPerSource = this.url.length > 1 ? Array(this.url.length).fill(1) : [dims.sizec];
 
     const imgdata: ImageInfo = {
       name: "TEST",
 
       atlasTileDims: [atlasDims.x, atlasDims.y],
       subregionSize: [tilesizex, tilesizey, dims.sizez],
       subregionOffset: [0, 0, 0],
-      combinedNumChannels: dims.sizec,
+      numChannelsPerSource: numChannelsPerSource,
```
frasercl:
```suggestion
      numChannelsPerSource,
```

frasercl (approved):

src/types.ts
ShrimpCryptid: Yeah, it's a manual formatter though so I can just undo it.

src/ImageInfo.ts
ShrimpCryptid: Whoops! Good catch

src/VolumeDrawable.ts
```diff
@@ -89,7 +89,7 @@ export default class VolumeDrawable {
 
     this.channelColors = this.volume.channelColorsDefault.slice();
 
-    this.channelOptions = new Array<VolumeChannelDisplayOptions>(this.volume.imageInfo.numChannels).fill({});
```
ShrimpCryptid, line 92: It turns out that this declaration would cause the whole array to be filled with the same object instance (`{}`). This caused a bug where disabling any channel would cause all channels to disappear on the next load, because the change made to one channel index would, in effect, be made on all channel indices.

src/VolumeDrawable.ts
ShrimpCryptid: Good question, and yeah, it would contain the meshes. I'm going to actually just remove this API because I don't think I ended up needing to call it anywhere.

PR #348, opened 2025-10-15
Read z size for non-OME TIFFs

frasercl:
Review time: tiny (<5min)

#327 added very basic support for TIFFs that don't conform to the OME metadata spec. But the images that motivated that work were 2D, so I missed how to properly read Z dimensions from this type of image. This is a 1-line change that does that.

### DIFF
--- a/src/loaders/TiffLoader.ts
+++ b/src/loaders/TiffLoader.ts
@@ -168,9 +168,11 @@ class TiffLoader extends ThreadableVolumeLoader {
           } catch (_e) {}
         }
 
+        // if `ImageDescription` is valid JSON with a `shape` field, we expect it to be an array of [t?, c?, z?, y, x].
         this.dims.sizex = shape[shape.length - 1] ?? image.getWidth();
         this.dims.sizey = shape[shape.length - 2] ?? image.getHeight();
         this.dims.sizez = shape[shape.length - 3] ?? (await tiff.getImageCount());
+
         // TODO this is a big hack/assumption about only loading multi-source tiffs that are not OMETIFF.
         // We really have to check each url in the array for sizec to get the total number of channels
         // See combinedNumChannels in ImageInfo below.


src/loaders/TiffLoader.ts
```diff
@@ -159,6 +159,7 @@ class TiffLoader extends ThreadableVolumeLoader {
         this.dims = new OMEDims();
         this.dims.sizex = image.getWidth();
         this.dims.sizey = image.getHeight();
+        this.dims.sizez = await tiff.getImageCount();
```
frasercl: Notice the comment above that says "DO NOT DO THIS, ITS SLOW." Regrettably, without extra metadata to store how many slices the TIFF contains, this also seems to be the only way to get this info. (I gave this a quick `console.time` and it adds ~400-600ms for a 30-slice image; a peek under the hood shows it's O(n) with the number of slices.)

toloudis: I had a few minutes to look at one of the tiffs from the BFF discussion that was going around, and it looked like its imageinfo was json of the form `{"shape":[z, y, x]}`.  AGAVE knows how to parse that data too.. See here just for reference: 
https://github.com/allen-cell-animated/agave/blob/main/renderlib/io/FileReaderTIFF.cpp#L344

AGAVE's tiff reader parses a few different flavors of tiff metadata, in a simple way to try to extract the basic dims.

toloudis: It was this file https://ftp.ebi.ac.uk/pub/databases/biostudies/S-BIAD/156/S-BIAD2156/Files/microtubules/cell_images/1046_262039_signal.tiff and AGAVE found 31 slices.   Unfortunately AGAVE doesn't load tiffs from urls, and I had to pre-download it first.

toloudis: Anyway, using getImageCount is better than nothing for sure and a decent fallback to try to guess at z.  I'm quite sure when I wrote that comment that I was testing it with variance dataset that had multichannel tiffs so it had something like 9*65 slices and was really taking a lot of extra time.

frasercl: Didn't spot that `ImageDescription`; added code to parse out `shape` if it exists

toloudis (approved):
LGTM.  I had to double check to make sure that negative array indexing returns undefined instead of throwing... 
With BFF being in the mix now, I can foresee adding more of the TIFF ImageDescription formats that you see in the agave code.

src/loaders/TiffLoader.ts
```diff
@@ -149,21 +149,40 @@ class TiffLoader extends ThreadableVolumeLoader {
...
-        this.dims.sizex = image.getWidth();
-        this.dims.sizey = image.getHeight();
+        let shape: number[] = [];
+        try {
+          const description = JSON.parse(image0Description);
+          if (Array.isArray(description.shape)) {
+            shape = description.shape;
+          }
+          // eslint-disable-next-line no-empty
+        } catch (_e) {}
```
ShrimpCryptid: Should this be logged to a console?

src/loaders/TiffLoader.ts
```diff
@@ -149,21 +149,40 @@ class TiffLoader extends ThreadableVolumeLoader {
...
+
+        this.dims.sizex = shape[shape.length - 1] ?? image.getWidth();
+        this.dims.sizey = shape[shape.length - 2] ?? image.getHeight();
+        this.dims.sizez = shape[shape.length - 3] ?? (await tiff.getImageCount());
```
ShrimpCryptid, line 174: Might be worth adding a comment here that shape is expected to be in `// [t?, c?, z, y, x]` order

ShrimpCryptid (approved):
Left some comments but LGTM!

src/loaders/TiffLoader.ts
frasercl: Probably not &mdash; what we're looking for here is the equivalent of a more sensible method returning `undefined` to indicate the input is not valid JSON. Just above, we try to parse the same field as XML, because OME-TIFF expects that format. So it's not unexpected if this description isn't parseable by this method.

PR #351, opened 2025-10-22
refactor: Added an abstract class that implements IDrawableObject to reduce boilerplate

ShrimpCryptid:
# Problem

This change supports #354, which implements instanced vector arrow rendering in Vol-E!

*Estimated review size: small, 10-15 minutes*

# Solution

- Generalized APIs for adding child objects (`addDrawableObject`) and deprecated line-specific APIs (`addLineObject`)
- Created an abstract class (`BaseDrawableObject`) for shared object scaling behavior and refactored `Line3d` to extend it.
  - This will be extended by a vector drawing class in the next PR.

## Type of change

## Screenshots (optional):

Screenshot in TFE to show that the track path still works in Vol-E with the new changes:
image

### DIFF
--- a/src/BaseDrawableMeshObject.ts
+++ b/src/BaseDrawableMeshObject.ts
@@ -10,7 +10,7 @@ import { IDrawableObject } from "./types";
  * As a default, subclasses should call `this.addChildMesh(mesh)` to register
  * any meshes that should be managed by the base class.
  */
-export default abstract class BaseDrawableMesh implements IDrawableObject {
+export default abstract class BaseDrawableMeshObject implements IDrawableObject {
   /**
    * Pivot group that all child meshes are parented to. Transformations are
    * applied to this group.
@@ -19,46 +19,28 @@ export default abstract class BaseDrawableMesh implements IDrawableObject {
   protected scale: Vector3;
   protected flipAxes: Vector3;
 
-  /**
-   * Child meshes that are part of this drawable object.
-   */
-  protected readonly meshes: Set<Mesh>;
-
   constructor() {
     this.meshPivot = new Group();
     this.scale = new Vector3(1, 1, 1);
     this.flipAxes = new Vector3(1, 1, 1);
-
-    this.meshes = new Set<Mesh>();
   }
 
   protected addChildMesh(mesh: Mesh): void {
-    if (!this.meshes.has(mesh)) {
-      this.meshes.add(mesh);
-      this.meshPivot.add(mesh);
-    }
+    this.meshPivot.add(mesh);
   }
 
   protected removeChildMesh(mesh: Mesh): void {
-    if (this.meshes.has(mesh)) {
-      this.meshes.delete(mesh);
-      this.meshPivot.remove(mesh);
-    }
-  }
-
-  protected forEachMesh(callback: (mesh: Mesh) => void): void {
-    for (const mesh of this.meshes) {
-      callback(mesh);
-    }
+    this.meshPivot.remove(mesh);
   }
 
   cleanup(): void {
-    this.forEachMesh((mesh) => {
-      mesh.geometry.dispose();
-      (mesh.material as Material).dispose();
+    this.meshPivot.traverse((obj) => {
+      if (obj instanceof Mesh) {
+        obj.geometry.dispose();
+        (obj.material as Material).dispose();
+      }
     });
     this.meshPivot.clear();
-    this.meshes.clear();
   }
 
   doRender(): void {
--- a/src/Line3d.ts
+++ b/src/Line3d.ts
@@ -5,7 +5,7 @@ import { LineSegmentsGeometry } from "three/addons/lines/LineSegmentsGeometry.js
 
 import { IDrawableObject } from "./types.js";
 import { MESH_NO_PICK_OCCLUSION_LAYER, OVERLAY_LAYER } from "./ThreeJsPanel.js";
-import BaseDrawableMesh from "./BaseDrawableMesh.js";
+import BaseDrawableMeshObject from "./BaseDrawableMeshObject.js";
 
 const DEFAULT_VERTEX_BUFFER_SIZE = 1020;
 
@@ -13,7 +13,7 @@ const DEFAULT_VERTEX_BUFFER_SIZE = 1020;
  * Simple wrapper for a 3D line segments object, with controls for vertex data,
  * color, width, and segments visible.
  */
-export default class Line3d extends BaseDrawableMesh implements IDrawableObject {
+export default class Line3d extends BaseDrawableMeshObject implements IDrawableObject {
   private lineMesh: LineSegments2;
   private bufferSize: number;
 


src/BaseDrawableObject.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Provides default implementations for some methods in the IDrawableObject
+ * interface, including handling for visibility, transformations, and cleanup.
+ *
+ * As a default, subclasses should:
+ * 1. Add all meshes that need to be disposed of during cleanup to the `meshes`
+ *    array.
+ * 2. Add all meshes as children of the `meshPivot` group for transformation
+ *    handling (rotation, scaling, translation).
+ */
+export default class BaseDrawableObject implements IDrawableObject {
```
toloudis: this probably should be named BaseMeshDrawableObject since the stuff to draw is a Group of Mesh.

src/BaseDrawableMeshObject.ts
```diff
@@ -0,0 +1,97 @@
...
+  doRender(): void {
+    // no op
+  }
+
+  setVisible(visible: boolean): void {
+    this.meshPivot.visible = visible;
+  }
+
+  get3dObject(): Group {
+    return this.meshPivot;
```
toloudis, line 55: I don't see where meshPivot gets the this.meshes as children.  Which confuses me as to what this class is helping with.

src/View3d.ts
```diff
@@ -977,29 +978,54 @@ export class View3d {
...
    * (0,0,0) is at the center of the Volume and the extent is from -0.5 to 0.5
    * in each axis.
+   * @deprecated Will be removed in the next major release. Use `addDrawableObject` instead.
    */
   addLineObject(line: Line3d): void {
-    if (this.image) {
-      this.image.addLineObject(line);
-      this.redraw();
-    }
+    return this.addDrawableObject(line);
```
toloudis, line 984: I am wondering if making these more generic/polymorphic makes it harder to reorder them for specific rendering needs.  just a random thought.

src/Line3d.ts
```diff
@@ -36,66 +35,10 @@ export default class Line3d implements IDrawableObject {
...
-    // no op
-  }
-
-  setAxisClip(_axis: "x" | "y" | "z", _minval: number, _maxval: number, _isOrthoAxis: boolean): void {
-    // no op
-  }
-
-  updateClipRegion(_xmin: number, _xmax: number, _ymin: number, _ymax: number, _zmin: number, _zmax: number): void {
-    // no op
+    this.meshes = this.lineMesh;
```
toloudis: `addChild(lineMesh);` perhaps?

ShrimpCryptid: Agree! Updated the API to have `addChildMesh` and `removeChildMesh` functions

src/BaseDrawableMesh.ts
```diff
@@ -0,0 +1,109 @@
...
+   * applied to this group.
+   */
+  protected readonly meshPivot: Group;
+  protected scale: Vector3;
+  protected flipAxes: Vector3;
+
+  /**
+   * Child meshes that are part of this drawable object.
+   */
+  protected readonly meshes: Set<Mesh>;
```
toloudis: this may be a strange question but why are they managed separately if they are in meshPivot.children?

src/BaseDrawableMesh.ts
```diff
@@ -0,0 +1,109 @@
...
+  constructor() {
+    this.meshPivot = new Group();
+    this.scale = new Vector3(1, 1, 1);
+    this.flipAxes = new Vector3(1, 1, 1);
+
+    this.meshes = new Set<Mesh>();
+  }
+
+  protected addChildMesh(mesh: Mesh): void {
+    if (!this.meshes.has(mesh)) {
```
toloudis: I wonder if this should be managed better from outside. Or maybe throw some kind of error if a mesh got added twice. How likely could it happen?

src/BaseDrawableMesh.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Abstract base class for drawable 3D mesh objects.
+ *
+ * Provides default implementations for some methods in the IDrawableObject
+ * interface, including handling for visibility, transformations, and cleanup.
+ *
+ * As a default, subclasses should call `this.addChildMesh(mesh)` to register
+ * any meshes that should be managed by the base class.
+ */
+export default abstract class BaseDrawableMesh implements IDrawableObject {
```
toloudis: the class name makes me think it's for one single mesh

src/BaseDrawableMesh.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  protected removeChildMesh(mesh: Mesh): void {
+    if (this.meshes.has(mesh)) {
+      this.meshes.delete(mesh);
+      this.meshPivot.remove(mesh);
+    }
+  }
+
+  protected forEachMesh(callback: (mesh: Mesh) => void): void {
+    for (const mesh of this.meshes) {
```
toloudis: this kind of code makes me wonder that maybe keeping the list of meshes AND the parent with its children is redundant.  threejs groups give methods of iterating their children.

src/BaseDrawableMeshObject.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  setTranslation(translation: Vector3): void {
+    this.meshPivot.position.copy(translation);
+  }
+
+  setScale(scale: Vector3): void {
+    this.scale.copy(scale);
+    this.meshPivot.scale.copy(scale).multiply(this.flipAxes);
```
toloudis, line 64: is it now redundant to store scale in here?  in fact could this whole object extend three.Group?  I am sorry for even asking this question.  I suppose in a way it's good that it's _not_ a threejs class.

src/Line3d.ts
```diff
@@ -33,69 +32,11 @@ export default class Line3d implements IDrawableObject {
     // artifacts can occur where contours are drawn around lines. This layer
     // (MESH_NO_PICK_OCCLUSION_LAYER) does not occlude/interact with the pick
     // buffer but still writes depth information for the volume.
+    this.meshPivot.layers.set(MESH_NO_PICK_OCCLUSION_LAYER);
```
toloudis, line 35: do children inherit their parent's layers? (it looks like the following line does the same to children of meshPivot)

src/BaseDrawableMeshObject.ts
toloudis: My comment about storing scale is dependent on how this.scale is used outside of the base class but tsc would probably find all of them quickly.

ShrimpCryptid: Arrow drawing in Vol-E uses `this.scale`, but also overrides `setScale` to do so. I'm debating if it's worth keeping around a copy of the scale though that can be easily accessed, or if classes extending `BaseDrawableMesh` (or whatever I rename this class) should do it themselves.

src/BaseDrawableMesh.ts
ShrimpCryptid: Premature optimization, haha. `meshPivot.children` is an array and I thought it would be faster for some operations (checking if an object exists) to use a Set. I did do some checks and `meshPivot.add` is safe against multiple object adds, so it's likely that Three already has a set representation internally.

src/BaseDrawableMesh.ts
ShrimpCryptid: Is BaseDrawableMeshObject any better?

src/BaseDrawableMesh.ts
ShrimpCryptid: Adding/removing child meshes is used by extending classes, and usually they have a predetermined mesh that they're adding (1-2 meshes only). I think it's unlikely that a mesh would get added more than once, but if it does Three already handles that case for us (the child is only added once).

src/Line3d.ts
ShrimpCryptid: https://github.com/mrdoob/three.js/issues/10959
Children do not! You have to set it on both.

PR #354, opened 2025-10-28
feat: Support drawing vector arrows

### LINKED ISSUE #286: TFE 3D - Show vector arrows
## Use Case

Support vector arrow drawing in 3D viewport to replicate existing TFE 2D viewport behavior.

## Solution

- [ ] Allow clients to provide an array of vectors with a 3D origin and direction/magnitude, as well as a color.
- [ ] Draw vector arrows from specified origin to specified direction, in the provided colors.

## Alternatives

- Should vector arrows be 3D, made up of cylinders + cones like AGAVE?
  - Or should they be lines only? lines with thickness?

ShrimpCryptid:
# Problem

Closes #286, "TFE 3D - Show vector arrows".

# Solution
- Generalized APIs for adding child objects (`addDrawableObject`) and deprecated line-specific APIs (`addLineObject`)
- Added `VectorArrows3d`, which supports instanced rendering for a field of vector arrows.
  - Thickness, color, and vector position + delta can be set per-vector.
  - The maximum instance count is dynamically increased.

*Estimated review size: medium, 30 minutes*

## Type of change

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/adbd5073-fdb7-4c1a-9c7f-cb65fc4ca866

### DIFF
--- a/src/BaseDrawableMeshObject.ts
+++ b/src/BaseDrawableMeshObject.ts
@@ -64,10 +64,6 @@ export default abstract class BaseDrawableMeshObject implements IDrawableObject 
     this.meshPivot.scale.copy(scale).multiply(this.flipAxes);
   }
 
-  onParentTransformUpdated(): void {
-    // no op
-  }
-
   setRotation(eulerXYZ: Euler): void {
     this.meshPivot.rotation.copy(eulerXYZ);
   }
--- a/src/VectorArrows3d.ts
+++ b/src/VectorArrows3d.ts
@@ -14,11 +14,13 @@ import { IDrawableObject } from "./types";
 import BaseDrawableMeshObject from "./BaseDrawableMeshObject";
 import { MESH_NO_PICK_OCCLUSION_LAYER } from "./ThreeJsPanel";
 
-const DEFAULT_DIAMETER = 0.002;
-
-// Unscaled arrowhead dimensions. These will be scaled by the diameter.
+// Unscaled arrowhead dimensions.
+const SHAFT_BASE_RADIUS = 0.5;
 const HEAD_BASE_RADIUS = 1.5;
 const HEAD_BASE_HEIGHT = 4;
+
+/** Default arrow shaft thickness, in world units. */
+const DEFAULT_DIAMETER = 0.002;
 
 const DEFAULT_INSTANCE_COUNT = 256;
 
@@ -96,7 +98,14 @@ export default class VectorArrows3d extends BaseDrawableMeshObject implements ID
   private static generateGeometry(): { head: BufferGeometry; shaft: BufferGeometry } {
     // TODO: Currently the shape of the arrow head is fixed. Allow configuring
     // this in the future?
-    const cylinderGeometry = new CylinderGeometry(0.5, 0.5, 1, 8, 1, false);
+    const cylinderGeometry = new CylinderGeometry(
+      SHAFT_BASE_RADIUS,
+      SHAFT_BASE_RADIUS,
+      2 * SHAFT_BASE_RADIUS, // height
+      8, // radial segments
+      1, // height segments
+      false // capped ends
+    );
     const coneRadius = HEAD_BASE_RADIUS;
     const coneHeight = HEAD_BASE_HEIGHT;
     const coneGeometry = new ConeGeometry(coneRadius, coneHeight, 12);
@@ -245,7 +254,6 @@ export default class VectorArrows3d extends BaseDrawableMeshObject implements ID
       // Points and deltas scaled to volume space.
       tempSrc.fromArray(this.positions, i * 3).multiply(combinedScale);
       tempDelta.fromArray(this.deltas, i * 3).multiply(combinedScale);
-      // Fallback in case 0-length diameter array is provided
       tempDiameter = this.diameter[i % this.diameter.length] ?? DEFAULT_DIAMETER;
       this.updateSingleArrowTransform(i, tempSrc, tempDelta, tempDiameter);
     }


src/BaseDrawableMeshObject.ts
```diff
@@ -27,6 +29,7 @@ export default class BaseDrawableObject implements IDrawableObject {
   constructor() {
     this.meshPivot = new Group();
     this.scale = new Vector3(1, 1, 1);
+    this.parentScale = new Vector3(1, 1, 1);
```
toloudis: why are both scale and parentScale in here?

src/types.ts
```diff
@@ -100,13 +100,18 @@ export interface ColorizeFeature {
   hideOutOfRange: boolean;
 }
 
+export type ScaleInfo = {
+  parentScale: Vector3;
+};
```
toloudis: why is this wrapped in a type?

src/VectorArrows3d.ts
```diff
@@ -0,0 +1,334 @@
...
+    this.meshes = [headMesh, shaftMesh];
+
+    return { headMesh, shaftMesh };
+  }
+
+  private increaseInstanceCountMax(instanceCount: number): void {
+    // Max instance count is set when instanced meshes are created. If we need
+    // to increase the max, we need to recreate the instanced meshes.
+    let newInstanceCount = this.maxInstanceCount;
+    while (newInstanceCount < instanceCount) {
```
toloudis, line 159: is the instanceCount for a single frame (t) ?

ShrimpCryptid: `instanceCount` is usually for the single frame, yes, and will vary between frames in how TFE uses it.

src/types.ts
ShrimpCryptid: Doesn't need to be, I wasn't sure if I'd need to extend scale info in the future. YAGNI?

src/VectorArrows3d.ts
toloudis: I just wanted to make sure it's a per frame count and not a total across all frames.  What do you mean by "usually"?

ShrimpCryptid: Oops, that's me being imprecise with my language. Yes, this data is updated per-frame, and should only contain the currently visible vectors on that frame.

If and when we add support for custom vector data in TFE, the number of vectors rendered could be totally arbitrary though.

src/types.ts
ShrimpCryptid: Update: This could be useful in the future if we want to store callbacks for when other scale properties are updated. TBD!

src/BaseDrawableMeshObject.ts
ShrimpCryptid: I had it there in case other functions needed to access it, but it is unused by every other extending class. I'll move it into the vector arrows class.

src/BaseDrawableMeshObject.ts
```diff
@@ -64,6 +64,10 @@ export default abstract class BaseDrawableMeshObject implements IDrawableObject
     this.meshPivot.scale.copy(scale).multiply(this.flipAxes);
   }
 
+  onParentTransformUpdated(): void {
```
toloudis: the IDrawableObject interface says this is optional, so could you delete it from here?

src/VectorArrows3d.ts
```diff
@@ -0,0 +1,349 @@
...
+    const combinedScale = new Vector3().copy(this.scale).multiply(this.flipAxes).multiply(this.worldScale);
+
+    const tempSrc = new Vector3();
+    const tempDelta = new Vector3();
+    let tempDiameter: number;
+    for (let i = 0; i < count; i++) {
+      // Points and deltas scaled to volume space.
+      tempSrc.fromArray(this.positions, i * 3).multiply(combinedScale);
+      tempDelta.fromArray(this.deltas, i * 3).multiply(combinedScale);
+      // Fallback in case 0-length diameter array is provided
```
toloudis: is a 0-length diameter array provided often?  if so, maybe the comment could say why

package.json
```diff
@@ -17,6 +17,7 @@
     "build": "npm run transpileES && npm run build-types",
     "build-types": "tsc -p tsconfig.types.json",
     "build-demo": "vite build public/ --config vite.config.ts --outDir ./demo",
+    "checks": "npm run lint & npm run typeCheck & npx vitest run",
```
toloudis, line 20: nit: I like this but out of scope?  seeing this line makes me think all the CI gh actions should use it, also makes me think it should be called out in readme or contributing docs.

src/VectorArrows3d.ts
```diff
@@ -0,0 +1,349 @@
...
+   * |     |   /   \
+   *  --x--   /-----\
+   *
+   *   x = pivot (0,0,0)
+   * ```
+   */
+  private static generateGeometry(): { head: BufferGeometry; shaft: BufferGeometry } {
+    // TODO: Currently the shape of the arrow head is fixed. Allow configuring
+    // this in the future?
+    const cylinderGeometry = new CylinderGeometry(0.5, 0.5, 1, 8, 1, false);
```
toloudis: nit: can't help but think DEFAULT_DIAMETER should be showing up in here.  but mostly, the ConeGeometry below uses symbolic constants and this one doesn't.

src/VectorArrows3d.ts
```diff
@@ -0,0 +1,349 @@
...
+    const coneGeometry = new ConeGeometry(coneRadius, coneHeight, 12);
+
+    // Rotate both to point along +Z axis
+    const rotateToPositiveZ = new Matrix4().makeRotationX(Math.PI / 2);
+    // Change cone pivot to be at the tip.
+    const coneTranslation = new Matrix4().makeTranslation(0, 0, -coneHeight / 2);
+    coneGeometry.applyMatrix4(coneTranslation.multiply(rotateToPositiveZ));
+    // Change cylinder pivot to be at the base.
+    const cylinderTranslation = new Matrix4().makeTranslation(0, 0, 0.5);
+    cylinderGeometry.applyMatrix4(cylinderTranslation.multiply(rotateToPositiveZ));
```
toloudis, line 120: I like this use of applyMatrix4 to get these set up the way you want them

src/VectorArrows3d.ts
```diff
@@ -0,0 +1,349 @@
...
+
+  /**
+   * Called when scaling of parent transforms has been updated or whenever
+   * vector data is updated.
+   */
+  public onParentTransformUpdated(): void {
+    // Measure world scale by temporarily resetting mesh pivot scale
+    this.meshPivot.scale.set(1, 1, 1);
+    let newWorldScale = new Vector3();
+    newWorldScale = this.meshPivot.getWorldScale(newWorldScale);
```
toloudis, line 189: love this solution, this seems nice and elegant

src/VectorArrows3d.ts
```diff
@@ -0,0 +1,349 @@
...
+    // of parent transforms, and prevents distortion or skewing of the mesh.
+    // Parent scaling is  applied to arrow positions and deltas (see
+    // `updateAllArrowTransforms`), rather than the meshes themselves.
+    const invertScale = new Vector3(1, 1, 1).divide(newWorldScale);
+    this.meshPivot.scale.copy(invertScale);
+
+    if (!newWorldScale.equals(this.worldScale)) {
+      this.worldScale.copy(newWorldScale);
+      if (this.positions && this.deltas) {
+        this.setArrowData(this.positions, this.deltas);
```
toloudis, line 203: this gets called pretty infrequently, right?  Does it happen on an arrow option slider drag in TFE?  But basically nowhere else?

toloudis (approved):

package.json
ShrimpCryptid: Mm, I think it's good if CI gh actions run each step individually. I added this for my convenience :X

src/VectorArrows3d.ts
ShrimpCryptid: It shouldn't, but it's possible? 

One scenario I can see happening is if there are no vector arrows on the current frame, so the positions, deltas, and diameter arrays are all empty. And then a later API call only provides positions and deltas.

I can also just remove the comment!

toloudis: if there are no vector arrows, how would we get all the way into this code?  Anyway I agree with some combination of removing the comment, (or adding a comment about the scenario where this array lookup can return undefined), removing the extra conditional, or checking for these edge cases way earlier

src/VectorArrows3d.ts
ShrimpCryptid: `setArrowData` is called in TFE on each frame change, when arrow scaling is updated, or when the underlying data is changed (e.g. changing motion delta smoothing).

ShrimpCryptid: But during playback it should only be called once for each frame

PR #355, opened 2025-11-12
Allow `undefined` image names

frasercl:
Review time: tiny (<5min)

Fixes multiple issues with reading image names from metadata:

- `OmeZarrLoader` calls all images without names "Volume".
- `TiffLoader` calls all images "TEST" (whether or not they have a name).
- Both are required to call images *something*, since the `name` field of `ImageInfo` (and `Volume`) cannot be `undefined`.

As to that last point, I don't think anything strictly depends on `name` being defined. The code added in allen-cell-animated/vole-app#409 to show image names is already ready for `undefined` `name`s.

### DIFF
--- a/src/loaders/TiffLoader.ts
+++ b/src/loaders/TiffLoader.ts
@@ -34,6 +34,7 @@ function getOME(xml: string | undefined): Element | undefined {
 }
 
 class OMEDims {
+  name: string | undefined = undefined;
   sizex = 0;
   sizey = 0;
   sizez = 1;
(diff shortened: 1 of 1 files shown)


src/Volume.ts
```diff
@@ -31,7 +31,7 @@ export default class Volume {
   public loadSpecRequired: Required<LoadSpec>;
   public channelLoadCallback?: PerChannelCallback;
   public imageMetadata: Record<string, unknown>;
-  public name: string;
+  public name: string | undefined;
```
meganrm, line 34: is there a reason to type this specifically as undefined instead of optional?

frasercl: I wanted to distinguish that `Volume` and `ImageInfo` should always *have* a `name` property, even if it's `undefined`. Though, thinking of it, I don't know what might go wrong technically without `name` there at all.

PR #377, opened 2026-02-27
feat: Support selection of multiple IDs, multiple colors

ShrimpCryptid:
# Problem

Supporting changes for https://github.com/allen-cell-animated/timelapse-colorizer/issues/310 in TFE's 3D mode! This change allows multiple onscreen segmentations to be selected at once, and also allows different colors to be applied. 

*Estimated review size: medium, 25-30 minutes*

# Solution

- Updated shader ID handling for clarity, added support for multiple outlined objects + outline colors.
- Added `outlinePalette` and `useOutlinePalette` texture + flag to the ColorizeFeature type.
- Added setter for selected ID LUT.

Misc:
- Removes unused setters/types for old method of highlighting IDs, which involved highlighting them in the fuse step.

## Type of change

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/93c6bc00-74f5-4df4-983c-b33864b4a7be

image

### DIFF
--- a/src/ContourPass.ts
+++ b/src/ContourPass.ts
@@ -166,6 +166,7 @@ export default class ContourPass {
     }
     this.selectedIdsTexture = new DataTexture(paddedSelectedIds, width, height, RedIntegerFormat, UnsignedByteType);
     this.selectedIdsTexture.internalFormat = "R8UI";
+    this.selectedIdsTexture.unpackAlignment = 1;
     this.selectedIdsTexture.needsUpdate = true;
     this.pass.material.uniforms.selectedIds.value = this.selectedIdsTexture;
   }
--- a/src/ContourPass.ts
+++ b/src/ContourPass.ts
@@ -98,10 +98,11 @@ export default class ContourPass {
   }
 
   /**
-   * Optional inner outline shown when using outline palettes for better contrast.
+   * Optional inner outline shown for better contrast, specified in integer
+   * pixels. Disabled if `thicknessPx` is 0.
    */
   public setInnerOutlineThickness(thicknessPx: number): void {
-    this.pass.material.uniforms.innerOutlineThickness.value = Math.floor(thicknessPx);
+    this.pass.material.uniforms.innerOutlineThickness.value = Math.floor(Math.max(0, thicknessPx));
   }
 
   private syncGlobalIdLookup(): void {
@@ -151,20 +152,21 @@ export default class ContourPass {
   }
 
   /**
-   * Sets the lookup table that maps from an ID to whether the ID is
-   * selected or not.
+   * Sets the lookup table that maps from an ID to whether the ID is selected or
+   * not.
    *
    * For some ID `i`, if `selectedIds[i] > 0`, the ID is selected. When
-   * `useOutlinePalette` is true, the value of `selectedIds[i]` is used
-   * as the index of the outline color in the outline palette.
+   * `useOutlinePalette` is true, `selectedIds[i] - 1` is the index of the
+   * outline color in the outline palette.
    *
-   * By default, the ID is a local (pixel) ID. If a global ID
-   * lookup has been set (`setGlobalIdLookup`), the ID is parsed as a global ID.
+   * By default, the ID is a local (pixel) ID. If a global ID lookup has been
+   * set (`setGlobalIdLookup`), the ID is parsed as a global ID.
    */
   public setSelectedIdLut(selectedIds: Uint8Array) {
     if (this.selectedIdsTexture) {
       this.selectedIdsTexture.dispose();
     }
+    // Pack into square texture
     const [width, height] = getSquarestTextureDimensions(selectedIds.length);
     let paddedSelectedIds = selectedIds;
     if (selectedIds.length < width * height) {
--- a/src/ContourPass.ts
+++ b/src/ContourPass.ts
@@ -43,11 +43,15 @@ type ContourUniforms = {
 const makeDefaultUniforms = (): ContourUniforms => {
   // RGBA float texture for pick buffer
   const pickBufferTex = new DataTexture(new Float32Array([0, 0, 0, 0]), 1, 1, RGBAFormat, FloatType);
+  pickBufferTex.internalFormat = "RGBA32F";
+  pickBufferTex.needsUpdate = true;
   // R32UI texture for local ID to global ID lookup
   const localIdToGlobalId = new DataTexture(new Uint32Array([0]), 1, 1, RedIntegerFormat, UnsignedIntType);
+  localIdToGlobalId.internalFormat = "R32UI";
   localIdToGlobalId.needsUpdate = true;
   // RGBA float texture for outline palette
   const outlinePaletteTex = new DataTexture(new Float32Array([1, 1, 1, 0]), 1, 1, RGBAFormat, FloatType);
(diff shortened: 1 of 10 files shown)


src/types.ts
```diff
@@ -124,8 +126,6 @@ export interface FuseChannel {
   lut: Uint8Array;
   // zero is a sentinel value to disable from fusion
   rgbColor: [number, number, number] | number;
-  // the selected id will have its intensity auto-mapped to a pick color
-  selectedID: number;
```
ShrimpCryptid, line 128: This was unused in FuseChannel

src/constants/shaders/contour.frag
```diff
@@ -16,14 +16,32 @@ uniform bool useGlobalIdLookup;
 /* Pick buffer. Used to determine IDs. */
```
ShrimpCryptid, line 16: I refactored `contour.frag` to do its ID lookup closer to the logic in TFE

src/VolumeDrawable.ts
```diff
@@ -731,10 +726,11 @@ export default class VolumeDrawable {
     // TODO only one channel can ever have this?
     if (!featureInfo) {
       this.fusion[channelIndex].feature = undefined;
-      this.contourRendering.setGlobalIdLookup(null);
```
ShrimpCryptid, line 734: This was causing a bug where the outlines would disappear if any colorize-related parameters were changed (like the color ramp min max, for example).

`setChannelColorizeFeature(channelIdx, null)` was being called on channels other than the one being colorized, which cleared the global ID lookup. This meant that objects could not be matched to their global IDs, and therefore would not get outlined.

src/FusedChannelData.ts
ShrimpCryptid, line 1: Misc change, attempted to fix a texture format mismatch warning in WebGL.

src/VolumeDrawable.ts
```diff
@@ -731,10 +726,11 @@ export default class VolumeDrawable {
...
     } else {
       this.fusion[channelIndex].feature = featureInfo;
       this.contourRendering.setOutlineColor(featureInfo.outlineColor, featureInfo.outlineAlpha);
+      this.contourRendering.setOutlinePaletteTexture(featureInfo.outlinePalette);
+      this.contourRendering.setUseOutlinePalette(featureInfo.useOutlinePalette);
       this.contourRendering.setGlobalIdLookup(featureInfo.frameToGlobalIdLookup);
     }
```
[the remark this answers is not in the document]
rugeli: I spent sometime tracing the "stale settings" after you removed `setGlobalIdLookup(null)` without calling it elsewhere instead. It seems like you handled it in shader's selection check so no outlines are drawn anyways? Nice fix!

rugeli (approved):
nice improvement! I had a few questions while reviewing, but it mostly make sense, moving from fuse() to contour pass seems efficient.

src/constants/shaders/contour.frag
```diff
@@ -32,14 +54,39 @@ uvec4 getUintFromTex(usampler2D tex, int index) {
...
+ */
+int getGlobalId(uint labelId) {
+  if (labelId == BACKGROUND_ID) {
+    return MISSING_DATA_ID;
   }
-  int localId = int(rawId) - int(localIdOffset);
+  int localId = int(labelId) - int(localIdOffset);
   if (!useGlobalIdLookup) {
-    return uint(localId + ID_OFFSET);
+    return localId - ID_OFFSET;
```
toloudis, line 89: this sign change is confusing, also the subtracted ID_OFFSET in the return statement.  After seeing the changes to isEdge, it makes a but more sense.

src/ContourPass.ts
```diff
@@ -16,29 +17,53 @@ import { clamp } from "three/src/math/MathUtils.js";
...
+  const pickBufferTex = new DataTexture(new Float32Array([0, 0, 0, 0]), 1, 1, RGBAFormat, FloatType);
+  // R32UI texture for local ID to global ID lookup
   const localIdToGlobalId = new DataTexture(new Uint32Array([0]), 1, 1, RedIntegerFormat, UnsignedIntType);
   localIdToGlobalId.needsUpdate = true;
+  // RGBA float texture for outline palette
+  const outlinePaletteTex = new DataTexture(new Float32Array([1, 1, 1, 0]), 1, 1, RGBAFormat, FloatType);
+  outlinePaletteTex.needsUpdate = true;
+  // R8UI texture for selected IDs
+  const selectedIds = new DataTexture(new Uint8Array([0]), 1, 1, RedIntegerFormat, UnsignedByteType);
+  selectedIds.internalFormat = "R8UI";
```
toloudis, line 58: it doesn't look like internalFormat is being explicitly set for the others. why not?

src/ContourPass.ts
```diff
@@ -108,8 +147,54 @@ export default class ContourPass {
    * @param id The ID to highlight. If a global ID lookup has been set
    * (`setGlobalIdLookup`), this should be a global ID.
    */
-  public setHighlightedId(id: number) {
-    this.pass.material.uniforms.highlightedId.value = id;
+  public setSelectedId(id: number) {
```
toloudis, line 154: I didn't notice any callers of setHighlightedId being changed to use setSelectedId. Are there client apps that call this, like the testbed app or vole-app?

src/VolumeDrawable.ts
```diff
@@ -508,16 +507,13 @@ export default class VolumeDrawable {
...
-    this.contourRendering.setHighlightedId(id);
-    if (this.fusion.length > 0) {
-      // TODO does it make sense to do this for a particular channel?
-      if (id !== this.fusion[channelIndex].selectedID) {
-        this.fusion[channelIndex].selectedID = id;
-        return true;
-      }
-    }
-    return false;
+  // TODO: Deprecate?
```
toloudis: why deprecate?  because setSelectedIdsLut replaces it?

toloudis (commented):
I left a few questions...
Also, I wonder when we can update the colorize functions in vole-app and the test app here to use only the new colorizer code paths.

src/VolumeDrawable.ts
ShrimpCryptid: Yes. I think this function (selecting a single ID) is still useful and is a simpler API, but now the shader has to support both pathways.

ShrimpCryptid: I'll update this comment too.

src/ContourPass.ts
ShrimpCryptid: This did change in `VolumeDrawable.ts`, old L512! Any client apps are calling `setSelectedID` on `View3D`, which calls it on `VolumeDrawable`, which calls it on the `ContourPass`.

src/VolumeDrawable.ts
ShrimpCryptid: Actually decided to remove this TODO, because I think having two different APIs (one lighter, one heavier-duty) makes sense to me based on client use.

src/ContourPass.ts
ShrimpCryptid: I think I set this while I was thrashing on trying to get my shaders to compile correctly. I think this is probably neutral-- it's most likely redundant with the DataTexture declaration but also doesn't hurt to specify?

toloudis (approved):
🚀

PR #387, opened 2026-03-31
feat: Add linting check for missing file extensions

ShrimpCryptid:
# Problem

Closes #375, "add eslint plugin check for missing file extensions."

I also added in the eslint plugin for import checks, since it's been helpful in TFE.

*Estimated review size: tiny, 5 minutes*

# Solution
- Added `eslint-plugin-import` and `eslint-import-resolver-typescript` packages to the project.
- Added a check for missing file extensions using the `no-restricted-syntax` rule.
- Updated some imports based on the new linting rules.

## Type of change

## Steps to Verify:

See that checks pass below!

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/e697621d-42fb-4ff2-b8a2-ca206e866435

1. `eslintrc.json`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -56,17 +56,6 @@
     "@typescript-eslint/no-this-alias": ["warn"],
     "prefer-const": ["warn"],
     "prefer-spread": ["warn"],
-    "import/extensions": [
-      "error",
-      "ignorePackages",
-      {
-        "js": "always",
-        "mjs": "always",
-        "jsx": "always",
-        "ts": "never",
-        "tsx": "never"
-      }
-    ],
     "no-var": ["warn"],
     // note you must disable the base rule as it can report incorrect errors
     "no-unused-vars": "off",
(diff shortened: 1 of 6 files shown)


.eslintrc.json
```diff
@@ -44,6 +59,28 @@
...
+    // Check for relative imports that are missing a file extension.
+    "no-restricted-syntax": [
+      "error",
+      {
+        // Breakdown of the regex: ^..?/([\w]+/)*[\w]+$
+        // 1. ^..?/ => checks for ./ or ../ at start of path
+        // 2. ([\w]+/)* => any additional paths between the start and final file name
+        // 3. [\w]+$ => a final file name that has no file extension (`.` character missing)
+        // The below regex is escaped (/ => \\x2f, . => \\., etc.) for parsing.
+        "selector": "ImportDeclaration[source.value=/^\\.\\.?\\x2f([\\w]+\\x2f)*[\\w]+$/]",
```
toloudis, line 72: Easier to read without the escaped characters.  It's hard for me to believe we should add this to every typescript repo we have.    Is there a web reference link that could be a source for this, that you could throw in the comments?

ShrimpCryptid: There isn't solid documentation for eslint's ImportDeclaration unfortunately... 

We don't need to do this for other repos, since they  use bundlers (webpack, vite) instead of only using babel directly. Babel doesn't handle file resolution, so needing to include `.js` extensions is a problem specific to `vole-core`'s configuration. 

(Maybe the real solution is to add `vite` to `vole-core`? https://vite.dev/guide/build#library-mode)

I did try configuring the eslint-imports plugin to handle this instead of writing a regex, but it didn't know how to alias `.ts` imports as `.js` and would flag every import in the entire project :(

toloudis: Would using vite's library mode for vole-core resolve this?  ... Also, this is not the only repo that provides ts -->  js classes as a library.

ShrimpCryptid: Ah, I was just thinking about the top-level viewers. Yeah, simularium-viewer also imports `.js` extensions...

Using vite would remove the need to strictly import with the `.js` extension, but I tried tinkering around with it and realized it would take more effort to get the config working correctly. And since all off the tools that consume vole-core are doing their own minifying + tree-shaking + bundling they probably don't need vole-core to be doing it too. So potentially more trouble than it's worth, ignore my suggestion :X

toloudis:
@frasercl does #395 negate this pr?

PR #391, opened 2026-04-08
Bugfix: frame dropping

### LINKED ISSUE #388: Many frames skipped
## Description

When playing a timelapse, sometimes many frames are skipped (for up to 10-20s) despite their data being loaded.

## Expected Behavior

If frames are loading more slowly than the framerate, they should be displayed within 100ms of being downloaded.

## Reproduction

* Load [this ZARR](https://vol-e.allencell.org/viewer?url=https%3A%2F%2Fallencell.s3.us-east-1.amazonaws.com%2Faics%2Femt_timelapse_dataset%2Fdata%2F3500005548_45_raw_converted.ome.zarr%2F)
* Open up dev tools and enable network throttling at a rate of 30Mbps down, 15Mbps up (this makes the issue much more consistent, but I have observed it in the wild, too).
* Switch to XY view and click play on the T axis.
* Watch closely for a moment where the T slider and timestamp update (indicating that the application believes it has rendered the next frame), but the image data doesn't. On my machine, there is a "freeze" like this that lasts for multiple seconds and multiple skipped frames about 90% of the time.

## Environment

* MacOS
* Reproduced on both Firefox and Chrome

pgarrison:
# Problem
Resolves #388 

# Root cause

`gpuFuse` references `Channels` fields, but those `Channels` are mutable, which allows the following race condition.

1. `fuse` for channel N checks that all `Channel.loaded` values are true and queues `gpuFuse` via `requestAnimationFrame`
2. `changeViewerSetting("time", N+1)` (vole-app) queues a react state update.
3. React update from 2 happens before the rAF from 1, triggering a `useEffect` in `vole-app`'s `App` that calls `setTime(view3d, N+1)`, which calls `vole-core`'s `Volume.setUnloaded`, updating the `Channel.loaded` values to false. 
4. rAF from 1 triggers `gpuFuse`, which checks `Channel.loaded`, sees false, and doesn't render the frame.

## Secondary bugs
The above is how the main bug works, but the fundamental issue is that mutable Channel fields can cause race conditions. There are a handful of those fields, beyond just `.loaded`. This PR should prevent all such bugs.

# Solution
* Remove the `Channel.loaded` check from `gpuFuse`.
* Instead of storing references to `Channel`s, `fuse` takes a snapshot of the channel data to fuse and puts it into a `Fusable` (better name, anyone?) which does not get modified. `gpuFuse` just reads from this `Fusable`.
* I also moved the `rgbColor` check (which I don't entirely understand) from `gpuFuse` to `fuse` so that all the logic about which channels to fuse is in one place.

## Type of change

## Steps to Verify:

See reproduction steps from #388.

## Performance

In my testing, this change gets us very roughly an extra 1-2FPS.

### DIFF
--- a/src/FusedChannelData.ts
+++ b/src/FusedChannelData.ts
@@ -36,7 +36,7 @@ import type { FuseChannel, NumberType } from "./types.js";
  * Immutable snapshot of the Channel fields needed by gpuFuse, captured at fuse() time.
  * Snapshot is required to avoid race conditions because Channels are mutable.
  */
-type Fusable = FuseChannel & {
+type FusableChannelState = FuseChannel & {
   snapshot: {
     dtype: NumberType;
     dataTexture: DataTexture;
@@ -55,7 +55,7 @@ export default class FusedChannelData {
 
   public maskTexture: DataTexture;
 
-  private fuseRequested: Fusable[] | null;
+  private fuseRequested: FusableChannelState[] | null;
 
   private fuseGeometry: PlaneGeometry;
   private fuseMaterialF: ShaderMaterial;
@@ -225,7 +225,7 @@ export default class FusedChannelData {
      * mutating the live Channel objects between fuse() and the next
      * requestAnimationFrame
      */
-    const readyToFuse: Fusable[] = [];
+    const readyToFuse: FusableChannelState[] = [];
     for (let i = 0; i < combination.length; ++i) {
       const c = combination[i];
       const idx = c.chIndex;
--- a/src/FusedChannelData.ts
+++ b/src/FusedChannelData.ts
@@ -207,9 +207,10 @@ export default class FusedChannelData {
   }
 
   fuse(combination: FuseChannel[], channels: Channel[]): void {
-    // we can fuse if we have any loaded channels that are showing.
-    // actually, we can fuse if no channels are showing (but they are loaded), too.
-    let canFuse = false;
+    // Snapshot which channels are loaded and enabled right now,
+    // so gpuFuse() renders from this decision rather than re-reading
+    // mutable `channel.loaded` (which may be cleared by a concurrent load).
+    const readyToFuse: FuseChannel[] = [];
     for (let i = 0; i < combination.length; ++i) {
       const c = combination[i];
       const idx = c.chIndex;
@@ -217,17 +218,18 @@ export default class FusedChannelData {
         // set the lut in this fuse combination.
         // can optimize by calling combineLuts more lazily
         c.lut = channels[idx].combineLuts(c.rgbColor, c.lut);
-        canFuse = true;
-        //break;
+        if (c.rgbColor) {
+          readyToFuse.push(c);
+        }
       }
     }
-    if (!canFuse) {
+    if (readyToFuse.length === 0) {
       this.channelsDataToFuse = [];
-      this.fuseRequested = [];
+      this.fuseRequested = null;
       return;
     }
 
-    this.fuseRequested = combination;
+    this.fuseRequested = readyToFuse;
     this.channelsDataToFuse = channels;
   }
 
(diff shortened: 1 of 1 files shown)


src/FusedChannelData.ts
```diff
@@ -249,51 +268,46 @@ export default class FusedChannelData {
...
-          mat.uniforms.useRepeatingCategoricalColors.value = feature.useRepeatingColor;
-          mat.uniforms.outlineColor.value = feature.outlineColor;
-          mat.uniforms.outlierColor.value = feature.outlierColor;
-          mat.uniforms.outOfRangeColor.value = feature.outOfRangeColor;
-          mat.uniforms.outlierDrawMode.value = feature.outlierDrawMode;
-          mat.uniforms.outOfRangeDrawMode.value = feature.outOfRangeDrawMode;
-          mat.uniforms.hideOutOfRange.value = feature.hideOutOfRange;
+      const chIndex = combination[i].chIndex;
+      const snap = this.channelSnapshots.get(chIndex)!;
+      const isColorize = combination[i].feature !== undefined;
```
pgarrison, line 278: From here to the end, the only changes are replacing `channels[chIndex]` with `snap`. The diff looks big because the conditional from line 252 was removed.

toloudis: At the time we call gpuFuse, we are actually truly ready to prepare the volume data for rendering by looping over the channels to fuse and issuing a draw call for each one. "fuse" means blending the color mapped data of each channel into one single rgb-valued texture from which we will render the main view.

The conditional on `combination[i].rgbColor` is supposed to indicate that the channel is indeed participating to be fused.  (If I recall correctly, we overloaded rgbColor to set it to 0, instead of a [r,g,b], for this purpose when the channel is disabled/invisible and not intended to be fused).

Then the conditional on `channels[chIndex].loaded` is there to also skip the channel from participating in the fuse if there is no data for it yet.

src/FusedChannelData.ts
```diff
@@ -32,6 +32,19 @@ import fuseShaderSrcI from "./constants/shaders/fuseI.frag";
 import colorizeSrcUI from "./constants/shaders/colorizeUI.frag";
 import type { FuseChannel, NumberType } from "./types.js";
 
+// Immutable snapshot of the Channel fields needed by gpuFuse, captured at fuse() time.
+// Snapshot is required to avoid race conditions because Channels are mutable.
+type Fusable = FuseChannel & {
```
ShrimpCryptid: FuseState? FuseRenderState?

src/FusedChannelData.ts
```diff
@@ -32,6 +32,19 @@ import fuseShaderSrcI from "./constants/shaders/fuseI.frag";
 import colorizeSrcUI from "./constants/shaders/colorizeUI.frag";
 import type { FuseChannel, NumberType } from "./types.js";
 
+// Immutable snapshot of the Channel fields needed by gpuFuse, captured at fuse() time.
+// Snapshot is required to avoid race conditions because Channels are mutable.
```
ShrimpCryptid: Nit: Can you make these into docstring comments (`/** ... */`) so they show up in Intellisense?

src/FusedChannelData.ts
```diff
@@ -207,33 +218,41 @@ export default class FusedChannelData {
   }
 
   fuse(combination: FuseChannel[], channels: Channel[]): void {
-    // we can fuse if we have any loaded channels that are showing.
-    // actually, we can fuse if no channels are showing (but they are loaded), too.
-    let canFuse = false;
+    // Snapshot channel state now so gpuFuse() is immune to concurrent loads
+    // mutating the live Channel objects between fuse() and the next rAF.
```
ShrimpCryptid: Nit: I would spell out `requestAnimationFrame` because I think the abbreviation/acronym wasn't very obvious to me

src/FusedChannelData.ts
pgarrison: Happy with FuseState, curious what others think.

src/FusedChannelData.ts
pgarrison: [91f801d](https://github.com/allen-cell-animated/vole-core/pull/391/commits/91f801df7d0b9720f9a5e19061447579a7d86fe1)

src/FusedChannelData.ts
pgarrison: [91f801d](https://github.com/allen-cell-animated/vole-core/pull/391/commits/91f801df7d0b9720f9a5e19061447579a7d86fe1)

src/FusedChannelData.ts
toloudis: naming: FusableChannelState seems to fully capture that this is a per-channel thing.

toloudis (approved):
LGTM, this is a nice change!

src/FusedChannelData.ts
toloudis: slightly wondering why not update FuseChannel to add the snapshot fields directly in there.

pgarrison: Yea, I wanted to update FuseChannel, but it's used in many other parts of the code that come before the 'snapshot' is ready. Having two separate types instead of an optional field makes it easy to statically validate that the snapshot is available in `gpuFuse`.

pgarrison: Renamed [cd924b5](https://github.com/allen-cell-animated/vole-core/pull/391/commits/cd924b525ae9dad20494d3c94bf3b69633ad8f31)

ShrimpCryptid (approved):
LGTM!

frasercl (approved):
Was skeptical at first of having to create an entirely new type for this, but I like what I see after understanding how it's used a bit more. Looks good!

PR #392, opened 2026-04-10
Feature/tri proj view

### LINKED ISSUE #463: Triple-axis projection view
## Use Case

Internal implementation for https://github.com/allen-cell-animated/vole-app/issues/376

toloudis:
# Problem

Adding a triple-projection view to vole-core, with its own mouse interaction mode in-viewport.
https://github.com/allen-cell-animated/vole-app/issues/376

# Solution

triple projection is centered in viewport.  side projections are to top and right.
Shows a single slice along each of x,y,z axis.
Interactive crosshairs are in white and can be dragged one at a time.   Double-click to jump both crosshairs to the click point.

ToDo:
* update events need to be passed out back to vole-app so that the react ui can update things like sliders, etc, as needed.
* we need to determine what, if any react ui is merited for this new functionality to be enabled in vole-app

## Type of change

## Screenshots (optional):


src/View3d.ts
```diff
@@ -430,6 +438,12 @@ export class View3d {
     this.updatePerspectiveScaleBar(img.volume);
     this.updateTimestepIndicator(img.volume);
 
+    // If we were in triple mode, re-enter it with the new image
+    if (wasTriple) {
+      this.image.setViewMode("TRIPLE", this.volumeRenderMode);
```
ShrimpCryptid: Should be using an enum

src/View3d.ts
```diff
@@ -401,6 +407,8 @@ export class View3d {
 
   // Add a new volume image to the viewer.  The viewer currently only supports a single image at a time, and will return any prior existing image.
   setImage(img: VolumeDrawable): VolumeDrawable | undefined {
+    const wasTriple = this.canvas3d.getViewMode() === Axis.TRIPLE;
+
```
ShrimpCryptid: Dan: Weird that this needs to special-case Triple axis mode

src/VolumeRenderImpl.ts
```diff
@@ -1,9 +1,22 @@
...
+/**
+ * Interface for objects that can provide triple-slice data and handle interaction.
+ * Implemented by TripleSliceVolume so ThreeJsPanel can interact with it directly
+ * without callback indirection.
+ */
+export interface TripleSliceSource {
+  getIndices(): { x: number; y: number; z: number };
+  getVolumeSize(): Vector3;
+  getPhysicalSize(): Vector3;
+  setSliceIndex(axis: "x" | "y" | "z", index: number): void;
```
ShrimpCryptid: `const enum` instead of specific strings?

ShrimpCryptid: Do we already have an Axis enum somewhere?

src/constants/shaders/slice.frag
```diff
@@ -74,14 +77,47 @@ void main() {
...
+  vec4 pos;
+  if (viewAxis == 1) {
+    // YZ view: UV.x -> Z, UV.y -> Y, slice along X (Y vertical to align with XY pane)
+    // Z_SLICE is an X voxel index; normalize by volumeSize.x
+    float xNorm = (volumeSize.x <= 1.0) ? 0.0 : float(Z_SLICE) / (volumeSize.x - 1.0);
+    pos = vec4(xNorm, vUv.y, vUv.x, 0.0);
+  } else if (viewAxis == 2) {
+    // XZ view: UV.x -> X, UV.y -> Z, slice along Y
+    // Z_SLICE is a Y voxel index; normalize by volumeSize.y
+    float yNorm = (volumeSize.y <= 1.0) ? 0.0 : float(Z_SLICE) / (volumeSize.y - 1.0);
```
toloudis: rename Z_SLICE now that it can represent a slice along any axis

src/constants/shaders/slice.frag
```diff
@@ -74,14 +77,47 @@ void main() {
...
-  // Normalize z-slice by total slices
-  vec4 pos = vec4(vUv, (SLICES == 1.0 && Z_SLICE == 0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0), 0.0);
+  // Compute the normalized slice coordinate
+  float sliceNorm = (SLICES == 1.0 && Z_SLICE == 0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0);
+
+  // Build the 3D sample position based on viewAxis
+  vec4 pos;
+  if (viewAxis == 1) {
+    // YZ view: UV.x -> Z, UV.y -> Y, slice along X (Y vertical to align with XY pane)
+    // Z_SLICE is an X voxel index; normalize by volumeSize.x
```
ShrimpCryptid: Z_SLICE should be renamed to `AXIS_SLICE` or something similar...

src/ThreeJsPanel.ts
```diff
@@ -390,6 +425,14 @@ export class ThreeJsPanel {
   }
 
   orthoScreenPixelsToPhysicalUnits(pixels: number, physicalUnitsPerWorldUnit: number): number {
+    if (this.viewMode === Axis.TRIPLE) {
```
toloudis: can this be generalized better and not special-cased like this?  maybe abstracted away somehow

src/ThreeJsPanel.ts
```diff
@@ -488,7 +531,7 @@ export class ThreeJsPanel {
   }
 
   updateScaleBarVisibility(): void {
-    const isOrtho = isOrthographicCamera(this.camera);
+    const isOrtho = isOrthographicCamera(this.camera) || this.viewMode === Axis.TRIPLE;
```
toloudis: might be able to remove, axis.triple uses ortho camera.

src/ThreeJsPanel.ts
```diff
@@ -590,6 +640,20 @@ export class ThreeJsPanel {
...
+        this.removeControlHandlers();
+
+        // In triple mode, disable all controls — no rotate/zoom/pan
+        this.perspectiveControls.enabled = false;
+        this.orthoControlsX.enabled = false;
+        this.orthoControlsY.enabled = false;
+        this.orthoControlsZ.enabled = false;
+        this.viewMode = Axis.TRIPLE;
+        // triple mode has its own controller and events
+        this.enterTripleSliceMode();
```
toloudis: could be renamed to enableTripleSliceControls?

src/ThreeJsPanel.ts
```diff
@@ -770,6 +963,78 @@ export class ThreeJsPanel {
...
+      [phys.x, phys.z], // XZ
+    ];
+    const paneRects = [panes.xy, panes.yz, panes.xz];
+
+    for (let i = 0; i < 3; i++) {
+      const rect = paneRects[i];
+      const [faceW, faceH] = faceDims[i];
+
+      this.renderer.setViewport(rect.x, rect.y, rect.w, rect.h);
+      this.renderer.setScissor(rect.x, rect.y, rect.w, rect.h);
```
toloudis: can we get these to happen inside of doRender for TripleSliceVolume? instead of viewport and scissor, should we jsut give the atlas2dslices consistent transforms?

src/ThreeJsPanel.ts
```diff
@@ -880,4 +1145,250 @@ export class ThreeJsPanel {
...
+    const y = clientY - rect.top;
+    const pane = panes[paneKey];
+
+    const u = Math.max(0, Math.min(1, (x - pane.x) / pane.w));
+    // v is inverted: top of CSS = 0, but top of UV should be 1
+    const v = Math.max(0, Math.min(1, 1 - (y - pane.y) / pane.h));
+    return { u, v };
+  }
+
+  private hitTestCrosshairLine(clientX: number, clientY: number, paneKey: "xy" | "yz" | "xz"): "u" | "v" | null {
```
toloudis: how can we hit test without giving threejspanel too much info about these regions?

src/ThreeJsPanel.ts
```diff
@@ -880,4 +1145,250 @@ export class ThreeJsPanel {
...
+      return;
+    }
+
+    const volSize = this.tripleSliceSource.getVolumeSize();
+
+    // Move both crosshairs to the double-click point
+    switch (pane) {
+      case "xy":
+        this.tripleSliceSource.setSliceIndex("x", Math.round(uv.u * (volSize.x - 1)));
+        this.tripleSliceSource.setSliceIndex("y", Math.round(uv.v * (volSize.y - 1)));
```
toloudis: should all this code only deal in uv 0-1 range values instead of whole volume pixels?  (think of multiresolution changes)

src/constants/shaders/colorizeUI.frag
toloudis, line 1: backticks in shader comments were breaking because of the vite-glsl-import plugin

src/utils/num_utils.ts
```diff
@@ -231,6 +231,16 @@ export function constrainToAxis(
   }
 }
 
+/**
+ * Clamps a slice index to the valid voxel range `[0, axisSize - 1]` for an axis of
+ * the given size, flooring fractional inputs. Returns 0 for a degenerate axis (`axisSize <= 1`).
```
ShrimpCryptid, line 236: Nit: Note about returning 0 for degenerate axes is redundant, since clamping to the range already implies this.

src/utils/tripleSliceLayout.ts
```diff
@@ -0,0 +1,81 @@
+import type { Vector3 } from "three";
+
+import type { TripleViewPanes } from "../types.js";
+
+/**
+ * Computes the uniform fit-scale and physical pane dimensions for triple-slice layout.
```
ShrimpCryptid:
```suggestion
 * Computes the uniform fit-scale and pixel pane dimensions for triple-slice layout.
```
Physical sounds like physical units in the volume imo

src/utils/tripleSliceLayout.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Pure function of the volume's physical size and the available area.
+ * @param phys Normalized physical size of the volume.
+ * @param availableW Available width (CSS pixels or world units).
+ * @param availableH Available height, in the same units as `availableW`.
+ * @param gap Gap between panes, in the same units as `availableW`/`availableH`.
+ * @returns The uniform scale (units per physical unit) and physical dimensions, or null if degenerate.
```
ShrimpCryptid: Should the comment specify what the degenerate case is? (e.g. gap is larger than available space)

Also, should this handle the case where px/py/pz are 0?

src/constants/shaders/slice.frag
```diff
@@ -31,17 +36,17 @@ vec2 offsetFrontBack(float t) {
 }
 
 vec4 sampleAtlas(sampler2D tex, vec4 pos) {
-  float bounds = float(pos[0] >= 0.0 && pos[0] <= 1.0 &&
-    pos[1] >= 0.0 && pos[1] <= 1.0 &&
-    pos[2] >= 0.0 && pos[2] <= 1.0);
+  float bounds = float(pos[0] >= 0.0 && pos[0] <= 1.0 && pos[1] >= 0.0 &&
+                       pos[1] <= 1.0 && pos[2] >= 0.0 && pos[2] <= 1.0);
```
ShrimpCryptid, line 41: Nit: Old formatting here was nicer imo, did this get auto-formatted?

src/constants/shaders/slice.frag
```diff
@@ -74,14 +79,51 @@ void main() {
...
+      normUv.y > clipMax2.y) {
     gl_FragColor = vec4(0.0);
     return;
   }
 
-  // Normalize z-slice by total slices
-  vec4 pos = vec4(vUv, (SLICES == 1.0 && Z_SLICE == 0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0), 0.0);
+  // Compute the normalized slice coordinate
+  float sliceNorm =
+      (SLICES == 1.0 && Z_SLICE == 0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0);
```
ShrimpCryptid:
```suggestion
  float sliceNorm =
      (SLICES == 1.0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0);
```

Should this always use the default 0 value when slices = 1, regardless of the z slice value? Otherwise it would return Infinity/NaN

src/constants/shaders/slice.frag
```diff
@@ -19,6 +19,11 @@ uniform int Z_SLICE;
 uniform float SLICES;
 uniform bool interpolationEnabled;
 uniform vec3 flipVolume;
+#define VIEW_AXIS_Z 0
+#define VIEW_AXIS_X 1
+#define VIEW_AXIS_Y 2
+uniform int viewAxis;
+uniform vec3 volumeSize;
```
ShrimpCryptid, line 27: Consider including a comment to specify that volumeSize = volume size in voxels?

src/Atlas2DSlice.ts
```diff
@@ -21,10 +21,22 @@ import { sliceFragmentShaderSrc, sliceShaderUniforms, sliceVertexShaderSrc } fro
 import type { VolumeRenderImpl } from "./VolumeRenderImpl.js";
 import { SettingsFlags, VolumeRenderSettings } from "./VolumeRenderSettings.js";
 import FusedChannelData from "./FusedChannelData.js";
-import type { FuseChannel } from "./types.js";
+import { Axis, type FuseChannel } from "./types.js";
 
 const BOUNDING_BOX_DEFAULT_COLOR = new Color(0xffff00);
 
+/** Maps the slice-along Axis to the integer expected by the viewAxis shader uniform. */
```
ShrimpCryptid:
```suggestion
/** Maps the slice along Axis to the integer expected by the viewAxis shader uniform. */
```

src/Atlas2DSlice.ts
```diff
@@ -70,18 +88,40 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
 
   /**
    * Syncs `this.settings.zSlice` with the corresponding shader uniform, or defers syncing until the slice is loaded.
+   * For non-XY views, zSlice represents the coordinate along the slicing axis (X for YZ, Y for XZ).
```
ShrimpCryptid: Should `this.settings.zSlice` just be renamed to `slice`?

src/Atlas2DSlice.ts
ShrimpCryptid: Same comment as before, can we rename `this.settings.zSlice` and `Z_SLICE` in the shader uniforms?

src/TripleSliceVolume.ts
```diff
@@ -0,0 +1,332 @@
...
+    _renderer: WebGLRenderer,
+    _camera: PerspectiveCamera | OrthographicCamera,
+    _depthTexture?: DepthTexture | Texture | null
+  ): void {
+    // Render all three slices; they are positioned in world space by updateLayout
+    for (let i = 0; i < 3; i++) {
+      this.renderers[i].get3dObject().visible = true;
+      this.renderers[i].doRender(_renderer, _camera);
+    }
+  }
```
ShrimpCryptid, line 144:
```suggestion
  doRender(
    renderer: WebGLRenderer,
    camera: PerspectiveCamera | OrthographicCamera,
    _depthTexture?: DepthTexture | Texture | null
  ): void {
    // Render all three slices; they are positioned in world space by updateLayout
    for (let i = 0; i < 3; i++) {
      this.renderers[i].get3dObject().visible = true;
      this.renderers[i].doRender(renderer, camera);
    }
  }
```

src/TripleSliceVolume.ts
```diff
@@ -0,0 +1,332 @@
...
+    // Re-share channel data in case the primary's was recreated
+    this.renderers[1].setSharedChannelData(this.renderers[0].getChannelData());
+    this.renderers[2].setSharedChannelData(this.renderers[0].getChannelData());
+
+    // Clamp indices to current volume bounds (resolution may differ across mode switches)
+    const volSize = this.volume.imageInfo.volumeSize;
+    const indices = this.baseSettings.tripleSliceIndices;
+    indices.x = Math.min(indices.x, Math.max(0, volSize.x - 1));
+    indices.y = Math.min(indices.y, Math.max(0, volSize.y - 1));
+    indices.z = Math.min(indices.z, Math.max(0, volSize.z - 1));
```
ShrimpCryptid: Use `clampSliceIndex` from `num_utils`?

src/TripleSliceVolume.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  private setCrosshairEndpoints(line: Line, p1: [number, number, number], p2: [number, number, number]): void {
+    const posAttr = line.geometry.getAttribute("position");
+    posAttr.setXYZ(0, p1[0], p1[1], p1[2]);
+    posAttr.setXYZ(1, p2[0], p2[1], p2[2]);
+    posAttr.needsUpdate = true;
+  }
+
+  /** Gap between panes in CSS pixels. */
+  private static readonly TRIPLE_VIEW_GAP = 2;
```
ShrimpCryptid, line 285: Nit: Make this a constant in this file instead of a class constant?

src/TripleSliceVolume.ts
```diff
@@ -0,0 +1,332 @@
...
+    const xzY = totalH / 2 - pz / 2;
+    this.renderers[2].get3dObject().position.set(xzX, xzY, 0);
+  }
+
+  /**
+   * Computes the per-pane rectangles for triple-slice view in CSS pixels (bottom-left origin).
+   * Delegates to the pure {@link computeTripleViewPanes} using this volume's physical size.
+   * @param canvasW Canvas width in CSS pixels
+   * @param canvasH Canvas height in CSS pixels
+   */
```
ShrimpCryptid, line 345:
```suggestion
  /**
   * Computes the per-pane rectangles for triple-slice view in CSS pixels (bottom-left origin).
   * @param canvasW Canvas width in CSS pixels
   * @param canvasH Canvas height in CSS pixels
   */
```
Seems unnecessary to talk about internal implementation?

src/TripleSliceControls.ts
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  // --- Event handlers ---
+
+  private onPointerDown(e: PointerEvent): void {
+    const pane = this.hitTestPane(e.clientX, e.clientY);
+    if (!pane) {
+      return;
+    }
+    const lineHit = this.hitTestCrosshairLine(e.clientX, e.clientY, pane);
```
ShrimpCryptid: Nit: rename to dragAxis? `lineHit` makes it sound like it's a boolean value.

src/ThreeJsPanel.ts
```diff
@@ -719,58 +810,70 @@ export class ThreeJsPanel {
...
-    this.camera.layers.set(MESH_LAYER);
-    this.renderer.setRenderTarget(this.meshRenderTarget);
-    this.renderer.render(this.scene, this.camera);
-
-    // Step 2. Render any passes that have to happen after the meshes are
-    // rendered but before volume rendering (e.g. pick buffer).
-    this.postMeshRenderFuncs.forEach((func) => {
-      func(this.renderer, this.camera, this.meshRenderTarget.depthTexture, this.scene);
-    });
+    if (!isTriple) {
```
ShrimpCryptid, line 825: Feels slightly weird to have all of these specifically guarding against the triple view mode. Should there be a lookup from viewmode to what rendering steps it supports?

src/VolumeDrawable.ts
```diff
@@ -963,6 +1007,23 @@ export default class VolumeDrawable {
...
+   * @param index Integer voxel index along that axis
+   */
+  setTripleSliceIndex(axis: AxisName, index: number): void {
+    const source = this.getTripleSliceSource();
+    if (source) {
+      source.setSliceIndex(axis, index);
+    } else {
+      // Fallback: update settings directly even if not in triple mode
+      const volSize = this.volume.imageInfo.volumeSize;
+      this.settings.tripleSliceIndices[axis] = clampSliceIndex(index, volSize[axis]);
```
ShrimpCryptid, line 1021: When would this happen?

src/constants/shaders/colorizeUI.frag
pgarrison: Alternate solutions:
- [Remove vite-plugin-glsl](https://github.com/allen-cell-animated/vole-core/pull/435)
- [Update vite-plugin-glsl](https://github.com/UstymUkhman/vite-plugin-glsl/pull/81)

src/utils/tripleSliceLayout.ts
toloudis: this function does not return pixels, it returns normalized physical dims (0-1 range).  made a small change to clarify this

src/ThreeJsPanel.ts
toloudis: the value of isTriple comes directly from viewmode:  
```const isTriple = this.viewMode === Axis.TRIPLE;```
it's just a little syntactic sugar thing

Comment (toloudis) [reply src/ThreeJsPanel.ts:825]: how about this compromise: if the logic gets any more complicated than this, we refactor into something a little more like a table of render passes.  I claim that for this current level of complexity it's still sufficiently readable. And I will put a todo in a comment

File: src/constants/shaders/slice.frag (hunk 5 above)
Comment (toloudis) [reply src/constants/shaders/slice.frag]: done!

File: src/Atlas2DSlice.ts
Comment (toloudis) [reply src/Atlas2DSlice.ts]: done!

File: src/TripleSliceVolume.ts (hunk 23 above)
Comment (toloudis) [reply src/TripleSliceVolume.ts:345]: done!

File: src/Atlas2DSlice.ts (hunk 19 above)
Comment (toloudis) [reply src/Atlas2DSlice.ts]: i chose the name `sliceIndex`

File: src/View3d.ts [hunk 27]
```diff
@@ -441,6 +441,15 @@ export class View3d {
     this.updatePerspectiveScaleBar(img.volume);
     this.updateTimestepIndicator(img.volume);
 
+    // If we're in triple-slice mode, re-run setCameraMode with the current mode to
+    // re-establish the crosshair source coupling with the new image. The previous
+    // image's TripleSliceVolume was just removed from the scene, so the canvas3d's
+    // controls would otherwise keep hit-testing and updating a stale source,
+    // breaking drag and double-click interaction.
+    if (this.canvas3d.getViewMode() === Axis.TRIPLE) {
+      this.setCameraMode(this.canvas3d.getViewMode().toUpperCase());
```
ShrimpCryptid, line 450: Feels odd that ViewMode and CameraMode have different capitalization... 😅

src/TripleSliceControls.ts
ShrimpCryptid: Or `selectedAxis` like it was changed to elsewhere

ShrimpCryptid (approved):
Reopened a comment or two that got missed, otherwise thanks for addressing the feedback🫡

pgarrison:
Tested this out and ran into the following bug.

# Steps to reproduce
1. `npm run start` from vole-core
2. On the vole-core demo app, select the "Timeseries JSON" dataset
3. Click "Triple"

Bug: XZ and YZ slices overlap the XY plane.

Note: If you swap steps 2 and 3, the bug is not triggered.

Screenshot 2026-08-17 at 10 41 11 AM

pgarrison:
Also, I'm not sure what the "Align" button is supposed to do, but it breaks in Triple Projection mode.

## Steps to reproduce
1. `npm run start` from vole-core
2. On the vole-core demo app, Click "Triple"
3. Select the "Timeseries JSON" dataset
4. Click "Align"

Bug: ZY and ZX slices move to the center of the canvas.

Screenshot 2026-08-17 at 11 04 35 AM

pgarrison (changes requested):
See above

toloudis:
> See above

Thank you for the extra testing.  I believe I have a fix for the issues you saw.  "Triple" mode should ignore any  "Align" transforms.  Align was meant to allow the toggling of an optional extra volume transform.

src/constants/shaders/colorizeUI.frag
pgarrison: (Resolved by https://github.com/allen-cell-animated/vole-core/pull/435)

src/constants/shaders/slice.frag
```diff
@@ -74,14 +80,51 @@ void main() {
...
+      normUv.y > clipMax2.y) {
     gl_FragColor = vec4(0.0);
     return;
   }
 
-  // Normalize z-slice by total slices
-  vec4 pos = vec4(vUv, (SLICES == 1.0 && Z_SLICE == 0) ? 0.0 : float(Z_SLICE) / (SLICES - 1.0), 0.0);
+  // Compute the normalized slice coordinate
+  float sliceNorm =
+      (SLICES == 1.0 && SLICE_INDEX == 0) ? 0.0 : float(SLICE_INDEX) / (SLICES - 1.0);
```
pgarrison, line 108: 1. `sliceNorm` is only used inside the `else` block below, so I suggest moving it into there.
2. It looks like the `(volumeSize.z <= 1.0) ? ...` logic would work for `sliceNorm` also, in which case this section could be simplified with:
```glsl
float normalizedIndex(float size) {
  return (size <= 1.0) ? 0.0 : float(SLICE_INDEX) / (size - 1.0);
}
```

src/Atlas2DSlice.ts
```diff
@@ -62,30 +81,52 @@ export default class Atlas2DSlice implements VolumeRenderImpl {
...
+        axisSize = volSize.y;
+        regionMin = this.volume.imageInfo.subregionOffset.y;
+        regionMax = regionMin + this.volume.imageInfo.subregionSize.y;
+        break;
+      default: // XY view: slicing along Z
+        axisSize = volSize.z;
+        regionMin = this.volume.imageInfo.subregionOffset.z;
+        regionMax = regionMin + this.volume.imageInfo.subregionSize.z;
+        break;
+    }
```
pgarrison, line 119: Perhaps this version is simpler?
```suggestion
    let axis: AxisName
    if (this.viewAxisValue == Axis.X || this.viewAxisValue == Axis.Y) {
      // YZ or XZ view
      axis = this.viewAxisValue;
    } else {
      axis = Axis.Z; // XY view: slicing along Z
    }
    const axisSize = volSize[axis];
    const regionMin = this.volume.imageInfo.subregionOffset[axis];
    const regionMax = regionMin + this.volume.imageInfo.subregionSize[axis];
```

src/test/TripleSliceVolume.test.ts
```diff
@@ -0,0 +1,153 @@
...
+
+describe("TripleSliceVolume", () => {
+  it("initializes each projection plane with its axis-specific dimensions", () => {
+    const volume = new Volume(imageInfo);
+    const renderer = new TripleSliceVolume(volume, new VolumeRenderSettings(volume));
+    const [xy, yz, xz] = renderer.get3dObject().children;
+
+    expect(projectionPlane(xy).scale.toArray()).toEqual([1, 0.5, 0.25]);
+    expect(projectionPlane(yz).scale.toArray()).toEqual([0.25, 0.5, 1]);
+    expect(projectionPlane(xz).scale.toArray()).toEqual([1, 0.25, 1]);
```
pgarrison, line 108: I'm afraid I don't understand why these are the expected dimensions

src/test/TripleSliceVolume.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    renderer.updateSettings(settings, SettingsFlags.TRANSFORM);
+    renderer.updateVolumeDimensions();
+
+    for (const [index, projection] of renderer.get3dObject().children.entries()) {
+      expect(projection.position.toArray()).toEqual(panePositions[index]);
+      expect([projection.rotation.x, projection.rotation.y, projection.rotation.z]).toEqual([0, 0, 0]);
+    }
+    expect(projectionPlane(xy).scale.toArray()).toEqual([1, 0.5, 0.25]);
+    expect(projectionPlane(yz).scale.toArray()).toEqual([0.25, 0.5, 1]);
+    expect(projectionPlane(xz).scale.toArray()).toEqual([1, 0.25, 1]);
```
pgarrison, line 146: If I understand correctly, this is the core Act & Assert part of the test, and the `expect` lines before and after are sanity checks? Section-header comments like Arrange, Act, Assert, Sanity Check, and Cleanup could make this test more readable.

src/ThreeJsPanel.ts
```diff
@@ -398,6 +417,14 @@ export class ThreeJsPanel {
...
   orthoScreenPixelsToPhysicalUnits(pixels: number, physicalUnitsPerWorldUnit: number): number {
+    // In triple mode, derive the conversion from the XY pane's width.
+    // (getTripleViewPanesCSS returns Y-flipped rects, but width is orientation-independent.)
+    const panes = this.getTripleViewPanesCSS();
+    if (panes && this.tripleSliceControls.source) {
+      const phys = this.tripleSliceControls.source.getPhysicalSize();
+      const worldUnitsPerPixel = phys.x / panes.xy.w;
+      return pixels * worldUnitsPerPixel * physicalUnitsPerWorldUnit;
+    }
     const worldUnitsPerPixel = 1 / (this.camera.zoom * this.getHeight());
```
pgarrison, line 428: Should `window.devicePixelRatio` be included in triple mode? If so:
```suggestion
    let worldUnitsPerPixel = 1 / (this.camera.zoom * this.getHeight());
    // In triple mode, derive the conversion from the XY pane's width.
    // (getTripleViewPanesCSS returns Y-flipped rects, but width is orientation-independent.)
    const panes = this.getTripleViewPanesCSS();
    if (panes && this.tripleSliceControls.source) {
      const phys = this.tripleSliceControls.source.getPhysicalSize();
      worldUnitsPerPixel = phys.x / panes.xy.w;
    }
```

src/ThreeJsPanel.ts
```diff
@@ -575,7 +600,13 @@ export class ThreeJsPanel {
   }
 
   switchViewMode(mode: string): void {
+    const wasTriple = this.viewMode === Axis.TRIPLE;
     mode = mode.toUpperCase();
+
+    if (wasTriple) {
+      this.tripleSliceControls.detach();
+    }
+
```
pgarrison, line 609: Nit: locality
```suggestion
    mode = mode.toUpperCase();

    if (this.viewMode === Axis.TRIPLE) {
      this.tripleSliceControls.detach();
    }
```

src/View3d.ts
```diff
@@ -520,12 +540,57 @@ export class View3d {
...
+  /**
+   * Set a triple-slice index for a given axis.
+   */
+  setTripleSliceIndex(axis: AxisName, index: number): void {
+    if (!this.image) {
+      return;
+    }
+    this.image.setTripleSliceIndex(axis, index);
     this.canvas3d.redraw();
   }
```
pgarrison, line 595: These functions appear to be unused, yet the vole-core demo app works. Are they needed? If not, there may be a cascading effect where some other dead code can be removed too.

(`TripleSliceVolume.setSliceIndex` is used instead of `setTripleSliceIndex`.)

src/View3d.ts
```diff
@@ -212,6 +216,13 @@ export class View3d {
     volume.addVolumeDataObserver(this);
     options = options || {};
     options.renderMode = this.volumeRenderMode;
```
pgarrison, line 218: This line is now irrelevant.

src/VolumeDrawable.ts
```diff
@@ -315,6 +342,10 @@ export default class VolumeDrawable {
   // @param {number} maxval -0.5..0.5, should be greater than minval
   // @param {boolean} isOrthoAxis is this an orthographic projection or just a clipping of the range for perspective view
   setAxisClip(axis: Axis, minval: number, maxval: number, isOrthoAxis?: boolean): void {
+    // Only x/y/z are valid for clip bounds indexing
+    if (axis !== Axis.X && axis !== Axis.Y && axis !== Axis.Z) {
+      return;
+    }
```
pgarrison, line 348:
```suggestion
setAxisClip(axis: AxisName, minval: number, maxval: number, isOrthoAxis?: boolean): void {
```
This guard can be handled by the compiler

src/VolumeDrawable.ts
pgarrison: This function is only called once, by View3d's `setTripleSliceIndex`, which is itself unused. Perhaps we can just get rid of this?

src/View3d.ts
pgarrison: (`TripleSliceVolume.setSliceIndex` is used instead)

src/TripleSliceControls.ts
```diff
@@ -0,0 +1,275 @@
...
+        uNorm = volSize.x > 1 ? indices.x / (volSize.x - 1) : 0.5;
+        vNorm = volSize.z > 1 ? indices.z / (volSize.z - 1) : 0.5;
+        break;
+    }
+
+    // Convert to CSS pixel positions within the pane
+    const verticalLineX = pane.x + uNorm * pane.w;
+    const horizontalLineY = pane.y + (1 - vNorm) * pane.h;
+
+    const threshold = TripleSliceControls.CROSSHAIR_GRAB_THRESHOLD;
```
pgarrison, line 157: Nit: as a shorthand for a constant, I would also put THRESHOLD in all caps.

src/TripleSliceControls.ts
```diff
@@ -0,0 +1,275 @@
...
+    const selectedAxis = this.hitTestCrosshairLine(e.clientX, e.clientY, pane);
+    if (selectedAxis) {
+      this.dragging = true;
+      this.dragPane = pane;
+      this.dragAxis = selectedAxis;
+      this.handleDrag(e.clientX, e.clientY, pane, selectedAxis);
+    }
+  }
+
+  private onPointerMove(e: PointerEvent): void {
```
pgarrison, line 190: For future work: it would be nice to use `hitTestCrosshairLine` here to set `cursor: grab` and `cursor: grabbing` CSS properties so that the interaction is more discoverable.

pgarrison (approved):
Cool! Main comment is about unused View3d API.

src/test/TripleSliceVolume.test.ts
toloudis: the volume has dims of 20, 10, 5 - which normalize to dims of 1.0, 0.5, and 0.25.  A lot of the dimension handling is based on normalizing the volume against its largest dimension

src/View3d.ts
toloudis: I believe this is anticipating driving the functionality from vole-app, which will be a follow-up pr

src/test/TripleSliceVolume.test.ts
pgarrison: Okay, so I think I would expect the following?
```
// (Z=5, Y=10, X=20) normalizes to (Z=0.25, Y=0.5, X=1)
expect(projectionPlane(xy).scale.toArray()).toEqual([1, 0.5, 0.25]); // X, Y, Z
expect(projectionPlane(yz).scale.toArray()).toEqual([0.25, 0.5, 1]); // Z, Y, X
expect(projectionPlane(xz).scale.toArray()).toEqual([1, 0.25, 0.5]); // X, Z, Y
```
Note that the final number is different: I wrote 0.5 where the test has a 1. But I'm still not really clear on whether the third dimension for these is actually used: they are planes, but the three.js docs say `.scale` is a `Vector3`.

pgarrison: I did a little more digging and found that the difference comes from the implementation:
```
case Axis.X: // YZ face: volume Z → mesh X, volume Y → mesh Y (Y vertical to align with XY)
  this.geometryMesh.scale.set(regionScale.z, regionScale.y, 1);
  break;
case Axis.Y: // XZ face: volume X → mesh X, volume Z → mesh Y
  this.geometryMesh.scale.set(regionScale.x, regionScale.z, 1);
  break;
default: // XY face (standard)
  this.geometryMesh.scale.copy(regionScale);
```
And it seems like the third coordinate for `.scale` is not used for these planes. So the main source of my confusion is:
- XY is treated differently from XZ and YZ, and
- the value `1` in the first coordinate has a different meaning from the value `1` in the third coordinate.

Therefore, I suggest the following:
```suggestion
    // (Z=5, Y=10, X=20) normalizes to (Z=0.25, Y=0.5, X=1)
    // The third coordinate of .scale is unused for these 2D planes
    expect(projectionPlane(xy).scale.toArray().slice(0, 2)).toEqual([1, 0.5]);
    expect(projectionPlane(yz).scale.toArray().slice(0, 2)).toEqual([0.25, 0.5]);
    expect(projectionPlane(xz).scale.toArray().slice(0, 2)).toEqual([1, 0.25]);
```

toloudis: ah, I agree - the test should effectively ignore the third component and will still be a good test

src/TripleSliceControls.ts
toloudis: I created an issue ticket for this in vole-app

PR #395, opened 2026-04-18
Bump `three`, `eslint`, and other dev dependencies

frasercl:
Updates `three` from r171 to r184.

This caused an odd type error in a couple `three`-internal interface declaration files, so I updated `typescript`, which waterfalled out into updating `eslint` and a number of related packages, and updating our `eslint` config for its newer format.


src/constants/colors.ts
```diff
@@ -12,12 +12,13 @@ interface HSVColor {
...
+/* eslint-disable @typescript-eslint/no-unused-expressions */
 function HSVtoRGB(h: number | HSVColor, s: number, v: number): [number, number, number] {
-  let r, g, b;
-  let hh = 0;
+  let r: number, g: number, b: number;
+  let hh: number;
   if (arguments.length === 1) {
     const hsv = h as HSVColor;
-    (s = hsv.s), (v = hsv.v), (hh = hsv.h);
+    ((s = hsv.s), (v = hsv.v), (hh = hsv.h));
```
toloudis: why not just change these "unused expressions" into assignment statements?

src/constants/colors.ts
```diff
@@ -28,31 +29,33 @@ function HSVtoRGB(h: number | HSVColor, s: number, v: number): [number, number,
...
   }
   return [Math.round(r * 255), Math.round(g * 255), Math.round(b * 255)];
 }
+/* eslint-enable @typescript-eslint/no-unused-expressions */
 
 // 1993 Park-Miller LCG
 function LCG(s: number) {
   return function () {
-    s = Math.imul(48271, s) | 0 % 2147483647;
+    s = Math.imul(48271, s) | (0 % 2147483647);
```
toloudis: It looks like the parentheses here are totally breaking this code, or else it may have already been broken?

src/loaders/zarr_utils/wrappers.ts
```diff
@@ -60,8 +60,10 @@ export default function wrapArray<
   });
 }
 
+type NewFetchStoreOptions = ConstructorParameters<typeof FetchStore>[1];
+
 export class RelaxedFetchStore extends FetchStore {
-  constructor(baseUrl: string, options?: RequestInit) {
+  constructor(baseUrl: string, options?: NewFetchStoreOptions) {
```
toloudis, line 66: this change seems to be less readable - was it fixing an error?  is there a way to make `ConstructorParameters<typeof FetchStore>[1]` more obvious?

src/loaders/RawArrayLoader.ts
```diff
@@ -157,7 +157,7 @@ class RawArrayLoader extends ThreadableVolumeLoader {
       const volSizePixels = this.data.shape[3] * this.data.shape[2] * this.data.shape[1];
       const ctor = ARRAY_CONSTRUCTORS[this.data.dtype];
       const channelData = new ctor(
-        this.data.buffer.buffer,
+        this.data.buffer.buffer as ArrayBuffer,
```
toloudis: surprising? what is the expected type vs the type of this.data.buffer.buffer?

src/Channel.ts
```diff
@@ -114,7 +113,7 @@ export default class Channel {
       }
     }
 
-    this.lutTexture.image.data.set(ret);
+    this.lutTexture.image.data = ret;
```
toloudis: is this really the same?

src/Channel.ts
```diff
@@ -257,19 +256,14 @@ export default class Channel {
...
-      tiley = Math.floor(i / numXtiles);
-      tileoffset = tilex * x + tiley * y * atlasrow;
+      const tilex = i % numXtiles;
+      const tiley = Math.floor(i / numXtiles);
+      const tileoffset = tilex * x + tiley * y * atlasrow;
       for (let j = 0; j < y; ++j) {
-        tilerowoffset = j * atlasrow;
-        destOffset = i * (x * y) + j * x;
+        const tilerowoffset = j * atlasrow;
+        const destOffset = i * (x * y) + j * x;
```
toloudis, line 266: I think the point of the "let" statements before was that this is a very tight loop running over all volume elements and I didn't want as much stack overhead pushing/popping new variables each time through.  Kind of a micro-optimization.  So will this code be equally optimal? Could be my own misunderstanding of js.

src/Lut.ts
```diff
@@ -328,10 +328,10 @@ export class Lut {
...
     let lastb = 0;
     let lasta = 0;
-    let r = 0;
-    let g = 0;
-    let b = 0;
-    let a = 0;
+    let r: number;
+    let g: number;
+    let b: number;
+    let a: number;
```
toloudis, line 334: readability suffers slightly here since the code looks less consistent with the surrounding lines.  but I understand the change

src/VolumeRenderSettings.ts
```diff
@@ -35,6 +35,7 @@ export enum Axis {
   /** Alias for NONE, indicates 3D mode */
   XYZ = "",
   /** No current axis, indicates 3D mode */
+  // eslint-disable-next-line @typescript-eslint/no-duplicate-enum-values
```
toloudis, line 38: so sad

tsconfig.json
```diff
@@ -4,10 +4,9 @@
     "allowSyntheticDefaultImports": true,
     "declaration": true,
     "declarationDir": "es/types",
-    "downlevelIteration": true,
     "lib": ["DOM", "ES2022", "ES6", "DOM.Iterable", "ScriptHost", "ES2016.Array.Include"],
-    "module": "ES2020",
-    "moduleResolution": "node",
+    "module": "Node16",
+    "moduleResolution": "node16",
```
toloudis, line 9: is Node16 newer or older than ES2020?

toloudis (commented):
a couple of questionable things but overall LGTM.  updating threejs can be fraught - have you looked at https://github.com/mrdoob/three.js/wiki/Migration-Guide ?

src/constants/colors.ts
frasercl: As in, spread them out onto multiple lines?

src/constants/colors.ts
frasercl: Hmm... may already have been broken? I assume eslint knows more about operator precedence than I do, but it might not.

Though what is the `| 0` doing here other than rounding? Surely that's already accomplished by `imul`?

src/loaders/zarr_utils/wrappers.ts
frasercl: This was fixing an error: the type of this parameter was formerly `RequestInit`, but is now an object containing multiple options, one with type `RequestInit`. Don't know why this was only caught after the updates in this PR.

We don't currently use this options argument anyways, it's here for completeness in case we need it in the future.

src/loaders/RawArrayLoader.ts
frasercl: The expected type is `ArrayBufferLike`, which is equivalent to `ArrayBuffer | SharedArrayBuffer`. It seems as though the latest TypeScript captures some new arcane API difference between the two that means `SharedArrayBuffer` is no longer acceptable where `ArrayBuffer` is expected.

On a second look at this just now, I found that the `DataView` type is generic over the kind of `ArrayBufferLike` this contains, so this fix can be moved to the type definition level to remove this cast.

src/constants/colors.ts
toloudis: As in use semicolons for each instead of commas.

src/constants/colors.ts
toloudis: yeah something is fishy there...

src/Channel.ts
frasercl: I'm not an expert so don't necessarily take my word on this, but I'm pretty certain modern compilers (hopefully including V8 and our other friendly JS JITs) are smart enough to recognize this condition and keep everything in registers, or at least to keep the stack state consistent. I might have seen an episode of Matt Godbolt's Advent of Compiler Optimizations about this?

tsconfig.json
frasercl: It's newer. This was throwing up deprecation warnings at me.

src/constants/colors.ts
```diff
@@ -28,31 +29,33 @@ function HSVtoRGB(h: number | HSVColor, s: number, v: number): [number, number,
...
   }
   return [Math.round(r * 255), Math.round(g * 255), Math.round(b * 255)];
 }
+/* eslint-enable @typescript-eslint/no-unused-expressions */
 
 // 1993 Park-Miller LCG
 function LCG(s: number) {
   return function () {
-    s = Math.imul(48271, s) | 0 % 2147483647;
+    s = (Math.imul(48271, s) | 0) % 2147483647;
```
toloudis, line 57: Consider replacing with 
https://gist.github.com/blixt/f17b47c62508be59987b?permalink_comment_id=3662485#gistcomment-3662485

src/Channel.ts
frasercl: It's replacing the buffer by reference rather than copying over its contents, which is not the same. It ought to be equivalent from three's perspective, since all it needs to do with the buffer is upload from it to the GPU, and I thought that `ret` was always created by this function so this was safe from a reference-sharing perspective. On a closer look just now, I see that's not actually the case. It might be safe *enough* and more efficient for avoiding a copy? Regardless, other types changes I just made mean that this change is no longer actually necessary.

src/constants/colors.ts
frasercl: Out of curiosity, do you know why we need seeded randomness here? Just for consistency between instances?

src/Channel.ts
toloudis: I won't worry about it until tested in a profiler in isolation

src/constants/colors.ts
toloudis: yeah for repeatability.   I'm not married to this exact method...

src/constants/colors.ts
```diff
@@ -28,31 +29,33 @@ function HSVtoRGB(h: number | HSVColor, s: number, v: number): [number, number,
   const t = v * (1 - (1 - f) * s);
   switch (i % 6) {
     case 0:
-      (r = v), (g = t), (b = p);
+      ((r = v), (g = t), (b = p));
```
toloudis:
```suggestion
      r = v; g = t; b = p;
```
would probably remove the need for the eslint-disable above

src/loaders/zarr_utils/wrappers.ts
toloudis: My only issue is that `ConstructorParameters<typeof FetchStore>[1]` means very little to me and looks brittle.

src/Channel.ts
toloudis: I did a very quick check and it looks pretty much equivalent.

src/constants/colors.ts
frasercl: had to replace it with a `prettier-ignore`

toloudis: just to keep them on the same line?

toloudis (approved):
I recommend fixing the LCG seeded pseudorandom generator with the linked version that doesn't have the same weird operator precedence mystery with `0`. 
Otherwise all LGTM!

rugeli (approved):

PR #396, opened 2026-04-20
Prefetch performance improvements

pgarrison:
This is a companion to https://github.com/allen-cell-animated/vole-app/pull/507. Please review the PR description there.

### DIFF
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -87,7 +87,7 @@ type ZarrChunkFetchInfo = {
 };
 
 const DEFAULT_FETCH_OPTIONS = {
-  maxPrefetchDistance: [5, 5, 5, 5] as [number, number, number, number],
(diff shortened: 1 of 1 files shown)


toloudis (approved):

toloudis:
I'm still curious about whether the cost of prefetched data arriving and entering our cache while playback is happening has a direct cost in terms of lag.  It sounds like you're going to look at that next? (For example, what if we never prefetch during playback?)

ShrimpCryptid (approved):

frasercl (approved):

pgarrison:
> I'm still curious about whether the cost of prefetched data arriving and entering our cache while playback is happening has a direct cost in terms of lag. It sounds like you're going to look at that next? (For example, what if we never prefetch during playback?)

@toloudis yes, prefetching has a direct lag cost. If we never prefetch, framerate goes way down (3-4x) and lag is dramatically improved.

PR #402, opened 2026-04-30
feat: Support for drawing and picking 3D spheres

ShrimpCryptid:
# Problem

Supports https://github.com/allen-cell-animated/timelapse-colorizer/issues/758, point rendering!

This change adds a new `Spheres3D` object for drawing centroids on shapes! This is used by TFE for centroid rendering. 

Spheres are instanced, which means they can all be rendered in one call to the GPU, with control for colors and sizes. 

*Estimated review size: large, 40 minutes*

# Solution

- Reorganized drawable objects into a new folder, `drawable`, with subfolders for `lines` and `spheres`.
- Added a new rendering layer, `MESH_PICK_LAYER`, which is rendered by `PickVolume` by passing in the current scene. Pickable meshes are composited with the volume pick behavior.
- Added the new `Spheres3d` object, which draws instanced spheres in the scene.
  - `Spheres3d` supports per-instance colors, transforms, and IDs.
  - When the `ids` attribute is set, `Spheres3d` will write to the pick buffer, allowing those spheres to be selected/highlighted.

## Type of change

## Screenshots (optional):

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/e990d018-3290-4bb1-8635-79913f9b1297

https://github.com/user-attachments/assets/3aa99323-8aea-4a96-a856-50eb431b3b51

| | |
| --- | --- |
| image | Spheres interact with volume |
| image | Spheres can be shown without the volume |
| image | Spheres can occlude the volume and be picked/highlighted | 

1. `Spheres3d.ts`
2. `ThreeJsPanel.ts`
3. `PickVolume.ts`


src/drawables/lines/Line3d.ts
```diff
@@ -2,9 +2,9 @@ import { Color, DataTexture, FloatType, LinearFilter, RGBAFormat, Texture } from
 import { LineSegments2 } from "three/addons/lines/LineSegments2.js";
```
ShrimpCryptid, line 2: I reorganized some drawable objects into their own subfolders for clarity.

src/drawables/IDrawableObject.ts
```diff
@@ -0,0 +1,21 @@
+import { Group, Vector3, Euler } from "three";
```
ShrimpCryptid, line 1: Moved from `src/types.ts`

src/ThreeJsPanel.ts
```diff
@@ -56,7 +63,8 @@ export type CameraState = {
 type AnimateFunction = (
   renderer: WebGLRenderer,
   camera: PerspectiveCamera | OrthographicCamera,
-  depthTexture?: DepthTexture | null
+  depthTexture?: DepthTexture | null,
+  scene?: Scene
```
ShrimpCryptid: Allow AnimateFunctions to be passed the current scene

src/PickVolume.ts
```diff
@@ -10,12 +10,15 @@ import {
   Matrix4,
```
ShrimpCryptid, line 10: PickVolume can now be passed in a reference to the current scene. If provided, it will render any objects in the scene that are on the pick mesh layer (`MESH_PICK_LAYER`) to the pick buffer before rendering the volume as a pick object.
  - Because the scene can change without `PickVolume` being notified, I removed the `needRedraw` caching flag.

src/drawables/spheres/Spheres3d.ts
```diff
@@ -0,0 +1,235 @@
...
+    this.addChildMesh(this.pickMesh);
+
+    return {
+      mesh: this.mesh,
+      pickMesh: this.pickMesh,
+      material: this.material,
+      pickMaterial: this.pickMaterial,
+      geometry: this.geometry,
+      idAttribute: this.idAttribute,
+    };
```
ShrimpCryptid, line 101: Returned just to make Typescript happy in the constructor

ShrimpCryptid (commented):
Leaving some additional comments

src/constants/shaders/volumePick.frag
```diff
@@ -220,9 +220,8 @@ void main() {
   bool hit = intersectBox(eyeRay_o, eyeRay_d, boxMin, boxMax, tnear, tfar);
 
   if (!hit) {
-    // return background color if ray misses the cube
-    // is this safe to do when there is other geometry / gObjects drawn?
-    gl_FragColor = vec4(0.0); //C1;//vec4(0.0);
+    // Discard pixel if ray misses the cube.
+    discard;
```
toloudis, line 224: ❤️

src/drawables/spheres/sphere.vert
```diff
@@ -0,0 +1,16 @@
+in uint instanceId;
+
+flat out uint IN_instanceId;
+flat out vec3 IN_instanceColor;
+
+void main() {
+    vec4 mvPosition = modelViewMatrix * instanceMatrix * vec4(position, 1.0);
```
toloudis, line 7: so ShaderMaterial automatically feeds in modelViewMatrix etc?  How quickly I've forgotten

src/drawables/spheres/Spheres3d.ts
```diff
@@ -0,0 +1,247 @@
...
+import { MESH_LAYER, MESH_PICK_LAYER } from "../../ThreeJsPanel.js";
+import { SphereMaterial, SphereMaterialInstanceAttributes, SpherePickMaterial } from "./SphereMaterial.js";
+
+const DEFAULT_INSTANCE_COUNT = 256;
+
+/**
+ * Drawable object for instanced rendering of spheres. Spheres can also be
+ * configured to be pickable for mouse interaction.
+ */
+export default class Spheres3d extends BaseDrawableMeshObject implements IDrawableObject {
```
toloudis: the `extends` and then `implements` is almost a code smell to me.   What functionality does BaseDrawableMeshObject provide beyond what's in here?
(From my old object oriented training, it's "okay" to implement many pure abstract interfaces but not okay to derive from multiple base classes. In this case you have just one of each, so it's barely okay by that metric!)
Do you ever use a Spheres3d polymorhpically as a BaseDrawableMeshObject or is that base class only used to supply common functions?

src/drawables/spheres/Spheres3d.ts
```diff
@@ -0,0 +1,247 @@
...
+  private ids: Uint32Array | null;
+  private colors: Float32Array | null;
+
+  private idAttribute: InstancedBufferAttribute;
+
+  private geometry: SphereGeometry;
+  private material: SphereMaterial;
+  private pickMaterial: SpherePickMaterial;
+
+  // There are two copies of the mesh, one which is rendered normally in the
```
toloudis: I paused here on the word "copies".  I think this comment would be simpler to parse by being more explicit:  

// `mesh` and `pickMesh` represent the same geometry, for visible rendering and rendering into the pick buffer respectively.

src/drawables/spheres/Spheres3d.ts
```diff
@@ -0,0 +1,247 @@
...
+      this.mesh.instanceColor.needsUpdate = true;
+    }
+  }
+
+  /**
+   * Sets colors for the spheres. Colors can be provided as a single Color
+   * (applied to all spheres) or as a Float32Array of RGB values. If the array
+   * has fewer colors than the number of spheres, colors will be repeated.
+   */
+  public setColors(colors: Float32Array | Color): void {
```
toloudis, line 169: I kinda get this but why not just use the colorize algorithm in the shader itself, like the image data does?  is it cheaper to do this on the cpu side?

src/drawables/spheres/Spheres3d.ts
```diff
@@ -0,0 +1,247 @@
...
+      const scaleIndex = i % (this.scales ? this.scales.length : 1);
+      const idIndex = i % (this.ids ? this.ids.length : 1);
+
+      position.fromArray(this.positions, posIndex).multiply(combinedScale);
+      const scale = this.scales[scaleIndex];
+
+      // Set per-instance matrix
+      matrix.compose(position, new Quaternion(), new Vector3(scale, scale, scale));
+      this.mesh.setMatrixAt(i, matrix);
+      this.pickMesh.setMatrixAt(i, matrix);
```
toloudis, line 202: seems like these should be able to share the buffer of matrices.  it would certainly be a speedup not to have to upload the same data to gpu twice, if possible

src/drawables/IDrawableObject.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Optional. Should be called when parent transforms are updated.
+   */
+  onParentTransformUpdated?(): void;
+  setRotation(eulerXYZ: Euler): void;
+  setFlipAxes(flipX: number, flipY: number, flipZ: number): void;
+  setOrthoThickness(thickness: number): void;
+  setResolution(x: number, y: number): void;
+  setAxisClip(axis: "x" | "y" | "z", minval: number, maxval: number, _isOrthoAxis: boolean): void;
+  updateClipRegion(xmin: number, xmax: number, ymin: number, ymax: number, zmin: number, zmax: number): void;
```
toloudis, line 19: Future: it would be interesting to audit and/or redesign this IDrawableObject api some time, because for example, why are setAxisClip and updateClipRegion different?  And I bet setResolution could go away if we were allowed to pass some parameters into doRender...

src/PickVolume.ts
```diff
@@ -124,7 +123,7 @@ export default class PickVolume implements VolumeRenderImpl {
   }
 
   public viewpointMoved(): void {
-    this.needRedraw = true;
+    // Do nothing
```
toloudis, line 126: why did needRedraw go away? I think I missed that.

src/PickVolume.ts
```diff
@@ -278,21 +265,33 @@ export default class PickVolume implements VolumeRenderImpl {
...
     renderer.setClearColor(0x000000, 0);
+    renderer.autoClear = false;
+    renderer.setRenderTarget(this.pickBuffer);
 
-    renderer.render(this.scene, camera);
+    renderer.clear();
+
+    // Render other pickable meshes into the pick buffer first.
+    if (scene) {
+      camera.layers.set(MESH_PICK_LAYER);
```
toloudis, line 282: Future: this is interesting, it fits really well to live here code-wise, but makes no logical sense.   It seems to me we should have some centralized setup more like:
```
canvas3d.setupPickBufferRendering
draw to pick buffer
canvas3d.setupOpaqueDepthTestedMeshRendering
draw opaque depth tested meshes
canvas3d.setupVolumeRendering
draw volume against current depth buffer
canvas3d.setupPathTraceRendering 
draw path traced accumulation pass
etc...
```
where all the passes with their setup calls are conditional based on viewer settings. 

This might break the Scene concept in threejs, I can't quite remember how they deal with multipass rendering where the same objects need different shading passes

toloudis (commented):
a few questions but this looks great

src/drawables/spheres/Spheres3d.ts
ShrimpCryptid: `BaseDrawableMeshObject` already implements `IDrawableObject`, so I probably don't need to repeat it. 

BaseDrawableMeshObject provides a bunch of default functionality for handling cleanup/scaling/transforms/axis flips etc., mostly internal attribute setters. It lets me skip a bunch of boilerplate when setting up new 3D objects.

src/drawables/spheres/Spheres3d.ts
ShrimpCryptid: I could, yeah... I wanted this (and VectorArrows3d) to be more flexible in case other clients of vole-core want to use them for visualization, but don't want to be locked into our specific colorization pattern.

I've so far been doing all the computation on the CPU side, and it doesn't seem to be causing performance hits because there are usually no more than a hundred objects onscreen that it needs to compute colors for.

src/PickVolume.ts
ShrimpCryptid: PickVolume is also responsible for rendering other objects to the pick buffer from the scene. If objects are changed in the scene, it doesn't have a way of knowing about that, so I have it always rendering no matter what.

src/drawables/spheres/Spheres3d.ts
toloudis: I'd definitely remove the `implements IDrawableObject` then, since BaseDMO is already implementing that.  Are you overriding some of the IDrawableObject methods in here, that were already implemented in BaseDMO?

ShrimpCryptid: Yes. Both this and `VectorArrows3d` override `setScale` and implement `onParentTransformUpdated` from BaseDMO to prevent squashing due to the volume dimensions; the other class that extends BaseDMO, `Line3d`, does not do this because it needs to be scaled relative to the volume, iirc.

src/drawables/spheres/Spheres3d.ts
ShrimpCryptid: So, it would be possible to have the same instancedMesh render the same geometry with two groups, each with a different material. But unfortunately, I wouldn't be able to separate out the layers. 

The longer-term, better solution would be to have the pick buffer attached to the main render like we discussed, and have the shader write to it in the same step.

src/PickVolume.ts
ShrimpCryptid: Made an issue for a pick buffer refactor: https://github.com/allen-cell-animated/vole-core/issues/404

src/drawables/spheres/Spheres3d.ts
toloudis: if the two meshes are truly always going to have the same matrices and same number of instances, then you could just set pickMesh.instanceMatrix = mesh.instanceMatrix; and then call setMatrixAt only once

toloudis (approved):
my concerns are some loss of performance things like extra buffer copies, and running the render pass all the time instead of conditionally; and also the weirdness around PickVolume.  But we can address it later if we do a holistic refactor on the rendering codepaths

ascibisz (approved):
Looks good to me (to the best of my reviewing ability). My understanding of shaders is quite limited - but I'm learning!

PR #405, opened 2026-05-05
Remove three.js types from public `LoadSpec` interface

### LINKED ISSUE #254: refactor: remove Box3 from LoadSpec
## Use Case

convert LoadSpec's one Box3 field and get rid of workers/utils.ts entirely

## Solution

convert the Box3 to simple array types and then the worker doesn't have to re-create classes when moving data across threads.

frasercl:
Review time: smallish (~15min)

Resolves #254: removes the three.js type `Box3` from `LoadSpec` in favor of a vanilla JS variant. This has two benefits:
- Client packages don't have to depend on three to talk to this package using this type. I believe this is the last spot where three.js types are used in the public interface, so *client packages now shouldn't have to depend on three at all*.
- We don't have to recreate the `Box3`'s prototype chain when a `LoadSpec` arrives from the volume load worker.


src/VolumeDrawable.ts
toloudis: lint says remove Box3 from here

frasercl: whoops

src/loaders/IVolumeLoader.ts
```diff
@@ -1,12 +1,26 @@
-import { Box3, Vector3 } from "three";
+import { Vector3, Box3 } from "three";
```
toloudis: should this have stayed in alphabetical order?

frasercl:
Might investigate the possibility of allowing `Box3` to still be used in this type, otherwise merging this change will require a major version bump.

src/loaders/VolumeLoaderUtils.ts
```diff
@@ -1,7 +1,7 @@
 import { Box3, Vector2, Vector3 } from "three";
 
 import { CImageInfo, type ImageInfo } from "../ImageInfo.js";
-import { LoadSpec } from "./IVolumeLoader.js";
+import { LoadSpec, type Region, regionToBox3 } from "./IVolumeLoader.js";
```
toloudis, line 4: nit: is there a place that is more "math utils"-like to put the regionToBox3 and box3toRegion functions?

src/workers/VolumeLoadWorker.ts
```diff
@@ -74,20 +73,20 @@ const messageHandlers: { [T in WorkerMsgType]: MessageHandler<T> } = {
   [WorkerMsgType.CREATE_VOLUME]: async (loadSpec, loaderId) => {
     const { loader } = getLoader(loaderId);
 
-    return await loader.createImageInfo(rebuildLoadSpec(loadSpec));
+    return await loader.createImageInfo(loadSpec);
```
toloudis, line 76: is this safe to do now because it now uses primitive types instead of the threejs classes?

toloudis (approved):
I love this change and wonder whether we have other threejs type leaks through the public api of this package

toloudis:
I wonder if there is some static analysis check we can do to make sure no threejs types are part of the public api...

ShrimpCryptid:
> I love this change and wonder whether we have other threejs type leaks through the public api of this package

I think there are some exposed types (Color, Vector3, Texture) through some of the drawable objects and the colorize configuration added to vole-core.

rugeli (approved):

src/workers/VolumeLoadWorker.ts
frasercl: Yep! All `rebuildLoadSpec` was doing was re-creating the `Box3` to get the prototype chain back.

src/loaders/VolumeLoaderUtils.ts
frasercl: There's utils/num_utils.ts, but it seems to be geared more towards formatting numbers for display

PR #406, opened 2026-05-08
refactor: Change `three` to a peer dependency

ShrimpCryptid:
# Problem

I ran into a bug in TFE where 3D views would fail, but only in built files. I eventually traced this to changes made in #395, and discovered that it was caused by having conflicting versions of three installed in TFE vs. `vole-core`.

Changing `three` to a peer dependency in `vole-core` would help prevent this in the future, so that client libraries only have one version of `three` installed. 

Couple questions:
- Should we set a maximum upper bound for what version of `three` we support, in case there are breaking changes on a minor version in the future?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
         "@babel/runtime": "^7.25.6",
         "geotiff": "^2.0.5",
         "serialize-error": "^11.0.3",
-        "three": "^0.184.0",
         "throttled-queue": "^3.0.0",
         "tweakpane": "^3.1.9",
         "zarrita": "^0.5.1"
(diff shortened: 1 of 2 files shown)


toloudis (approved):
LGTM

frasercl (approved):
Left and then deleted a comment thinking this was TFE, about how I'd discovered that my understanding of `peerDependencies` was wrong and that _library code_ needed to peer-depend on packages that provided members of its public interface rather than downstream client packages. You were clearly one step ahead of me here.

I was thrown off by the fact that `vole-app` currently peer-depends on `three` which it shouldn't have to do... but it shouldn't have to depend on `three` at all after #405 merges.

PR #419, opened 2026-06-05
update to BSD-3-clause

toloudis:
# Problem

To ease use by some external collaborators/users we need to open up the license

# Solution

BSD-3-clause, with input from Graham on this choice.

SEE ALSO https://github.com/allen-cell-animated/vole-app/pull/530


ShrimpCryptid (approved):

ShrimpCryptid:
Does vole-core also need to have a `LICENSE` file in the root directory, in addition to `LICENSE.txt`?

toloudis:
> Does vole-core also need to have a `LICENSE` file in the root directory, in addition to `LICENSE.txt`?

Not sure if this is a best practice or not.  Any preference for one vs the other? Seems like a redundancy to me.  I'll double check to see what most repos have.

toloudis:
looks like LICENSE is preferred, I'm making that change

frasercl (approved):

PR #431, opened 2026-07-14
Fix dimension ordering for out-of-order dimensions

frasercl:
Review time: tiny (<5min)

# Problem

Certain OME-Zarr images with nonstandard dimension ordering (ZCYX) were not being assembled correctly.

# Solution

The OME-Zarr utility function `orderByDimension` correctly handled *missing* dimensions, but would not reorder dimensions in nonstandard order. A small tweak solved this.

### DIFF
--- a/src/loaders/zarr_utils/utils.ts
+++ b/src/loaders/zarr_utils/utils.ts
@@ -82,19 +82,26 @@ export function remapAxesToTCZYX(axes: OMEAxis[]): TCZYX<number> {
   return axesTCZYX;
 }
 
-/** Reorder an array of values [T, C, Z, Y, X] to the given dimension order */
-export function orderByDimension<T>(valsTCZYX: TCZYX<T>, orderTCZYX: TCZYX<number>): T[] {
+/**
+ * Reorder an array of `values` in TCZYX order to the dimension order specified by `orderTCZYX`.
+ *
+ * Each element of `orderTCZYX` is the index of each dimension in its volume's native dimension
+ * order, or `-1` if the volume does not contain that dimension. For example, the array
+ * `[-1, 1, 0, 2, 3]` represents ZCYX order: T not present; C at index `1`; Z at index `0`; Y at
+ * index `2`; X at index `3`.
+ */
+export function orderByDimension<T>(values: TCZYX<T>, orderTCZYX: TCZYX<number>): T[] {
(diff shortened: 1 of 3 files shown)


src/loaders/zarr_utils/utils.ts
ShrimpCryptid: Would love it if this could be expanded with some more details or an example...

ShrimpCryptid (approved):
Took me a minute to understand the changes to the unit tests, but LGTM!

src/test/zarr_utils.test.ts
```diff
@@ -228,13 +228,13 @@ describe("zarr_utils", () => {
   describe("orderByDimension", () => {
     it("orders an array in dimension order based on the given indices", () => {
       const order: TCZYX<number> = [3, 1, 4, 0, 2];
-      expect(orderByDimension(VALS_TCZYX, order)).to.deep.equal([1, 2, 3, 4, 5]);
+      expect(orderByDimension(VALS_TCZYX, order)).to.deep.equal([4, 2, 5, 1, 3]);
```
toloudis: [Edit] I totally misunderstood this and have rewritten my summary to make more sense.
Seeing if I understand: this unit test takes a TCZYX array representing the dimension order [Y, C, X, T, Z] (which is represented as 3,1,4,0,2) and takes the actual TCZYX dimensions [1,2,3,4,5] as stored in the variable VALS_TCZYX. and returns the dimensions in the specified order, selecting Y first, then C, and so on.  This results in [4,2,5,1,3].

src/loaders/zarr_utils/utils.ts
toloudis: Yeah it could be clearer about how the ordering is applied to the values, and how -1s are treated

BrianWhitneyAI (approved):
Thank you!

src/test/zarr_utils.test.ts
frasercl: This function takes an array in TCZYX order and reorders it to the volume's native dimension order. The test data it's reordering is the array `[1, 2, 3, 4, 5]`. No part of the interface is 1-based, that's just the test data. I can definitely see how that could be confusing here though! I think the input is allowed to be an array of any type, so maybe the strings `["t", "c", "z", "y", "x"]` could be in there?

toloudis (approved):

PR #445, opened 2026-07-30
Pre-fetch low-res ZARR chunks to dedicated cache

pgarrison:
# Problem
It's difficult for users to rapidly preview and jump to a specific point in T or Z.
 
https://github.com/allen-cell-animated/vole-app/issues/523

Companion PR: https://github.com/allen-cell-animated/vole-app/pull/555

# Solution

- Automatically pre-fetch low-res chunks (this PR) and display them while either slider is scrubbed (vole-app PR).
- Use a dedicated cache for the low-res chunks to ensures that non-scrubbing interaction doesn't evict the low-res chunks, which would make display slow during scrubbing.
  - This adds a significant memory cost: this PR makes Vol-E use two 250MB caches instead of 1.
  - I selected 250MB because it's the same size as the current cache (simple) and the test AICS file I'm working with takes up 200MB at the coarsest level.
- When a ZARR is first loaded, low-priority requests are queued to fetch the maximum amount of the coarsest level that will fit in the 250MB cache.
  - For simplicity, I focused on making the experience nice for the case where the entire coarsest level fits into the cache. If it does not, the pre-fetched chunks are potentially unrelated to what the user is currently looking at.
  - The pre-fetch is the only time chunks are inserted into the low-res cache. Low-res chunks that are requested based on user interaction but aren't in the low-res cache get cached in the regular cache.
    - This means that if the user scrubs back and forth over a region, they can potentially get 250MB+250MB of cached low-res data. 

## Other changes
- Fixed a bug where `orderByDimension` was being called without `sourceIdx`, which could cause errors if multiple 

## Type of change

- ~Breaking change: I added `lowResCache` in the middle of the `OMEZarrLoader.createLoader` arguments list, to be next to the `cache` parameter. We could instead put it at the end and make this PR non-breaking~ Update: moved to end of the parameter list to make it non-breaking.
- New feature

## Steps to Verify:

* Manual testing with both the demo app (this PR adds the scrub feature there too)
* One unit test for the main cache behavior.

## Screenshots (optional):

https://github.com/user-attachments/assets/46e7df6a-c6cc-4eee-a245-fc8fdf9982c2

Demo settings:
- Input image on S3: https://allencell.s3.us-east-1.amazonaws.com/aics/emt_timelapse_dataset/data/3500005548_45_raw_converted.ome.zarr/
- I'm on the AICS VPN (somewhat slow)
- The low-res cache pre-fetching completed before I started scrubbing, because I took time to adjust the contrast and color settings first.

1. [OmeZarrLoader.beginLowResPrefetch](https://github.com/allen-cell-animated/vole-core/pull/445/changes#diff-e16c264f49783669cb26db812457a87dfb3b24b45c0348d0ef4b5a625085b076R598) is the primary function. The rest of the code is calling or supporting this.

# Areas for future improvement
- Instead of just pre-fetching an arbitrary 250MB of low-res chunks and then never inserting into that cache again, Vol-E could pre-fetch based on what is currently displayed. If the user selects a single channel, Vol-E could pre-fetch coarse chunks for just that channel, evicting other channels from the low-res cache.
- Instead of always selecting the coarsest level, Vol-E could use heuristics about the dimensions, or the size of the resolution level, to pick.
  - Slightly simpler code cleanup: decide whether vole-app or vole-core is responsible for picking the resolution level for the low-res cache, and have the other just follow that decision.
- To use less RAM, Vol-E could cache on OPFS, or use two-tiered caching with RAM (smaller) and OPFS (this is similar to how browsers already cache network requests: most are cached on disk).
- Issuing one request per chunk is not so performant for sharded ZARRs: pre-fetching can coalesce into one request per shard.

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -1006,6 +1006,14 @@ function setSyncMultichannelLoading(sync: boolean) {
   myState.loader.forEach((loader) => loader.syncMultichannelLoading(sync));
 }
 
+function setLowResPreview(enabled: boolean) {
+  // Known brittleness: this line and OmeZarrLoader.beginLowResPrefetch pick the same level for low-res
+  // pre-fetching and low-res display by selecting the coarsest level, which relies on both
+  // implementations having similar logic.
+  const scaleLevelBias = enabled ? myState.volume.imageInfo.numMultiscaleLevels : 0;
+  void myState.volume.updateRequiredData({ scaleLevelBias }, onChannelDataArrived);
+}
+
 function playTimeSeries(onNewFrameCallback: () => void) {
   window.clearTimeout(myState.timerId);
   setSyncMultichannelLoading(true);
@@ -1399,14 +1407,23 @@ function main() {
       }
     }
   });
-  // only update when DONE sliding: change event
+  // Continuously load coarse previews while actively dragging the slider.
+  timeSlider?.addEventListener("input", () => {
+    setLowResPreview(true);
+    if (goToFrame(timeSlider.valueAsNumber)) {
+      if (timeInput) {
+        timeInput.value = timeSlider.value;
+      }
+    }
+  });
+  // Restore normal automatic level selection for the final frame.
   timeSlider?.addEventListener("change", () => {
-    // trigger loading new time
     if (goToFrame(timeSlider?.valueAsNumber)) {
       if (timeInput) {
         timeInput.value = timeSlider.value;
       }
     }
+    setLowResPreview(false);
   });
   timeInput?.addEventListener("change", () => {
     // trigger loading new time
--- a/src/loaders/OmeZarrLoader.ts
+++ b/src/loaders/OmeZarrLoader.ts
@@ -35,7 +35,7 @@ import { VolumeLoadError, VolumeLoadErrorType, wrapVolumeLoadError } from "./Vol
 import { relaxedFetch, withVoleInstrumentation } from "./zarr_utils/wrappers.js";
 import { assertMetadataHasMultiscales, toOMEZarrMetaV4, validateOMEZarrMetadata } from "./zarr_utils/validation.js";
 import { remapUri } from "../utils/url_utils.js";
-import type { TypedArray } from "../types.js";
+import { ARRAY_CONSTRUCTORS, type TypedArray } from "../types.js";
 
 const CHUNK_REQUEST_CANCEL_REASON = "chunk request cancelled";
 
@@ -92,11 +92,46 @@ const DEFAULT_FETCH_OPTIONS = {
   maxPrefetchChunks: 30,
 };
 
+function getChunkByteSize(level: NumericZarrArray): number {
+  const voxelsPerChunk = level.chunks.reduce((accumulator, chunkSize) => accumulator * chunkSize, 1);
+  return voxelsPerChunk * ARRAY_CONSTRUCTORS[level.dtype].BYTES_PER_ELEMENT;
+}
+
+/**
+ * @param level One resolution level of a ZARR
+ * @yields One coords array per chunk in the level
+ */
+function* iterateChunkCoords(level: NumericZarrArray): Generator<number[]> {
+  // level.chunks[dimension] is the width of a chunk in the given dimenion
+  const chunksPerDimension = level.shape.map((size, dimension) => Math.ceil(size / level.chunks[dimension]));
+
+  /**
+   * @param coords A partial coordinates array (e.g., [1,2,3])
+   * @yields Coordinates arrays that start with the values in coords (e.g., [1,2,3,4,5])
+   */
+  function* iterateCoordsWithPrefix(coords: number[]): Generator<number[]> {
+    const nextDimension = coords.length;
+    if (nextDimension === chunksPerDimension.length) {
+      // coords is fully filled out with values in all dimensions
+      yield coords;
+      return;
+    }
+
+    for (let thisDimIdx = 0; thisDimIdx < chunksPerDimension[nextDimension]; thisDimIdx++) {
+      yield* iterateCoordsWithPrefix([...coords, thisDimIdx]);
+    }
+  }
+
+  yield* iterateCoordsWithPrefix([]);
+}
+
 class OMEZarrLoader extends ThreadableVolumeLoader {
   /** The ID of the subscriber responsible for "actual loads" (non-prefetch requests) */
   private loadSubscriber: SubscriberId | undefined;
   /** The ID of the subscriber responsible for prefetches, so that requests can be cancelled and reissued */
   private prefetchSubscriber: SubscriberId | undefined;
+  /** The ID of the subscriber responsible for filling the dedicated low-resolution cache. */
+  private lowResWarmSubscriber: SubscriberId | undefined;
 
   // TODO: this property should definitely be owned by `Volume` if this loader is ever used by multiple volumes.
   //   This may cause errors or incorrect results otherwise!
@@ -114,6 +149,8 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
     private requestQueue: SubscribableRequestQueue,
     /** Optional shared cache for decoded chunks. Keyed by `<baseUrl><arrayPath>/<chunk coords>`. */
     private cache: VolumeCache | undefined,
+    /** Optional shared cache for decoded chunks at the coarsest resolution level. */
+    private lowResCache: VolumeCache | undefined,
     /** Options to configure (pre)fetching behavior. */
     private fetchOptions: ZarrLoaderFetchOptions = DEFAULT_FETCH_OPTIONS,
     /** Direction(s) to prioritize when prefetching. Stored separate from `fetchOptions` since it may be mutated. */
@@ -129,9 +166,11 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
    *  levels with exactly the same size in every source. If matching level(s) are available, the loader will produce a
    *  volume containing all channels from every provided zarr in the order they appear in `urls`. If no matching sets
    *  of scale levels are available, creation fails.
-   * @param scenes The scene(s) to load from each URL. If `urls` is an array, `scenes` may either be an array of values
-   *  corresponding to each URL, or a single value to apply to all URLs. Default 0.
-   * @param cache A cache to use for storing fetched data. If not provided, a new cache will be created.
+   * @param scenes The index (or indices) of the image pyramid to load from each URL. If `urls` is an array, `scenes`
+   *  may either be an array of values corresponding to each URL, or a single value to apply to all URLs.
+   *  This value is used to index into the ZARR's `multiscales` array. Default 0.
+   * @param cache A cache to use for storing fetched data.
+   * @param lowResCache A dedicated cache for the coarsest resolution level.
    * @param queue A queue to use for managing requests. If not provided, a new queue will be created.
    * @param fetchOptions Options to configure (pre)fetching behavior.
    */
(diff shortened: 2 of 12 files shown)


public/index.ts
```diff
@@ -1006,6 +1006,14 @@ function setSyncMultichannelLoading(sync: boolean) {
   myState.loader.forEach((loader) => loader.syncMultichannelLoading(sync));
 }
 
+function setLowResPreview(enabled: boolean) {
+  // Known brittleness: this line and OmeZarrLoader.beginLowResPrefetch pick the same level for low-res
+  // pre-fetching and low-res display by selecting the coarsest level, which relies on both
+  // implementations having similar logic.
```
toloudis, line 1012: if the coarsest level is estimated to be too big, could we disable the previews?  what metric might you use? And at the other end of the spectrum, it could be 1 pixel by 1 pixel...

toloudis, line 1012: could the logic be refactored into a single utility function across this and OmeZarrLoader.beginLowResPrefetch? to resolve the "brittleness" comment?

public/index.ts
```diff
@@ -1399,14 +1407,23 @@ function main() {
       }
     }
   });
-  // only update when DONE sliding: change event
+  // Continuously load coarse previews while actively dragging the slider.
+  timeSlider?.addEventListener("input", () => {
+    setLowResPreview(true);
+    if (goToFrame(timeSlider.valueAsNumber)) {
```
toloudis, line 1413: does this have a way to "cancel" old fetches when people scrub the slider too fast?

src/loaders/OmeZarrLoader.ts
```diff
@@ -92,11 +92,46 @@ const DEFAULT_FETCH_OPTIONS = {
   maxPrefetchChunks: 30,
 };
 
+function getChunkByteSize(level: NumericZarrArray): number {
```
toloudis, line 95: move into zarr_utils perhaps?

src/loaders/OmeZarrLoader.ts
```diff
@@ -92,11 +92,46 @@ const DEFAULT_FETCH_OPTIONS = {
...
+  const voxelsPerChunk = level.chunks.reduce((accumulator, chunkSize) => accumulator * chunkSize, 1);
+  return voxelsPerChunk * ARRAY_CONSTRUCTORS[level.dtype].BYTES_PER_ELEMENT;
+}
+
+/**
+ * @param level One resolution level of a ZARR
+ * @yields One coords array per chunk in the level
+ */
+function* iterateChunkCoords(level: NumericZarrArray): Generator<number[]> {
```
toloudis, line 104: maybe also a candidate to move into zarr_utils somewhere

src/workers/VolumeLoadWorker.ts
```diff
@@ -18,7 +18,10 @@ import { rebuildLoadSpec } from "./util.js";
 
 type LoaderEntry = { loader: ThreadableVolumeLoader; copyOnLoad: boolean };
 
+const LOW_RES_CACHE_DEFAULT_SIZE = 250_000_000;
```
toloudis, line 21: Because this is potentially occupying user resources, this could use a comment to say whether the default is arbitrary or based on some back of the envelope calculation

toloudis (commented):
There are a couple of lint warnings that need to get cleaned up. 

I think the new functionality is good.  My main concern is that this is not useful if the lowest resolution doesn't show anything meaningful.  But I'm not sure we can really control that.  I would think to try for a min of 32x32 or 64x64 (or more?).

@frasercl are you ok with introducing a separate cache like this?  it's in effect behaving like a separate partition of a bigger cache.

public/index.ts
pgarrison: My follow-up PR #451 uses Cameron's new event model to resolve this!

pgarrison: What about this:
- If the lowest resolution level doesn't fit into the low-res cache (currently 250MB), previews are disabled.
- Among levels that would fit in the cache, if any are under 20MB (total uncompressed size, all channels), select the largest of them.
  - Rationale: At a 1.6x compression ratio (estimated from one of our ZARRs), 20MB uncompressed is ~12.5MB compressed, which could take ~2 seconds on a 50Mbps connection.
- Otherwise, select the smallest level.

I am actually not against displaying 1x1 pixel previews, if that's the best option. A ZARR with dimensions (C=10, T=1000, Z=1000) would use 20MB (uncompressed) to store a single uint16 per XY slice. That single pixel could be enough to answer "where in this file is the signal brightest", for example.

pgarrison: Oh perhaps you mean that the coarsest level could be too big on the rendering side. Vol-E currently sets `MAX_ATLAS_EDGE = 4096`, so that could be re-used here to disable scrub previews if `max(X,Y) > 4096`.

public/index.ts
pgarrison: I believe #455 should handle this gracefully (but I haven't merged the two branches together to test).
- For chunks fetched as part of low-res prefetching, scrubbing does not cancel the requests, since the pre-fetch still retains a subscriber to the fetch.
- For chunks that don't fit in the low-res cache, scrubbing quickly triggers a request, but only the currently-needed chunks have subscribers, so requests for chunks scrubbed over are aborted.

src/loaders/OmeZarrLoader.ts
```diff
@@ -432,21 +481,25 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
   }
 
   private prefetchChunk(
-    source: ZarrSource,
+    sourceIndex: number,
     scaleLevel: NumericZarrArray,
     coords: TCZYX<number>,
-    subscriber: SubscriberId
+    subscriber: SubscriberId,
+    forLowResCache: boolean = false
```
frasercl: Lint says unnecessary type annotation

src/test/zarr_wrappers.test.ts
```diff
@@ -0,0 +1,43 @@
+import type { AbsolutePath, AsyncReadable, AsyncWritable } from "zarrita";
```
frasercl: Lint says unused import `AbsolutePath`

src/loaders/OmeZarrLoader.ts
```diff
@@ -458,6 +511,8 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
 
   /** Reads a list of chunk keys requested by a `loadVolumeData` call and sets up appropriate prefetch requests. */
   private beginPrefetch(keys: ZarrChunkFetchInfo[], scaleLevel: number): void {
+    this.beginLowResPrefetch();
```
frasercl, line 483: This function's prefetch behavior is pretty stale and may need to change to match the low-res prefetch... issue for later

src/loaders/OmeZarrLoader.ts
```diff
@@ -538,6 +595,37 @@ class OMEZarrLoader extends ThreadableVolumeLoader {
...
+        if (chunkBytes > availableBytes) {
+          console.info(`Prefetching ${prefetchChunks} low-res chunks. ${(this.lowResCache.maxSize - availableBytes)/1_000_000}MB will be used.`);
+          return;
+        }
+        const coordsTCZYX = this.orderByTCZYX(coords, 0, sourceIndex);
+        this.prefetchChunk(sourceIndex, level, coordsTCZYX, this.lowResWarmSubscriber, true);
+        availableBytes -= chunkBytes;
+        prefetchChunks += 1;
+      }
+    }
```
frasercl: Would almost be inclined to give up on using the preview cache entirely if all or most of the image doesn't fit in it. Feels plausible to me that external data might have a smallest level that's still very large, such that the low-res cache size covers a relatively small percentage of the image's full 5D extent, and we end up devoting a lot of memory to a region of the volume that the user immediately scrubs away from and never returns to.

As far as I can tell, this would take care of the concern @toloudis was talking about with size: since external users can point this tool at any arbitrary zarr, we ought to make sure before filling the user's memory that the zarr has the right topology to make that memory worth filling.

frasercl (approved):
Approving preemptively but agree with a number of @toloudis's comments.

The multiple separate caches approach does bug me a bit. I don't feel like we ought to consider it a permanent solution. One alternate approach we might consider further down the line is using a single cache that's at least partially priority-based rather than fully LRU. But I think this is likely the best approach that makes use of the tools we currently have, and clearly stands to deliver a lot of benefit to users in the immediate term.

public/index.ts
toloudis: by "coarsest level too big", I meant "too big to effectively provide high speed preview".  I'm not sure if that directly corresponds to MAX_ATLAS_EDGE. I suspected it might be a lower threshold.

pgarrison: I just did a test with scrub previews at 4095x4095. 11 frames fit into the low-res cache, and scrubbing through them on my machine was smooth enough to be useful (though not snappy like the lower-res previews): cache size seems to be much more of limiting factor than render performance.

toloudis: I guess that sounds good...?
How old is your laptop? M4? M1? 
And, 11 frames out of how many total T?
I wonder if we should start chunking no more than one single Z slice per chunk... This is where chunking choices become a whack-a-mole game of different performance characteristics.

src/loaders/OmeZarrLoader.ts
pgarrison: Moved to a new file `chunks.ts` [1bc2cd5](https://github.com/AllenCell/vole-core/pull/445/commits/1bc2cd5e205dd4bc1752c7fd52ea1050e69b3463)

src/loaders/OmeZarrLoader.ts
pgarrison: Moved to a new file `chunks.ts` [1bc2cd5](https://github.com/AllenCell/vole-core/pull/445/commits/1bc2cd5e205dd4bc1752c7fd52ea1050e69b3463)

src/workers/VolumeLoadWorker.ts
pgarrison: [cc99090](https://github.com/AllenCell/vole-core/pull/445/commits/cc99090ff0dfb7b032c78bf4fc7a50945faab96c)

src/loaders/OmeZarrLoader.ts
pgarrison: [6a648a8](https://github.com/AllenCell/vole-core/pull/445/commits/6a648a851564a3c843ab1d6413f94733728fb8c6)

src/test/zarr_wrappers.test.ts
pgarrison: [6a648a8](https://github.com/AllenCell/vole-core/pull/445/commits/6a648a851564a3c843ab1d6413f94733728fb8c6)

public/index.ts
pgarrison: In [41bb253](https://github.com/AllenCell/vole-core/pull/445/commits/41bb2536dbea3ad20db0b31d4eb9ba36093206f1) I disable the low-res cache if less than 70% of the source's total TCZYX extent fits in the cache. In that case:
* No data is pre-fetched to the low-res cache
* When scrubbing, the lowest resolution level is still selected, which triggers fetches
* Data fetched while scrubbing goes to the regular cache

Some clean-up tweaks are in [d9c2f93](https://github.com/AllenCell/vole-core/pull/445/commits/d9c2f931912712fb76ce865d126171960889bb65).

If we also want to implement this idea, it would be a separate PR:
> Among levels that would fit in the cache, if any are under 20MB (total uncompressed size, all channels), select the largest of them.

src/loaders/OmeZarrLoader.ts
pgarrison: In [41bb253](https://github.com/AllenCell/vole-core/pull/445/commits/41bb2536dbea3ad20db0b31d4eb9ba36093206f1) I disable the low-res cache if less than 70% of the source's total TCZYX extent fits in the cache. In that case:
* No data is pre-fetched to the low-res cache
* When scrubbing, the lowest resolution level is still selected, which triggers fetches
* Data fetched while scrubbing goes to the regular cache

Some clean-up tweaks are in [d9c2f93](https://github.com/AllenCell/vole-core/pull/445/commits/d9c2f931912712fb76ce865d126171960889bb65).

toloudis (approved):
LGTM, eager to see if users enjoy this

Issue #448, opened 2026-08-05
Event model


frasercl:
## Use Case

In general, vole-core notifies clients of asynchronous tasks and events with callbacks, either assigned on an object or passed into the function that initiates them. These events include:

- Renders
- Channel finished loading
- Entire volume began or finished loading

This is a relatively brittle approach. Handling events by assigning a callback to an object property opens us up to conflicts if two unrelated sections of code want to be notified of the same event. (See vole-app's [`RotationControls`](https://github.com/allen-cell-animated/vole-app/blob/c136c42f93fc221cdad41b1e1546d0e37f83bc71/src/aics-image-viewer/components/dimension_sliders/RotationSliders.tsx#L78) for a concrete example of where this could cause problems.) Loading events have the option to assign a callback this way *or* to pass a callback into the method call that initiates the load, which leads to code that spreads load handling out into many awkward places and is tougher to reason about.

## Solution

Three.js demonstrates a good alternate approach: most of its objects inherit from a common `EventDispatcher` class, with methods `addEventListener`, `removeEventListener`, and `dispatchEvent`. Moving to a similar model could lead to client code that more cleanly expresses its event handling behavior, as well as library code with fewer bespoke callback-setting methods and event types.

ShrimpCryptid:
Dan: Should this be marked as done? Doesn't have a clear completion 
Cameron: Could be, okay calling this completed and adding a ticket down the line for adding event model to loaders etc.

PR #449, opened 2026-08-05
Event dispatcher (event model 1)

frasercl:
Review time: small (~10min)

# Problem

Gets the ball rolling on #448 by creating an `EventDispatcher` class and using it to refactor how client code listens for renders.

# Solution

Replaces `View3d` methods `setOnRenderCallback` and `setRenderUpdateListener` with new events `render` and `renderIteration`, respectively. `render` fires on any re-render of the canvas; `renderIteration` fires whenever the current `Volume` needs to report a new render iteration count, either on render mode change or on a render that refines a pathtraced image.

Having the `render` event in place of `setOnRenderCallback` would allow us to resolve a TODO over in vole-app in [`RotationControls`](https://github.com/allen-cell-animated/vole-app/blob/c136c42f93fc221cdad41b1e1546d0e37f83bc71/src/aics-image-viewer/components/dimension_sliders/RotationSliders.tsx#L78). I've tested that this alternative approach works there.

Internally, `View3d` continues to use callback-assignment to connect to the objects that actually report these events (`ThreeJsPanel`, `VolumeDrawable`, `PathTracedVolume`). Jury's out on whether this internal path ought to be event-ized as well.

# Future work

If we agree on this API so far, I see the possible next steps as:

- Investigate replacing the current `VolumeObserver` interface used to pass events up to `View3d` from images with events
- Replace the current tangle of callback-assignment and callback-passing used for listening to data loading updates with events
- Investigate whether more internal message passing ought to work off of events

### DIFF
--- a/src/EventDispatcher.ts
+++ b/src/EventDispatcher.ts
@@ -1,0 +1,34 @@
+export type Event<M extends Record<string, object>, E extends keyof M> = M[E] & { type: E };
+export type Listener<M extends Record<string, object>, E extends keyof M> = (event: Event<M, E>) => void;
+
+export default class EventDispatcher<EventMap extends Record<string, object>> {
+  #listeners: { [E in keyof EventMap]?: Listener<EventMap, E>[] } = {};
+
+  public addEventListener<E extends keyof EventMap>(event: E, listener: Listener<EventMap, E>): void {
+    if (this.#listeners[event] === undefined) {
+      this.#listeners[event] = [];
+    }
+    if (this.#listeners[event].indexOf(listener) === -1) {
+      this.#listeners[event].push(listener);
+    }
+  }
+
+  public removeEventListener<E extends keyof EventMap>(event: E, listener: Listener<EventMap, E>): void {
+    const listeners = this.#listeners[event];
+    if (listeners !== undefined) {
+      const index = listeners.indexOf(listener);
+      if (index > -1) {
+        listeners.splice(index, 1);
+      }
+    }
+  }
+
+  public dispatchEvent<E extends keyof EventMap>(event: Event<EventMap, E>): void {
+    const listeners = this.#listeners[event.type];
+    if (listeners !== undefined) {
+      for (const listener of listeners) {
+        listener(event);
+      }
+    }
+  }
+}
--- a/src/EventDispatcher.ts
+++ b/src/EventDispatcher.ts
@@ -2,7 +2,7 @@
 type EventMap = Record<string, object | void>;
 // An event is an object containing the event payload plus its type name (or just the name for events with no payload).
 export type Event<M extends EventMap, E extends keyof M> = M[E] extends void ? { type: E } : M[E] & { type: E };
-// Listeners are functions that receive events.
+// A listener is a function that receives an event of a specific type.
 export type Listener<M extends EventMap, E extends keyof M> = (event: Event<M, E>) => void;
 
 export default class EventDispatcher<M extends EventMap> {
--- a/src/EventDispatcher.ts
+++ b/src/EventDispatcher.ts
@@ -1,10 +1,14 @@
-export type Event<M extends Record<string, object>, E extends keyof M> = M[E] & { type: E };
-export type Listener<M extends Record<string, object>, E extends keyof M> = (event: Event<M, E>) => void;
+// An event dispatcher's *event map* maps event type names to event payloads.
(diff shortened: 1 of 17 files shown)


src/EventDispatcher.ts
```diff
@@ -0,0 +1,39 @@
...
+      this.listeners[event].push(listener);
+    }
+  }
+
+  public removeEventListener<E extends keyof M>(event: E, listener: Listener<M, E>): void {
+    const listeners = this.listeners[event];
+    if (listeners !== undefined) {
+      const index = listeners.indexOf(listener);
+      if (index > -1) {
+        listeners.splice(index, 1);
```
pgarrison, line 25: If `this.listeners[event]` is a set, this and `addEventListener` can be substantially simplified.

pgarrison (approved):

src/View3d.ts
```diff
@@ -49,14 +50,16 @@ const allGlobalLoadingOptions = {
 /**
  * @class
  */
-export class View3d {
+export class View3d extends EventDispatcher<{
+  render: void;
+  renderIteration: { iteration: number; isPathtrace: boolean };
+}> {
```
toloudis: I think it could be good for readability to lift this out and define it above View3d as something like View3dEvents or View3dEventTypes

src/EventDispatcher.ts
toloudis: do listeners ever need to be called in order?

src/View3d.ts
```diff
@@ -882,6 +877,11 @@ export class View3d {
 
     this.volumeRenderMode = mode;
     if (this.image) {
+      const isPathtrace = mode === RenderMode.PATHTRACE;
+      this.image.setRenderUpdateListener((iteration) => {
+        this.dispatchEvent({ type: "renderIteration", iteration, isPathtrace });
+      });
```
toloudis, line 915: The two layers here, with setRenderUpdateListener, are making me wonder. It seems like setRenderUpdateListener should go away somehow. Should this.image be dispatching the event?  Would that be intended as a future change?  Or do you prefer everything to come out of View3D...

toloudis (approved):
This is a good first step.  A small unit test module for EventDispatcher would make it even better

src/EventDispatcher.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default class EventDispatcher<M extends EventMap> {
+  private listeners: { [E in keyof M]?: Listener<M, E>[] } = {};
+
+  public addEventListener<E extends keyof M>(event: E, listener: Listener<M, E>): void {
+    if (this.listeners[event] === undefined) {
+      this.listeners[event] = [];
+    }
+    if (this.listeners[event].indexOf(listener) === -1) {
+      this.listeners[event].push(listener);
```
pgarrison, line 16: My IDE complains here that "Object is possibly 'undefined'", despite it being handled above.

pgarrison:
Note: this is a breaking change, since it removes `setOnRenderCallback`. I just pushed a companion vole-app patch here:
https://github.com/allen-cell-animated/vole-app/commit/7871fc8767bcad613d05ed2e0509c359224116a1

frasercl:
> Note: this is a breaking change, since it removes `setOnRenderCallback`. I just pushed a companion vole-app patch here: [allen-cell-animated/vole-app@7871fc8](https://github.com/allen-cell-animated/vole-app/commit/7871fc8767bcad613d05ed2e0509c359224116a1)

This is headed to a feature branch, which I plan to eventually merge into an ongoing branch for a new major version, `v5`.

src/EventDispatcher.ts
frasercl: It seems possible enough to me? More than that, I'm personally just more fond of lists than sets for this purpose. We don't need to ask the computer to hash these functions.

src/View3d.ts
frasercl: I agree that it would make more sense for this event to be attached to the image, and think that's a good future change. I would have tried to do that in this first proof-of-concept branch if it had been obvious and easy how to. Trouble is, `this.image` is a `VolumeDrawable`, and not publicly accessible.

frasercl:
Following conversation with @pgarrison, who wants this ASAP to address current loading synchronization bugs in #451, I've put back the old callback-assignment methods with deprecation warnings and am sending this straight to main.

PR #451, opened 2026-08-10
Emit loadStart event for scrub previews

pgarrison:
# Context
My open PRs ([vole-core](https://github.com/allen-cell-animated/vole-core/pull/445), [vole-app](https://github.com/allen-cell-animated/vole-app/pull/555)) for adding low-res previews introduce a bug, and I think the best fix is to use @frasercl's open PR #449. So the base branch for this PR is a merge commit that includes both relevent vole-core changes: I will update this when #449 merges.

# Problem
If the user manually selects a scale level from a ZARR, then scrubs, the loading indicator does not disappear after the scrub finishes (because the same resolution level is used during and after the scrub, no load is triggered when the user releases the scrub).

Currently, vole-app just assumes that data needs to be loaded, and displays the indicator. But vole-app does not have enough information to actually decide if loading is happening.

# Solution

Make vole-core tell vole-app when loading is in progress, by using a `loadStartEvent`.

# Rationale
It would make a lot of sense to also have `loadComplete` and `loadError` events. But they (mainly the `loadComplete` event) bloat this PR and, if implemented naively, add concurrency issues with stale loads. So this PR just adds the simple `loadStart`.

## Type of change

## Steps to Verify:

With the [companion PR](https://github.com/allen-cell-animated/vole-app/pull/559) in vole-app:
- Load https://allencell.s3.us-east-1.amazonaws.com/aics/emt_timelapse_dataset/data/3500005548_45_raw_converted.ome.zarr/
- Switch to manual scale level selection
- Drag the T slider, and don't release it until the loading spinner goes away.
- Release the slider.
- Verify that there is no loading spinner

### DIFF
--- a/src/Volume.ts
+++ b/src/Volume.ts
@@ -9,6 +9,7 @@ import { MAX_ATLAS_EDGE, pickLevelToLoadUnscaled } from "./loaders/VolumeLoaderU
 import type { NumberType, TypedArray } from "./types.js";
 import { type ImageInfo, CImageInfo, defaultImageInfo } from "./ImageInfo.js";
 import type { VolumeDims } from "./VolumeDims.js";
+import EventDispatcher from "./EventDispatcher.js";
 
 interface VolumeDataObserver {
   onVolumeData: (vol: Volume, batch: number[]) => void;
@@ -22,7 +23,11 @@ interface VolumeDataObserver {
  * @class
  * @param {ImageInfo} imageInfo
  */
-export default class Volume {
+export default class Volume extends EventDispatcher<{
+  loadStart: undefined;
+  loadComplete: undefined;
+  loadError: { error: unknown };
+}> {
   public imageInfo: CImageInfo;
   public loadSpec: Required<LoadSpec>;
   public loader?: IVolumeLoader;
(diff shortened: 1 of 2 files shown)


interim17 (approved):

src/Volume.ts
```diff
@@ -22,7 +23,9 @@ interface VolumeDataObserver {
  * @class
  * @param {ImageInfo} imageInfo
  */
-export default class Volume {
+export default class Volume extends EventDispatcher<{
+  loadStart: undefined;
+}> {
```
frasercl: Recommend `void` as the type of events with no payload:

```suggestion
export default class Volume extends EventDispatcher<{
  loadStart: void;
}> {
```

Could also optionally break this event map type out into its own named type. Will likely make more and more sense that way the more events we add to `Volume`.

src/Volume.ts
```diff
@@ -233,6 +237,7 @@ export default class Volume {
       ...this.loadSpecRequired,
       subregion: this.loadSpecRequired.subregion.clone(),
     };
+    this.dispatchEvent({ type: "loadStart" });
```
frasercl, line 242: Bit of a weird API corner that this will only fire when the volume triggers its own load and not when a load is triggered externally with a loader... but as far as I'm concerned the way to smooth it out is to make it no longer possible to do that

frasercl (approved):
Looks good!

src/Volume.ts
pgarrison: Made a `VolumeEvent` type [4464b57](https://github.com/AllenCell/vole-core/pull/451/commits/4464b57d6db1d1fd08d82d9396b01fe280df37d0)

src/Volume.ts
pgarrison: Agreed. I think the fix is a refactor out of scope for this PR, and would actually benefit from the new event model to communicate about updates to the `imageInfo`.

PR #452, opened 2026-08-11
Low res text

### LINKED ISSUE #523: Enable updating view when time slider and z are scrubbed
## Use Case

We may want to consider prefetching the data so that users can begin scrubbing through time as quickly as we can allow. Users want to scrub through time to get a sense for things before using play button.

-loading indicator in viewer
-second indicator for how close the app is to having all data
-fetching much lower resolution initially is ok

## Solution

_Please describe your ideal solution_

## Alternatives

pgarrison:
# Problem

The UX design for the low-res previews while scrubbing includes the text "Low-resolution preview" at the bottom center of the display.

With the vole-app [companion PR](https://github.com/allen-cell-animated/vole-app/pull/561), this resolves https://github.com/allen-cell-animated/vole-app/issues/523

# Solution

Add a new text element. vole-app is responsible for identifying when scrubbing is happening. vole-core is responsible for identifying when the scale level displayed is lower-resolution than what would be displayed if the user weren't scrubbing.

# Limitations
On smaller screens, the "Low-resolution preview" text can overlap the scale bar. If this is an issue, there are two main routes forward:
- Remove the vole-core APIs for putting the indicators in arbitrary corners, and always assume the low-res indicator is bottom-center while the scale bar and timestep are bottom-right
- Move the low-res preview text to the bottom right

## Type of change

## Steps to Verify:

- Load this test file: https://allencell.s3.us-east-1.amazonaws.com/aics/emt_timelapse_dataset/data/3500005548_45_raw_converted.ome.zarr/
- Scrub: indicator is displayed
- Switch to manual mode, and scrub: no indicator

## Screenshots

https://github.com/user-attachments/assets/bde4896c-54ff-4fe1-9e32-1ba99d3e7a71

### DIFF
--- a/public/index.ts
+++ b/public/index.ts
@@ -1007,11 +1007,11 @@ function setSyncMultichannelLoading(sync: boolean) {
 }
 
 function setLowResPreview(enabled: boolean) {
-  // Known brittleness: this line and OmeZarrLoader.beginLowResPrefetch pick the same level for low-res
-  // pre-fetching and low-res display by selecting the coarsest level, which relies on both
-  // implementations having similar logic.
-  const scaleLevelBias = enabled ? myState.volume.imageInfo.numMultiscaleLevels : 0;
-  void myState.volume.updateRequiredData({ scaleLevelBias }, onChannelDataArrived);
+  if (enabled) {
+    void view3D.enableScrubIndicator(myState.volume);
+  } else {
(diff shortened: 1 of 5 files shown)


interim17 (approved):
LGTM

rugeli (approved):

PR #460, opened 2026-08-21
Prevent tweakpane creation from crashing

### LINKED ISSUE #456: Tweakpane errors when opened
# Steps to reproduce

1. Open  vole-app  with an image loaded.
2. Press Ctrl+Alt+1 (Control+Option+1 on macOS).

# Actual behavior

The error overlay appears with:

```
[object Object]
handleError@webpack-dev-server/client/overlay.js:251:58
```

# Version
The error overlay only appears in dev builds. In prod the error is ignored and logged to the console - no user-facing issue.

pgarrison:
# Problem

If `loadSpecRequired.multiscaleLevel` is undefined, tweakpane throws an error when trying to add a UI element for `multiscaleLevel`, because it expects a strictly numeric input. This causes vole-app to have a partially-generated tweakpane, without the Pathtrace and Prefetch settings (since those come after the line that errors).

But `vole-app` intentionally uses `undefined` as the value for `multiscaleLevel` when auto mode is on.

Resolves #456

# Solution

Don't add a UI element for `multiscaleLevel` if it is undefined.
- In the vole-core demo app, the tweakpane allows adjustment of `multiscaleLevel`.
- In vole-app, the tweakpane only shows `multiscaleLevel` if the user puts vole-app in manual mode before opening the tweakpane (not ideal, but better than before).

## Type of change

## Steps to Verify:

1. Open vole-app with an image
2. Press Ctrl+Alt+1
3. The tweakpane opens, no error is logged to the console, and the Pathtrace and Prefetch sections are included

### DIFF
--- a/src/VolumeDrawable.ts
+++ b/src/VolumeDrawable.ts
@@ -938,10 +938,13 @@ export default class VolumeDrawable {
     scaleFolder
       .addInput(this.volume.loadSpecRequired, "scaleLevelBias")
       .on("change", ({ value }) => this.volume.updateRequiredData({ scaleLevelBias: value }));
-    // which level to load
-    scaleFolder
-      .addInput(this.volume.loadSpecRequired, "multiscaleLevel")
-      .on("change", ({ value }) => this.volume.updateRequiredData({ multiscaleLevel: value }));
+    // which level to load - only included if multiscaleLevel is a number
+    // (tweakpane can't handle undefined)
+    if (this.volume.loadSpecRequired.multiscaleLevel !== undefined) {
+      scaleFolder
+        .addInput(this.volume.loadSpecRequired, "multiscaleLevel")
+        .on("change", ({ value }) => this.volume.updateRequiredData({ multiscaleLevel: value }));
+    }
 
     const pathtrace = pane.addFolder({ title: "Pathtrace", expanded: false });
     pathtrace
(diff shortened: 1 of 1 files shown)


src/VolumeDrawable.ts
```diff
@@ -938,10 +938,13 @@ export default class VolumeDrawable {
     scaleFolder
       .addInput(this.volume.loadSpecRequired, "scaleLevelBias")
       .on("change", ({ value }) => this.volume.updateRequiredData({ scaleLevelBias: value }));
-    // which level to load
-    scaleFolder
-      .addInput(this.volume.loadSpecRequired, "multiscaleLevel")
-      .on("change", ({ value }) => this.volume.updateRequiredData({ multiscaleLevel: value }));
+    // which level to load - only included if multiscaleLevel is a number
+    // (tweakpane can't handle undefined)
+    if (this.volume.loadSpecRequired.multiscaleLevel !== undefined) {
```
rugeli: nit: given your comment, would checking `typeof this.volume.loadSpecRequired.multiscaleLevel === "number"` match the intent more directly?

rugeli (approved):
looks good to me! left one nit

src/VolumeDrawable.ts
meganrm: I had this same question. and is there anything you want to do if it's false?

meganrm (approved):

src/VolumeDrawable.ts
pgarrison: Addressed in [3b0d524](https://github.com/AllenCell/vole-core/pull/460/commits/3b0d524325a1614a00d02bec46b56ffb6804e028)

It's not clear to me if there's anything to be done when `multiscaleLevel` is undefined. Setting a default (e.g., 0) for tweakpane to use doesn't work, at least not without great complication, because of how tweakpane's data binding works.

PR #475, opened 2026-09-11
add a hover cursor and a dropshadow to the crosshair lines

### LINKED ISSUE #461: triple slice mode crosshair lines could use cursor change for discoverability
## Use Case

in triple axis mode, the lines could be made more obvious to grab adn drag if the cursor changed when hovering.

## Solution

In pointerMove, it would be nice to use hitTestCrosshairLine here to set cursor: grab and cursor: grabbing CSS properties so that the interaction is more discoverable.

toloudis:
# Problem

Resolves #461 

# Solution

Do a hit test on mousemove, and a cursor change when hovering over the crosshairs.
Also, add a second line as a "drop shadow" to the crosshair lines to make them more visible when they are over something very light colored.

## Type of change

https://github.com/user-attachments/assets/882b29df-52f7-4f02-8a54-dfe0a1326353


interim17 (approved):
LGTM

src/TripleSliceControls.ts
```diff
@@ -72,6 +72,7 @@ export default class TripleSliceControls {
     this.boundDblClick = undefined;
     this.dragging = false;
     this.dragAxis = undefined;
+    this.host.containerdiv.style.cursor = "";
```
ShrimpCryptid, line 75: Is there a difference between setting cursor to the empty string vs `auto`?

src/TripleSliceVolume.ts
```diff
@@ -83,20 +91,40 @@ export default class TripleSliceVolume implements VolumeRenderImpl, TripleSliceS
     // Request full volume data for YZ/XZ slicing
     volume.updateRequiredData({ subregion: new Box3(new Vector3(0, 0, 0), new Vector3(1, 1, 1)) });
 
-    // Create crosshair overlay lines (2 per pane, on OVERLAY_LAYER)
+    // Create crosshair overlay lines (2 per pane, on OVERLAY_LAYER), plus a darker
+    // shadow line behind each one for a subtle drop-shadow effect.
     this.crosshairMaterial = new LineBasicMaterial({ color: 0xffffff, depthTest: false, depthWrite: false });
```
ShrimpCryptid, line 96: Nit: Make the original color a constant too?

src/TripleSliceVolume.ts
```diff
@@ -36,9 +36,17 @@ export default class TripleSliceVolume implements VolumeRenderImpl, TripleSliceS
   private volume: Volume;
   private baseSettings: VolumeRenderSettings;
 
-  // Crosshair lines: 2 per pane (vertical + horizontal)
+  // Crosshair lines: 2 per pane (vertical + horizontal), plus a darker "drop shadow"
+  // line offset by a fixed number of screen pixels below/right of each one.
   private crosshairMaterial: LineBasicMaterial;
+  private crosshairShadowMaterial: LineBasicMaterial;
   private crosshairLines: [Line, Line, Line, Line, Line, Line]; // [xyV, xyH, yzV, yzH, xzV, xzH]
```
ShrimpCryptid, line 43: Nit: Consider making this a docstring comment

src/TripleSliceVolume.ts
```diff
@@ -83,20 +91,40 @@ export default class TripleSliceVolume implements VolumeRenderImpl, TripleSliceS
...
+      this.createCrosshairLine(this.crosshairMaterial, 1),
+      this.createCrosshairLine(this.crosshairMaterial, 1),
+      this.createCrosshairLine(this.crosshairMaterial, 1),
+      this.createCrosshairLine(this.crosshairMaterial, 1),
     ];
-    // Add each pair to its renderer's Group
-    this.renderers[0].get3dObject().add(this.crosshairLines[0], this.crosshairLines[1]);
-    this.renderers[1].get3dObject().add(this.crosshairLines[2], this.crosshairLines[3]);
-    this.renderers[2].get3dObject().add(this.crosshairLines[4], this.crosshairLines[5]);
+    // Add each pane's shadow lines before its main lines so the main lines draw on top.
```
ShrimpCryptid, line 118: Is this ordering and defining render order both necessary?

ShrimpCryptid (approved):

src/TripleSliceControls.ts
toloudis: From what I read, empty string is preferred so as not to overwrite any external css.. in other words, it's preferred when restoring from a temporarily assigned cursor style.

src/TripleSliceVolume.ts
toloudis: if the "drop shadow" does not actually overlap with the main lines, then the order doesn't matter. Except for the place where the lines cross - there you want the main lines on top always.

Issue #471, opened 2026-09-09
Document all EventDispatchers


toloudis:
## Use Case

It is hard to tell what are all the events that clients of vole-core could subscribe to.
As a developer who will use vole-core, I want to know what are all the event dispatchers that are exported, and what are all the events and payload types that they emit.

## Solution

Document this both at the export layer and in comments.

PR #480, opened 2026-09-18
Build docs with `typedoc`

### LINKED ISSUE #433: add typedoc to package.json
## Use Case

to do refactoring it is useful to have a visualization of apis, classes and maybe call graphs

## Solution

typedoc can provide most of this.  We can add it as devdependency and add a package.json script to generate the visualization for developer use.   Doesn't need to be part of ci/cd.

## Alternatives

other doc generation tools?  typedoc seems pretty good.

frasercl:
# Problem

External users of this package could use more visibility into what the package's interface looks like. Moreover, *we* could *really* use that visibility to more easily identify odd edges, confusing conventions, old code, and the like; and we could use the pressure of that visibility to motivate us to improve it.

Resolves #433; puts us in a better position for #471.

# Solution

- Switches docs builder from `documentation` (see original maintainer's [note on the project's status](https://github.com/documentationjs/documentation/wiki/Current-maintenance-status) from, uhh, eight years ago) to `typedoc`
- Removes years-old auto-generated docs
- Adds docs to existing GitHub pages build action (typedoc does not require its own action, as I'd initially thought &mdash; see [here](https://github.com/TypeStrong/typedoc/issues/1485#issuecomment-1796185086) for more)

If I've done it right, this should give us consistently up-to-date API documentation at https://AllenCell.github.io/vole-core/docs.


interim17 (approved):

toloudis:
Because this is top of mind for me at the current moment -- while we're at it, would it be easy to add URL_SPEC.md to this as a separate page?

toloudis (approved):
LGTM

ShrimpCryptid (approved):

